PR #831, opened 2014-12-08
Fix crashes for all plugins inheriting from MessageFilter

abencz:
Fix in `MessageFilterDisplay` base class to call the `processMessage` callback via a QT signal. This fixes a bunch of RViz crashes when running RViz as part of rqt.

`MessageFilterDisplay` now keep a shared pointer to a received message, but no additional copies of the data are made.

### DIFF
--- a/src/rviz/message_filter_display.h
+++ b/src/rviz/message_filter_display.h
@@ -60,8 +60,12 @@ public:
                                               this, SLOT( updateTopic() ));
     }
 
+Q_SIGNALS:
+  void messageReceived();
+
 protected Q_SLOTS:
   virtual void updateTopic() = 0;
+  virtual void _processMessage() {}
 
 protected:
   RosTopicProperty* topic_property_;
@@ -86,6 +90,7 @@ public:
     : tf_filter_( NULL )
     , messages_received_( 0 )
     {
+      connect(this, SIGNAL( messageReceived() ), this, SLOT( _processMessage() ));
       QString message_type = QString::fromStdString( ros::message_traits::datatype<MessageType>() );
       topic_property_->setMessageType( message_type );
       topic_property_->setDescription( message_type + " topic to subscribe to." );
@@ -181,8 +186,14 @@ protected:
       ++messages_received_;
       setStatus( StatusProperty::Ok, "Topic", QString::number( messages_received_ ) + " messages received" );
 
-      processMessage( msg );
+      last_message_ = msg;
+      Q_EMIT messageReceived();
     }
+
+  void _processMessage()
+  {
+    processMessage(last_message_);
+  }
 
   /** @brief Implement this to process the contents of a message.
    *
@@ -192,6 +203,9 @@ protected:
   message_filters::Subscriber<MessageType> sub_;
   tf::MessageFilter<MessageType>* tf_filter_;
   uint32_t messages_received_;
+
+private:
+  typename MessageType::ConstPtr last_message_;
 };
 
 } // end namespace rviz


samuelba:
@wjwwood can you please merge this pull request. It fixes a lot of problems with the rqt_rviz plugin.

wjwwood:
@samuelba Unless you can address our shared concerns about this fix or have new information for us to consider, I'm not going to come to a new conclusion. For example, we haven't clarified if this will have an impact on other parts of rviz (though it is @abencz's opinion that it will not affect the rest of the system and that may be true), we have not discussed what if anything needs to be changed in the `PointCloud2Display`, nor have we addressed my concerns about the change in the queueing behavior which I described. There's more work to be done here, and I don't have time to do it at the moment.

mikepurvis:
We've been running/blooming from our fork since this was originally proposed, and it does indeed resolve rqt_rviz crashes in our robot GUIs. As part of migrating to Xenial/Kinetic, we'd really like to get back on upstream for some of these key packages, but obviously that's not an option when critical fixes remain unmerged.

@wjwwood Can you clarify what it is you are looking for that would prove one way or another that this change is either a) needed or b) harmless?

wjwwood:
> @wjwwood Can you clarify what it is you are looking for that would prove one way or another that this change is either a) needed or b) harmless?

I don't doubt that this change addresses the issue, but as I mentioned above, and @hersh agreed with me, this puts a lot of new work in the GUI thread which could cause responsiveness issues. Also the way it is done would, in my opinion, allow for lost data in the `last_message_` member.

What I would need to merge something like this, is a patch that addresses those concerns by either fixing the problem in a different way or find a way to avoid the two issues I mentioned above with the current strategy, as well as a +1 from me and @hersh (or some other rviz maintainer if he doesn't have time).

The fact that you've been running this patch for a while is encouraging, but honestly the issues that I think this patch introduces are subtle and I doubt you would notice them during normal use (or if you did, you'd attribute them to system load if you didn't know to look for it). It would possibly present as a stuttering GUI during large point cloud rendering or failing to render each received point cloud.

You might not care about, or even encounter, those issues but others might, and by asking me to merge this despite my reservations, you're asking me to take on that risk of future work to address the new issues. You might think this unreasonable, but from my perspective this is trading an obvious issue (it crashes) with one or more issues that are very subtle and hard to track down. This is the very reason that I *review* code changes in addition to running CI and doing some basic acceptance testing.

I'll do my best to help anyone interested in trying to solve this issue, but I won't merge it as-is.

rhaschke:
Is this still an issue in rqt?
As pointed out in https://github.com/ros-visualization/rviz/pull/831#issuecomment-67273377, indeed rviz executes all ROS callbacks from the global queue in the gui thread in [`VisualizationManager::onUpdate()`](https://github.com/ros-visualization/rviz/blob/779cbe8e1103e6c151c43f84de22162de8c1461a/src/rviz/visualization_manager.cpp#L347), which is triggered by a [`QTimer`](https://github.com/ros-visualization/rviz/blob/779cbe8e1103e6c151c43f84de22162de8c1461a/src/rviz/visualization_manager.cpp#L227-L228) periodically.
The exception are all displays derived from `PointCloudCommon`, which manage their own `CallbackQueue` and an `AsyncSpinner`.

The current PR has the following drawbacks:
- for displays derived from `PointCloudCommon`, the asynchrony is lost
- only the last received message is actually processed
- there is no locking mechanism to protect access to the variable `last_message_`

rhaschke:
Closing due to inactivity.

PR #838, opened 2015-01-20
allow tools to get access to any keyevent

hdeeken:
The current shortkey handling for RViz Tools is quite rigid, since it prevents tools from internally using shortkeys that overlay those defined to switch between tools. An issue, I stumbled across various times now and always worked around by choosing 'free' keys, but since those are limited, I would like to a) start a general discussion on the handling of shortkeys and get an idea what the community/the maintainers think about this issues, b) propose a minimal invasive patch that extendes the shortkey handling and c) offer my help to work out a proper shortkey handling.

a) General Discussion

I checked the source code an realized that the shortkeys that call a Tool are placed within the QActions which represent the Tools in the Toolbar and that if a shortkey is pressed the Toolbar deactivates the current tool and activates the tool connected to the QAction/shortkey. This leads to two counter-intuitive behaviors.

At first, it leads to deactivating and directly activating a tool again, if one presses the respective shortkey twice. Suppose a tool gathers some sort of data internally, which is flushed on deactivate(). In the current use case, this data would be lost, while the tool is still active, seem annoying to me. I would rather expect that pressing the shortkey of an inactive tool would activate it, and pressing the shortkey of an already active tool deactivates it and leads back to the default tool. Data might still be lost, but at least the UI would be intuitively stress this, due to the change to default.

Secondly, and more importantly, shortkeys that are meant not to activate a tool, but to support a tools usage are always overriden by the toolbar shortkeys, if those happen to be alike. As an example, consider the 'Select' Tool which features pressing 'f' to focus the camera on the currently selected item. If now another tool registers 'f' as it's shortkey and is added to the toolbar, pressing 'f' while using 'Select' results in switching to the other tool, not in focusing the item. Which is even more annoying, since the internal shortkeys of an active tool should have priority over the switching to another tool.

What do you think? Do you agree?

b) Patch

I cooked up a small patch that would get rid of these problems in a minimal invasive manner. Here's what i did:

1) The Tool BaseClass was extended by a boolean called 'keep_control_', which indicates if tool want's to get access to all keypresses that occur, even those that would usually activate another tool. By default this boolean is set to false, and thus there is no need for all Tools out there to be changed in any way.

2) If a shortkey is caught by on of the QActions representing the Tools it is checked whether the current tool want's to keepControl() or not. If so, the shortkey of the QAction is turned into a KeyEvent and handed down to the current tool, so that it can use it for it's purposes. If not, the tool is switched. Or reset to default, if the shortkey happens to be the one of the current tool, which is in fact the method to switch from a "control keeping" tool to any other.

These are the changes that come with the PR. To test the new keep_control_ feature, one can checkout [this repository](http://github.com/uos/rviz_fps_plugin)  on the "keepControl" branch, where I extended the tool to declare keep_control_. I would be glad if these changes would be pulled. 

c) Designing a proper shortkey management

But I think we could even go farther and develop an even better shortkey handling for RViz Tools. My PR is quite rigid in itself again, since a tool that "keeps control" just get's all key events even if it does not use them at all. I think I would be great if we did the following:

Let the Tool base class declare fields for the keys they operate on, each with a name, description and getter and setter methods. There should be a "primary shortkey" for activation and a set of "secondary shortkeys" for internal usage. Then let's get rid of the shortkeys within the QActions and just use them for directly switching tools by clicking the toolbar and move all Tool related shortkey handling into the handleChar method of the ToolManager. Where should be checked if the pressed key is within the secondary shortkey collection of the current tool. If so it should be passed down to the tool. If not it should be checked if the key deactivates the current tool or activates another and act accordingly. If a key that is both secondary key of the current tool and primary of another this could be mentioned in the status bar, so users are aware of this.
Further the tool manager could keep track of all used keys and detect potential conflicts. With the name, description and getters and setters at hand overview and modification GUI could be implemented to resolve those conflicts or just support custom key layouts, which could be stored in the config files.

I already fooled around with some of these ideas and would offer to develop this functionality of you RViz using people out there think that this would be beneficial for all of us. If so, please feel free to provide any feedback or feature ideas that might be useful to consider.

### DIFF
--- a/src/rviz/tool.h
+++ b/src/rviz/tool.h
@@ -76,7 +76,7 @@ public:
 
   char getShortcutKey() { return shortcut_key_; }
 
-  bool keepControl() { return keep_control_;}
+  bool accessAllKeys() { return access_all_keys_; }
 
   virtual void activate() = 0;
   virtual void deactivate() = 0;
@@ -159,7 +159,7 @@ protected:
   DisplayContext* context_;
 
   char shortcut_key_;
-  bool keep_control_;
+  bool access_all_keys_;
 
   QIcon icon_;
 
--- a/src/rviz/tool.h
+++ b/src/rviz/tool.h
@@ -76,6 +76,8 @@ public:
 
   char getShortcutKey() { return shortcut_key_; }
 
+  bool keepControl() { return keep_control_;}
+
   virtual void activate() = 0;
   virtual void deactivate() = 0;
 
@@ -157,6 +159,7 @@ protected:
   DisplayContext* context_;
 
   char shortcut_key_;
+  bool keep_control_;
 
   QIcon icon_;
 
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -1058,7 +1058,7 @@ void VisualizationFrame::onToolbarActionTriggered( QAction* action )
     if( tool != current_tool )
     {
       // check if the current tool want's to keep control
-      if( current_tool->keepControl() )
+      if( current_tool->accessAllKeys() )
       {
         // if yes, this means that the incoming shortkey has to be passed to the tool manager
         QKeyEvent* key = new QKeyEvent( QEvent::KeyPress , toKey( action->shortcut() ), Qt::NoModifier,  action->shortcut() );
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -1029,16 +1029,18 @@ uint VisualizationFrame::toKey( QString const & str )
 
     // We should only working with a single key here
     if( seq.count() == 1 )
+    {
         keyCode = seq[0];
-    else {
+    }
+    else
+    {
         // Should be here only if a modifier key (e.g. Ctrl, Alt) is pressed.
         assert( seq.count() == 0 );
-
         // Add a non-modifier key "A" to the picture because QKeySequence
         // seems to need that to acknowledge the modifier. We know that A has
         // a keyCode of 65 (or 0x41 in hex)
         seq = QKeySequence( str + "+A" );
-        assert( seq.count() == 1);
+        assert( seq.count() == 1 );
         assert( seq[0] > 65 );
         keyCode = seq[0] - 65;
     }
@@ -1061,7 +1063,10 @@ void VisualizationFrame::onToolbarActionTriggered( QAction* action )
       if( current_tool->accessAllKeys() )
       {
         // if yes, this means that the incoming shortkey has to be passed to the tool manager
-        QKeyEvent* key = new QKeyEvent( QEvent::KeyPress , toKey( action->shortcut() ), Qt::NoModifier,  action->shortcut() );
+        QKeyEvent* key = new QKeyEvent( QEvent::KeyPress,
+                                        toKey( action->shortcut() ),
+                                        Qt::NoModifier,
+                                        action->shortcut() );
         manager_->handleChar( key, render_panel_ );
         // since the shortkey triggers the action button (in the tool panel) to be checked,
         // the action has to be dechecked again and the current tool needs to be checked,
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -1022,12 +1022,67 @@ void VisualizationFrame::addTool( Tool* tool )
   remove_tool_menu_->addAction( tool->getName() );
 }
 
+uint VisualizationFrame::toKey( QString const & str )
+{
+    QKeySequence seq( str );
+    uint keyCode;
+
+    // We should only working with a single key here
+    if( seq.count() == 1 )
+        keyCode = seq[0];
+    else {
+        // Should be here only if a modifier key (e.g. Ctrl, Alt) is pressed.
+        assert( seq.count() == 0 );
+
+        // Add a non-modifier key "A" to the picture because QKeySequence
+        // seems to need that to acknowledge the modifier. We know that A has
+        // a keyCode of 65 (or 0x41 in hex)
+        seq = QKeySequence( str + "+A" );
+        assert( seq.count() == 1);
+        assert( seq[0] > 65 );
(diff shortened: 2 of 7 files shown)


src/rviz/tool.h
```diff
@@ -76,6 +76,8 @@ Q_OBJECT
 
   char getShortcutKey() { return shortcut_key_; }
 
+  bool keepControl() { return keep_control_;}
```
wjwwood: Can you place a ` ` between the `;` and the `}`?

src/rviz/visualization_frame.cpp
```diff
@@ -1022,12 +1022,67 @@ void VisualizationFrame::addTool( Tool* tool )
...
   {
-    manager_->getToolManager()->setCurrentTool( tool );
+    // if the incoming tool is different from the current one
+    if( tool != current_tool )
+    {
+      // check if the current tool want's to keep control
+      if( current_tool->keepControl() )
+      {
+        // if yes, this means that the incoming shortkey has to be passed to the tool manager
+        QKeyEvent* key = new QKeyEvent( QEvent::KeyPress , toKey( action->shortcut() ), Qt::NoModifier,  action->shortcut() );
```
wjwwood: Can you put some of the arguments on new lines to keep the overall line length down? I was imagining something like this:

```c++
QKeyEvent* key = new QKeyEvent( QEvent::KeyPress,
                                toKey( action->shortcut() ),
                                Qt::NoModifier,
                                action->shortcut() );
```

Also make sure to avoid spaces between arguments and the following `,` after then and only use one space between a `,` and the following argument (both were not the case in this line).

wjwwood:
Thanks for taking the time to put together this pull request and thoughtful description.

I think the technical approach for this first step is solid, and after doing a code review the changes look good too (other than a few stylistic details). One thought I had though was, should we update any of the existing tools to use this new keep control mechanism? For example you mentioned the selection tool which can use `f` to focus the camera to the item being selected. Should update that tool (and maybe others) to use the new mechanism as part of this pull request?

Pending the above questions and comments, I think I would like to merge this pull request soon so that we can get some soak time on it in the next release.

As for the "future" work you mentioned in section `c)`, I agree with most of it. I think that the finer control over which keys are captured makes sense, but I am not passionate about how it gets done technically, as long as it provides a reasonable interface for tool developers and good usability for users. I also think it is important to provide some feedback to users when the keys collide.

Once this pull request is merged, it would be great if you could continue this conversation in a new issue or pull request and be the champion for this general topic in rviz.

Thanks again for the contribution!

hdeeken:
hi wjwwood, 
thanks for the kind reply!  I'll get back to this on monday and fix the style issues and check where applying keep control on existing tools makes sense. I wasn't sure if altering the existing tools would be preferred, but I think it's a good idea. I could come up with a little documentation as well, but currently Tools are documented [outside](http://docs.ros.org/indigo/api/rviz_plugin_tutorials/html/tool_plugin_tutorial.html) the [wiki](http://wiki.ros.org/rviz/Tutorials/Plugins%3A%20New%20Tool%20Type).  I probably can't extend the docs there, right? What's the policy about these external tutorials? Maybe it's better to migrate them into the wiki?

Regarding the future work, I'm happy to prepare a proposal for an extended key handling. I'll try to allocate some time for this asap and create an PR once I'm done.

hdeeken:
I now fixed the style issues and renamed the parameters from keepControl to accessAllKeys, which seems to be more suitable name for me. But...

I would recommend not to merge this PR just yet. I just started to work on the "future work" and would like to include one feature: If a tool is selected by clicking the icon in the toolbar, the tool should be changed no matter if the current tool accesses all keys. Currently one has to first deactivate the tool first, before clicking the toolbar allows switching. I was aware of this the other day, and though that's fine, but right now this seems to be a bug to me, considering that we would change commonly used tools like "Select" as well.  I'll fix that today.

hdeeken:
237c48994d8dc994c57f12a619ec320d1b0a195b fixes the bug i mentioned above. The shortkey logic is now moved entirely into the tool manager. Which now allows to switch tools via click on the tool bar, even if the current tool has access to all key events. 

I further made changed the selection to, it is the only standard rviz tool that currently uses shortkeys internally.

wjwwood:
Cool, those changes make total sense to me. I was going to play test it a bit before merging, I'll try to do that today.

Otherwise, are you happy with the current state of the pull request, is it ok to merge from your perspective?

wjwwood:
I just tested it out and the tool hotkeys seem to work as expected (so there at least wasn't a regression) as far as I can tell.

hdeeken:
Yeah, I'm happy with the PR and would say let's merge it, to give it "soak time" and the like! Can you tell me, when the next release will be happening?

hdeeken:
P.S. 
Do you have an opinion/advice regarding the documentation?

wjwwood:
> Can you tell me, when the next release will be happening?

Soon, I have several fixes that need to go in and I want to maximize the time between releasing the next version and a sync of the new packages to the public `.deb` repository.

> Do you have an opinion/advice regarding the documentation?

Honestly I haven't had time to get a handle on the documentation for rviz. There is quite a lot of stuff on the wiki in varying states, but the user guide is probably to most applicable user facing section for this issue:

http://wiki.ros.org/rviz/UserGuide#Tools

The developer documentation is a little more sparse, but there are related tutorials for making new tools:

http://wiki.ros.org/rviz/Tutorials/Plugins%3A%20New%20Tool%20Type

Which actually just redirects to:

http://docs.ros.org/indigo/api/rviz_plugin_tutorials/html/tool_plugin_tutorial.html

I think this last link would be the right place to add documentation about the changes in this pull request. That documentation is generated from:

https://github.com/ros-visualization/visualization_tutorials/blob/indigo-devel/rviz_plugin_tutorials/src/doc/tool_plugin_tutorial.rst

Which is in the `visualization_tutorials` repository.

hdeeken:
Perfect, the last link was exactly what I was looking for. I'll add some docs there. 
And thank you very much for proof reading, suggestions and merging!

Issue #650, opened 2013-06-26
Width of Path in RViz can not be changed?


cnpcshangbo:
I am going to report my project with a record of RViz, however, Width of Path can not be changed. If I can get the Path to be thicker, the audience may see the lines more clearly.

tlind:
I implemented this in pull request #842. This required the addition of a new display style "Billboards" for paths, similar to how it is done in the Grid display, since the width cannot be changed for the default "Lines" style.

wjwwood:
I've merged @tlind's pr #842, so I'm going to close this for now.

PR #842, opened 2015-02-04
Adjustable line width for Path display + offset property

tlind:
This adds a new display style called "Billboards" to the Path display, similar to how it is implemented for the Grid display. For the "Billboards" style, the line width can be adjusted, which implements enhancement request #650 and is helpful when e.g. displaying paths using a projector. The old display style "Lines" is preserved and left untouched for backwards compatibility.

I also added an "Offset" property to the Path display, that allows to shift the displayed path in relation to the origin of the fixed frame. Useful to e.g. shift the Z coordinate when displaying lots of information in Rviz that otherwise overlaps.

### DIFF
--- a/src/rviz/default_plugin/path_display.cpp
+++ b/src/rviz/default_plugin/path_display.cpp
@@ -43,6 +43,7 @@
 #include "rviz/properties/color_property.h"
 #include "rviz/properties/float_property.h"
 #include "rviz/properties/int_property.h"
+#include "rviz/properties/vector_property.h"
 #include "rviz/validate_floats.h"
 
 #include "rviz/ogre_helpers/billboard_line.h"
@@ -77,6 +78,10 @@ PathDisplay::PathDisplay()
                                              "Number of paths to display.",
                                              this, SLOT( updateBufferLength() ));
   buffer_length_property_->setMin( 1 );
+
+  offset_property_ = new VectorProperty( "Offset", Ogre::Vector3::ZERO,
+                                         "Allows you to offset the path from the origin of the reference frame.  In meters.",
+                                         this, SLOT( updateOffset() ));
 }
 
 PathDisplay::~PathDisplay()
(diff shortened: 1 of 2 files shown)


src/rviz/default_plugin/path_display.cpp
```diff
@@ -80,35 +101,108 @@ void PathDisplay::reset()
...
+  context_->queueRender();
+}
+
 void PathDisplay::destroyObjects()
 {
+  // Destroy all simple lines, if any
   for( size_t i = 0; i < manual_objects_.size(); i++ )
   {
-    Ogre::ManualObject* manual_object = manual_objects_[ i ];
+    Ogre::ManualObject*& manual_object = manual_objects_[ i ];
```
wjwwood: Can you explain why you made this particular change?

src/rviz/default_plugin/path_display.cpp
```diff
@@ -80,35 +101,108 @@ void PathDisplay::reset()
...
     if( manual_object )
     {
       manual_object->clear();
       scene_manager_->destroyManualObject( manual_object );
+      manual_object = NULL; // ensure it doesn't get destroyed again
```
wjwwood: My guess is so that you can do this and set the pointer in the vector to `NULL`?

wjwwood:
Other than the comment this looks good, and I tested it locally and it worked as described!

One thing I ran into was that occasionally due to the nature of the billboards the paths would not render after flipping the camera around to look at them from the other side.The normal side would not flip to point at the camera until I zoomed really far in. I think this is just inherent in the billboards style point, so I don't think it needs to be addressed here. I think this is g2g.

src/rviz/default_plugin/path_display.cpp
tlind: Exactly, that is the reason.

PR #845, opened 2015-02-12
Add a simple 'About' dialog to the help menu.

gavanderhoorn:
As per subject.

rviz_about

Now I don't have to scroll back in the console all the time (working with multiple workspaces).

I thought about adding the version to the window title, but that seemed a bit involved for now.

I'm afraid I've only been able to test this on Ubuntu Precise.


wjwwood:
That's awesome, thanks!

PR #849, opened 2015-02-14
Check if position of wrench is not nan

garaemon:
Sometime position of wrench is set to nan, this patch is to avoid that issue

### DIFF
--- a/src/rviz/default_plugin/wrench_display.cpp
+++ b/src/rviz/default_plugin/wrench_display.cpp
@@ -123,6 +123,12 @@ namespace rviz
 	    return;
 	  }
 
+        if ( position.isNaN() )
+          {
+            ROS_ERROR_THROTTLE(1.0, "Wrench position contains NaNs. Will skip rendering as long as it will have valid position");
+            return;
+          }
+
         // We are keeping a circular buffer of visual pointers.  This gets
         // the next one, or creates and stores it if the buffer is not full
         boost::shared_ptr<WrenchStampedVisual> visual;


adolfo-rt:
The patch is prone to spamming the logs pretty hard in case wrenches become NaN for prolonged periods of time.

How about just not doing anything?.

Alternatively, if you want to give feedback, how about printing a warning (not error) when you start receiving NaNs, and then another one when you stop?. It could look something like:

```
[WARN] Wrench message contains NaNs. Will skip rendering as long as it's the case...
...
...
[WARN] Wrench message no longer contains NaNs (skipped 120 messages in 4.0s).
```

I'd find this much more informative and log-friendly on NaN bursts.

wjwwood:
I generally +1 what @adolfo-rt said. At the very least you should consider printing a message with more context as `position is nan` is pretty generic and doesn't give me any hints as to what has a position of nan (in this case a Wrench message).

src/rviz/default_plugin/wrench_display.cpp
```diff
@@ -122,6 +122,12 @@ namespace rviz
 		       msg->header.frame_id.c_str(), qPrintable( fixed_frame_ ));
 	    return;
 	  }
+	
+	if ( position.isNaN() )
```
wjwwood: Please make sure the indention is consistent with the rest of the file and uses spaces not tabs.

garaemon:
I updated message and use whitespace instead of hard tabs.

wjwwood:
Sorry to nitpick, but this new sentence may not be grammatically correct: "Will skip rendering as long as it will have valid position"

Maybe something like this would make more sense: "Skipping render as long as the position is valid" I'm not entirely sure this is what you're trying to convey.

adolfo-rt:
> Maybe something like this would make more sense: "Skipping render as long as the position is valid" I'm not entirely sure this is what you're trying to convey.

Rather the opposite:
"Wrench position contains NaNs. Skipping render as long as the position is **in**valid"

Alternative wording based on very similar example found [here](http://wiki.ros.org/Verbosity%20Levels)
"Received unexpected NaN value in wrench position. Skipping rendering as long as it's the case."

wjwwood:
I would be fine with either of those proposed by @adolfo-rt, I think the current one is at best unclear, and at worst it has a typo (`invalid` vs `valid`).

garaemon:
I updated this patch (thanks to @k-okada )

wjwwood:
lgtm.

wjwwood:
Indigo/Jade forward port: https://github.com/ros-visualization/rviz/pull/906

PR #852, opened 2015-02-24
init tool's shortcut_key_ and fixed toKeys()

### LINKED ISSUE #851: rviz sometimes crashes on startup because of assertion in "rviz::ToolManager::toKey()"
About 50% of the time I start rviz now, I'm seeing this crash before the window even comes up:
```
rviz: .../tool_manager.cpp:134: uint rviz::ToolManager::toKey(const QString&): Assertion `seq[0] > 65' failed.
```
I'm building this from the latest source.

The assertion was originally introduced here: https://github.com/ros-visualization/rviz/commit/8277331f6ac0743dcde4bc8284b9e3985e74e774 but has sinced moved to tool_manager.cpp.

The assertions should be replaced with conditionals, and the function signature should be changed to return a boolean success/error indicator and the output value should be returned by a reference parameter, like so:
```
bool toKey(QString const& str, uint& key_out);
```

hdeeken:
* initialized the shortcut_key_ param with '/0' to be able to check whether a tool has a shortkey assigned or not
* made the tool manager check if a tool has a shortkey before converting the char to a key code
* (hopefully) fixed the toKeys method by removing the assertions, making at a boolean returning function and allowing a single key only,
  as this is what is to be expected from the shortcut_key_ param

this should fix #851

### DIFF
--- a/src/rviz/tool.cpp
+++ b/src/rviz/tool.cpp
@@ -43,6 +43,7 @@ Tool::Tool()
   : property_container_( new Property() )
 {
   access_all_keys_ = false;
+  shortcut_key_ = '\0';
 }
 
 Tool::~Tool()
--- a/src/rviz/tool_manager.cpp
+++ b/src/rviz/tool_manager.cpp
@@ -111,31 +111,20 @@ void ToolManager::save( Config config ) const
   }
 }
 
-uint ToolManager::toKey( QString const & str )
+bool ToolManager::toKey( QString const& str, uint& key )
 {
-    QKeySequence seq( str );
-    uint keyCode;
+  QKeySequence seq( str );
 
-    // We should only working with a single key here
-    if( seq.count() == 1 )
-    {
-        keyCode = seq[0];
-    }
-    else
-    {
-        // Should be here only if a modifier key (e.g. Ctrl, Alt) is pressed.
-        assert( seq.count() == 0 );
-
-        // Add a non-modifier key "A" to the picture because QKeySequence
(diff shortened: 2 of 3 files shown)


wjwwood:
This basically drops the "multi key" support? I'm with @hersh I'm not exactly sure what the issue was. Also, if the appropriate fix is to remove that code, then what did it do before you removed it?

My understanding was the the `A` (65) needed to be added (`str + "+A"`) for some side effect and then manually remove that extra `A` with `- 65`, is that no longer the case?

hdeeken:
No worries, the removed code is neither important to keep nor does it drop any features.

I introduced the toKey() method, because we need a conversion between shortcut_key_ parameter, a char, to the corresponding Qt key code, because we need this lookup later on.* The original method was quite intended to convert any QString to the corresponding key code, but to include keys, like SHIFT, it is necessary to do this str+A, key-65 magic due to the way Qt treats these modifier keys. But we do not need this magic, since a tool's shortcut is char and meant to be a letter of the alphabet, we will not encounter any modifier keys here. So removing that is fine.

*Regarding Multi Key Support : That's established in ToolManager::handleChar and is not bothered by the toKey() method, as long as the map that links keycodes to tools is properly setup. What's relevant here is that the QtKeyEvents keycode properly maps to the respective tool, that's why we have to deal with all these Qt stuff in the first place.

wjwwood:
Ok, sounds good to me. I'll assume this solves the issue for @hersh and merge. Please comment here otherwise.

hersh:
Thanks guys!

Issue #854, opened 2015-02-26
Alpha transparency of mesh markers is broken


hersh:
Back in the old days when I worked on rviz, I made mesh markers obey the "alpha" field of their color, like all the other marker types do.  I made a test program that shows a bunch of combinations.  Recently (on HEAD) I was trying to show a semi-transparent mesh marker and it was shown opaque.  I re-ran my test program (src/test/mesh_marker_test.cpp) and I see that it no longer behaves as I intended.

Here is the new image:
![rviz-alpha-bug](https://cloud.githubusercontent.com/assets/1562730/6384040/0ae81d36-bd0c-11e4-8486-72d40c2b3534.png)

You can see the spheres in the image show the alphas as intended.

Here is an old correct image, from 9c5bbf9d6d41309aecd575f7310c26f471b8dff0 :
![rviz-no-bug](https://cloud.githubusercontent.com/assets/1562730/6384186/f4309d50-bd0d-11e4-837c-37ae1ae26ac5.png)

This version still worked: 4be608f13d65ec0ebb7ee46240e3274ff28627a0
This version has the bug: f75ad94

trainman419:
It looks like this was introduced in #752 

It looks like your old program and @brindza 's changes are both achieving the same results, through different APIs.

Thoughts @brindza ?

hersh:
Damn, it looks like I "+1"ed the original change.  I guess I didn't run my test program at the time. :(

hersh:
Perhaps the issue is that the colorized meshes in #752 and the pr2-base meshes I show here use different features of the collada file type.  The test program mesh_marker_test.cpp should really include something like the rgb axes shown in #752 which showed the problem there.

NikolausDemmel:
Just browsing through the issue list. I guess this can be closed, given #893 is released?

NikolausDemmel:
(Maybe the port forward to jade is still missing, at least there is a label #893 that says so)

wjwwood:
@NikolausDemmel I must have accidentally put that tag on there. Jade and Indigo are released from the same branch.

wjwwood:
I'll let @hersh decide if this can be closed based on #893

LemonPi:
This is still an issue

rhaschke:
The default use case (`mesh_use_embedded_materials = False`) works as expected:
![image](https://user-images.githubusercontent.com/5376030/181629306-378e8808-679e-4076-b286-810494198030.png)

[...]

PR #855, opened 2015-02-26
Revert "color tint" change to mesh markers to fix transparency bug.

hersh:
This reverts commit f75ad9499bcc498cddc060a6b10f0b2eb3968a04.
That change introduced bug #854, where the alpha transparency information
for mesh markers was lost.

I suppose this will break @brindza 's code which perhaps depends on the new behavior or API, but it fixes mine, which depends on the old behavior.

Maybe this is passive-aggressive of me... I haven't yet taken the time to understand what his changes were supposed to do, so I don't have an idea about how to make his stuff work and mine as well.

### DIFF
--- a/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
+++ b/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
@@ -65,14 +65,13 @@ void MeshResourceMarker::reset()
   //destroy entity
   if (entity_)
   {
-    context_->getSceneManager()->destroyEntity(entity_);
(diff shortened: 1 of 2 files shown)


hersh:
Looks like this ignores #752.  We should work out the correct code first, I guess.

wjwwood:
@hersh I'm afraid I don't have time to dig into the issue, but I'm happy to help with reviewing, testing, and releasing.

Also, maybe we should get an acceptance testing guide put together. It seems that you have some test programs which have a visual expectation. If I had a list of those tests and the expected result of them, then I could check them before merging or at least before releasing.

PR #857, opened 2015-03-02
minor gui improvements

hdeeken:
i stumbled upon a minor issue while creating a tool that behaves like a trigger. in the rare case where a tool invokes another tool during it's activation the gui becomes inconsistent. 

then i found that there's no method to redraw a tool's icon and text in the toolbar, so i wrote a small refresh method.

### DIFF
--- a/src/rviz/tool_manager.cpp
+++ b/src/rviz/tool_manager.cpp
@@ -183,7 +183,7 @@ void ToolManager::setCurrentTool( Tool* tool )
     current_tool_->activate();
   }
 
-  Q_EMIT toolChanged( tool );
+  Q_EMIT toolChanged( current_tool_ );
 }
 
 void ToolManager::setDefaultTool( Tool* tool )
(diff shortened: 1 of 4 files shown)


wjwwood:
Looks reasonable to me. I guess you are using this new signal in a plugin? I didn't see it used anywhere in the existing code base.

hdeeken:
@wjwwood: Yep, I need that signal for the [rviz_fps_plugin](https://github.com/uos/rviz_fps_plugin). i'd like to switch the toolbar icon to idicate if the tool is in walk or fly mode. Speaking of the tool... If you find the time, would you mind give that tool a try? It's for navigating rviz like in ego-perspective using WASD and the arrow keys. My colleagues and I find it find it quite useful, maybe we can integrate it in RViz? If you're interested, I'll be glad to create a PR, just let me know. ;)

This GUI bug was found while I was creating two other [tools](https://github.com/uos/rviz_homing_plugin), which are meant to store a home pose for the robot and send it there by just clicking the toolbar. Quite handy when the robot needs to be reset to the same starting point during experiments.

P.S. My ideas to extend the shortkey  are slowy coming to fruition as well. ;)

Issue #811, opened 2014-09-12
TF display !pos.isNaN() && "Invalid vector supplied as parameter" assert


skohlbr:
Starting recently, we get the following segfault every few seconds with our Atlas setup:

<pre>
rviz: /build/buildd/ogre-1.7.4/OgreMain/src/OgreNode.cpp:413: virtual void Ogre::Node::setPosition(const Ogre::Vector3&): Assertion `!pos.isNaN() && "Invalid vector supplied as parameter"' failed.
</pre>

This error happens only if "world" is set as a fixed frame. If set it to a frame belonging to the robot, things work fine. This suggests that something about the transform between world and rest of the robot leads to the crash. Essentially there are two issues here:
- rviz should not segfault as soon as some kind of invalid tf data is received and spit out a cryptic assert that gives the user no indication of what went wrong. Instead it should notify the user with a warning, but possibly keep on running.
- There might or might not be a bug that leads rviz to crash even if it received "good" tf data (as detailed below I fail to see what is "bad" about our tf data so far).

A bag file that reproduces this crash easily is available for download here: https://drive.google.com/folderview?id=0B1hU91jkd7VwQ2ZQdXgzY2V6dFE&usp=sharing
Just playing back the bagfile (without --clock) and having rviz running with fixed frame "world" and a tf display active will crash rviz 1.5 seconds into the bagfile. We changed rviz sources to not assert but spit out an error to console, ignore the transforms and proceed. Apart from the error occuring every few seconds, things appear to run fine. We also use the transform between world and robot for various other parts in our system, and none of those crash. All of this leads me to believe that the observed behavior could be an issue with rviz (or how it uses tf).

A complete backtrace of rviz installed from .debs (Hydro/Precise/AMD64) of the resulting crash is this:

<pre>
rviz: /build/buildd/ogre-1.7.4/OgreMain/src/OgreNode.cpp:413: virtual void Ogre::Node::setPosition(const Ogre::Vector3&): Assertion `!pos.isNaN() && "Invalid vector supplied as parameter"' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff645c4f5 in raise () from /lib/x86_64-linux-gnu/libc.so.6
(gdb) bt
#0  0x00007ffff645c4f5 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff645fc5b in abort () from /lib/x86_64-linux-gnu/libc.so.6
[...]

PR #860, opened 2015-03-09
[WIP] Display Type Vector3Stamped

goretkin:
I am writing this on top of Groovy, but aside from the changes in https://github.com/ros-visualization/rviz/commit/c7b8d44c511d4171b0ae13a6edbc7cc8d8a0c9bf, it should apply to more recent releases.

This is essentially a duplicate of the Wrench code. If I had to visualize a Twist, it would be exactly the Wrench code. Is there a way to avoid this duplication?

Visualizing Wrenches worked fine for me, but for visualizing Vector3, for one iteration `scale_` was NaN, which triggers an Ogre assertion error. I don't know why this happens and why it's different behavior from the Wrench code, but for now I check for NaN.

### DIFF
--- a/plugin_description.xml
+++ b/plugin_description.xml
@@ -150,6 +150,12 @@
     </description>
   </class>
 
+  <class name="rviz/Vector3Stamped" type="rviz::Vector3StampedDisplay" base_class_type="rviz::Display">
+    <description>
+      Displays from geometry_msgs/Vector3Stamped message
+    </description>
+  </class>
+
   <class name="rviz/PublishPoint" type="rviz::PointTool" base_class_type="rviz::Tool">
     <description>
       Allows you to click on a point and publish it as a PointStamped message.
--- a/src/rviz/default_plugin/CMakeLists.txt
+++ b/src/rviz/default_plugin/CMakeLists.txt
@@ -33,6 +33,7 @@ qt4_wrap_cpp(MOC_FILES
   tools/initial_pose_tool.h
   tools/interaction_tool.h
   tools/point_tool.h
+  vector3_display.h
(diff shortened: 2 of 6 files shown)


goretkin:
I wonder if this is related: https://github.com/ros-visualization/rviz/issues/811

wjwwood:
Thanks for the contribution, but as I said in the other issue, I won't be releasing new versions of Groovy. I'd be happy to consider taking this on the `indigo-devel` branch though. Unfortunately I don't have time to do it myself in the near term. I'll keep it on the list though.

goretkin:
I'll make a new pull request against indigo once I install it somewhere so that I can make sure it runs. Thanks!

DaikiMaekawa:
Hey @goretkin, @wjwwood 

I'd like to display geometry_msgs/Vector3 on RViz.
Did you already create and merge new pull request for indigo? If so, how do I have to do?
Actually, I couldn't find your vector3 codes yet.

Thanks for your reply.

goretkin:
Hi @DaikiMaekawa sorry, I never took the time to get ROS indigo running.

There are 4 commits in this PR. I don't know what the merge commits are, they might have been a mistake. However, if you apply the two top commits, you should get the Vector3 code working. You'll need this commit: https://github.com/goretkin/rviz/commit/509503d4753851e181081a9256d7ef947d8fd77d and its parent.

If you can confirm that it's working on Indigo, it would be great if you could make a PR to Indigo!

The question still remains on how to do this without duplication. There should probably be some base code to visualize vectorial quantities, with properties such as scale and color. Then Wrench, Twist, and Vector3 would all rely on that.

PR #862, opened 2015-03-11
Allow force and torque in Wrench to be scaled separately

goretkin:
since force and torque are different quantities, the units of their scalings are different (meters per Newton and meters per Newton-meter, respectively), and so they should really have separate scalings. Practically, the torques in my Wrench are usually much smaller than the forces if I use the same scaling.

I am on Groovy, but it looks like this patch can apply to all releases up to present.

### DIFF
--- a/src/rviz/default_plugin/wrench_display.cpp
+++ b/src/rviz/default_plugin/wrench_display.cpp
@@ -35,10 +35,15 @@ namespace rviz
                                      "0 is fully transparent, 1.0 is fully opaque.",
                                      this, SLOT( updateColorAndAlpha() ));
 
-	scale_property_ =
-            new rviz::FloatProperty( "Arrow Scale", 2.0,
-                                     "arrow scale",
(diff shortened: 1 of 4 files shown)


wjwwood:
Thanks for the pull request, the changes look ok, but I'm going to release any more Groovy versions of rviz. Can you please open the pull request against the `indigo-devel` branch?

chapulina:
+1

wjwwood:
I'm going to merge this into groovy-devel, but I will not be releasing a new version.

goretkin:
And we'll have to forward-port it to latest versions, right?

wjwwood:
Yeah, #870 is for indigo, I'll backport to hydro when I do the next release.

wjwwood:
I also back ported @chapulina's commit fixing the indention up: d24eadc

goretkin:
Thank you so much!

PR #866, opened 2015-03-19
Potential fix for DELETEALL marker action

### LINKED ISSUE #865: Delete all rviz markers
On the tutorial ( http://wiki.ros.org/rviz/Tutorials/Ma... ) and docs for the visualisation markers, there's supposed to be an option for action "DELETEALL" which would remove all markers. I've tried this and it has not deleted any markers. Is this function actually in effect? Is there anything else I need to specify apart from marker.action?

wjwwood:
See:

- http://answers.ros.org/question/205322/delete-all-rviz-markers/
- https://github.com/ros-visualization/rviz/issues/865

Fixes #865

### DIFF
--- a/src/rviz/default_plugin/marker_display.cpp
+++ b/src/rviz/default_plugin/marker_display.cpp
@@ -206,10 +206,16 @@ void MarkerDisplay::deleteMarkersInNamespace( const std::string& ns )
 
 void MarkerDisplay::deleteAllMarkers()
 {
+  std::vector<MarkerID> to_delete;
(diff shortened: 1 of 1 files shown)


mikeferguson:
@wjwwood I think #874 fixes the real issue, delete all works fine without this patch as long as you have a valid frame_id

wjwwood:
@mikeferguson I agree, though I *think* I might still merge this one. The patch still makes sense to me.

mikeferguson:
That sounds good too.

PR #867, opened 2015-03-20
feature get render panel

otim:
I needed this for building a custom rviz rqt plugin.


otim:
I just saw that VisualizationManager has this function. Closing this.

PR #871, opened 2015-04-02
Updated version of pull request #864

caguero:
This pull request is targeted to indigo-devel. #864


wjwwood:
I reverted this based on @jbohren's feedback.

PR #873, opened 2015-04-03
ogre_helpers: Fix STL import and detect binary files masquerading as ASCII ones

jbohren:
The RViz STL importer can only read binary files.

ASCII STL files are supposed to be designated by the first five chars of the file being "solid"

Lots of binary STL files out there don't care for your standards and start with "solid" anyway. I think the SolidWorks STL exporter does this.

This warns people if their STL files are malformed, and reports errors if they're not binary files. If it's determined that they're not valid binary files, where valid means "won't crash rviz" then it returns false and reports the problem.

Note that this takes more care to consider broken STL files than Assimp does. Assimp will reject binary files which start with "solid". If we use the same test, many ROS urdf models will stop working.

This also replaces the fragile c-style buffer heap allocation with a `std::vector` which won't leak.

This patch is not API compatible.
This patch is most likely not ABI compatible.

https://en.wikipedia.org/wiki/STL_(file_format)#ASCII_STL

### DIFF
--- a/src/rviz/ogre_helpers/stl_loader.cpp
+++ b/src/rviz/ogre_helpers/stl_loader.cpp
@@ -73,7 +73,9 @@ bool STLLoader::load(const std::string& path)
   long fileSize = ftell( input );
   fseek( input, 0, SEEK_SET );
 
-  uint8_t* buffer = new uint8_t[ fileSize ];
+  std::vector<uint8_t> buffer_vec(fileSize);
+  uint8_t* buffer = &buffer_vec[0];
+
   long num_bytes_read = fread( buffer, fileSize, 1, input );
   if( num_bytes_read != fileSize )
   {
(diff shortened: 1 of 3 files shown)


caguero:
Thanks for the patch and the detailed description. We'll test it soon.

src/rviz/ogre_helpers/stl_loader.cpp
```diff
@@ -82,15 +84,50 @@ bool STLLoader::load(const std::string& path)
...
+    ROS_WARN_STREAM("Your STL file \""<<origin<<"\" is malformed. It starts with the word \"solid\", indicating that it's an ASCII STL file, but it does not contain the word \"endsolid\" so it is either a malformed ASCII STL file or it is actually a binary STL file. Proceeding assuming binary format.");
+  }
+
+  // make sure there's enough data for a binary STL header and triangle count
+  if(num_bytes <= 84) {
+    ROS_ERROR_STREAM("Your STL file \""<<origin<<"\" is malformed. It appears to be a binary file but does not contain enough data for the 80 byte header and 16-bit integer triangle count");
+    return false;
+  }
+
+  // we're pretty sure this is a malformed binary file at this point, one last check to make sure it won't segfault
```
caguero: This comment claims that this is a malformed binary file at this point. I think we cannot assume this here, all the checks are good so far.

src/rviz/ogre_helpers/stl_loader.cpp
```diff
@@ -82,15 +84,50 @@ bool STLLoader::load(const std::string& path)
...
-  bool success = load(buffer);
-  delete [] buffer;
+  return this->load(buffer, num_bytes_read, path);
+}
 
-  return success;
+bool STLLoader::load(uint8_t* buffer, const size_t num_bytes, const std::string& origin)
+{
+  // check for ascii since we can only load binary types with this class
+  std::string buffer_str = std::string((char*)buffer,num_bytes);
```
caguero: I'd suggest to replace this casting by a <reinterpret_cast>.

src/rviz/ogre_helpers/stl_loader.cpp
```diff
@@ -82,15 +84,50 @@ bool STLLoader::load(const std::string& path)
...
+
+  if(buffer_str.substr(0,5)  == std::string("solid")) 
+  {
+    // file says that it is ascii, but why should we trust it?
+
+    // check for "endsolid" as well
+    if(buffer_str.find("endsolid",5) != std::string::npos) {
```
caguero: Style: Move the `{` to the next line for consistency.

src/rviz/ogre_helpers/stl_loader.cpp
```diff
@@ -82,15 +84,50 @@ bool STLLoader::load(const std::string& path)
...
+  // make sure there's enough data for a binary STL header and triangle count
+  if(num_bytes <= 84) {
+    ROS_ERROR_STREAM("Your STL file \""<<origin<<"\" is malformed. It appears to be a binary file but does not contain enough data for the 80 byte header and 16-bit integer triangle count");
+    return false;
+  }
+
+  // we're pretty sure this is a malformed binary file at this point, one last check to make sure it won't segfault
+  // get the number of triangles
+  unsigned int num_triangles = *(uint16_t*)(buffer + 80);
+  if(num_bytes < 84+num_triangles*50) {
```
caguero: Style: Move the { to the next line for consistency

caguero:
@wjwwood, can we write a simple test that tries to load a few STL files:

* Invalid binary STL file with size < 84 bytes.
* Invalid binary STL file with size >= 84 bytes.
* Valid ascii STL file.

src/rviz/ogre_helpers/stl_loader.cpp
```diff
@@ -82,15 +84,50 @@ bool STLLoader::load(const std::string& path)
...
+      ROS_ERROR_STREAM("Your STL file \""<<origin<<"\" is malformed. It starts with the word \"solid\" and also contains the word \"endsolid\", indicating that it's an ASCII STL file, but RVIZ can only load binary STL files so it will not be loaded. Please convert the file to a binary format.");
+      return false;
+    }
+
+    // chastise the user for malformed files
+    ROS_WARN_STREAM("Your STL file \""<<origin<<"\" is malformed. It starts with the word \"solid\", indicating that it's an ASCII STL file, but it does not contain the word \"endsolid\" so it is either a malformed ASCII STL file or it is actually a binary STL file. Proceeding assuming binary format.");
+  }
+
+  // make sure there's enough data for a binary STL header and triangle count
+  if(num_bytes <= 84) {
```
caguero: Style: Move the { to the next line for consistency.

wjwwood:
@jbohren Thanks for taking a second swing at this, it all makes more sense now, we'll get it merged asap.

@caguero Yes, we can, I'll have to take a look at what/how we should approach a test like that.

wjwwood:
@caguero I re-enabled the tests in `rviz` in this pull request: #875

Maybe when we both have some time this week we can put together a test.

caguero:
Sounds good.

wjwwood:
@jbohren I'm going to close this in favor of #878. Please have a look at it when you get a chance.

It is also against `indigo-devel`, but I've also marked it with the `needs_back_port` label, so I'll back port it to Hydro next time I do a Hydro release.

PR #874, opened 2015-04-03
deleting markers does not require transform

mikeferguson:
Currently if you send a DELETE or DELETE_ALL with no `header.frame_id`, or an untransformable `header.frame_id`, the deletion does not happen. There is no reason to need the `frame_id` for delete, we should go ahead and do the delete anyways (MoveIt uses a similar logic in the tf filter fail for object insertion/deletion).

Probably also fixes what @JessChL was seeing in #865

### DIFF
--- a/src/rviz/default_plugin/marker_display.cpp
+++ b/src/rviz/default_plugin/marker_display.cpp
@@ -250,6 +250,11 @@ void MarkerDisplay::incomingMarker( const visualization_msgs::Marker::ConstPtr& 
 void MarkerDisplay::failedMarker(const ros::MessageEvent<visualization_msgs::Marker>& marker_evt, tf::FilterFailureReason reason)
 {
   visualization_msgs::Marker::ConstPtr marker = marker_evt.getConstMessage();
+  if (marker->action == visualization_msgs::Marker::DELETE ||
+      marker->action == 3)  // TODO: visualization_msgs::Marker::DELETEALL when message changes in a future version of ROS
+  {
+    return this->processMessage(marker);
+  }
   std::string authority = marker_evt.getPublisherName();
   std::string error = context_->getFrameManager()->discoverFailureReason(marker->header.frame_id, marker->header.stamp, authority, reason);
   setMarkerStatus(MarkerID(marker->ns, marker->id), StatusProperty::Error, error);


src/rviz/default_plugin/marker_display.cpp
```diff
@@ -250,9 +250,21 @@ void MarkerDisplay::incomingMarker( const visualization_msgs::Marker::ConstPtr&
...
+  else if (marker->action == 3)  // TODO: visualization_msgs::Marker::DELETEALL when message changes in a future version of ROS
+  {
+    deleteAllMarkers();
+  }
+  else
+  {
+    std::string authority = marker_evt.getPublisherName();
+    std::string error = context_->getFrameManager()->discoverFailureReason(marker->header.frame_id, marker->header.stamp, authority, reason);
+    setMarkerStatus(MarkerID(marker->ns, marker->id), StatusProperty::Error, error);
+  }
```
wjwwood: Would it be better to pass the marker message to `MarkerDisplay::processMessage()` if the action is `DELETE` or `DELETEALL` (3)? That was there's no need to duplicate logic. I was imagining something like this:

```c++
if (marker->action == visualization_msgs::Marker::DELETE ||
    marker->action == 3)  // TODO: visualization_msgs::Marker::DELETEALL when message changes in a future version of ROS
{
  // Process the marker anyways since it is a delete or deleteall and doesn't need to be transformed.
  return this->processMessage(marker);
}
std::string authority = marker_evt.getPublisherName();
std::string error = context_->getFrameManager()->discoverFailureReason(marker->header.frame_id, marker->header.stamp, authority, reason);
setMarkerStatus(MarkerID(marker->ns, marker->id), StatusProperty::Error, error);
```

wjwwood:
@mikeferguson Thanks for looking at this, I think you're right and it could be a fix. @JessChL can you try this out?

src/rviz/default_plugin/marker_display.cpp
mikeferguson: That probably is cleaner -- I was just replicating the design I had seen a while ago in MoveIt. I'll update the PR shortly

mikeferguson:
Updated the PR to reduce code duplication. Tested on my application, works as expected.

wjwwood:
Awesome thanks.

wjwwood:
I'm going to go ahead and merge this, @JessChL, please give feedback on this later if you get a chance.

PR #876, opened 2015-04-06
mesh_resource_marker: re-enabling mesh transparency

### LINKED ISSUE #751: COLLADA mesh's material is not visualized correctly
when a collada file is used as marker, it's not visualized correctly. Always the color
of the model is rendered as white. However if we use texture mapping in a collada file, rviz 
can visualize it correctly.

On the other hand, if that file is used as robot_description, RobotModel can visualize it correctly.

https://github.com/garaemon/rviz_collada_marker
this is the package includes the test code and you can test it by `roslaunch rviz_collada_marker test.launch`

screenshot from 2014-04-24 21 38 22

jbohren:
@hersh This highlights the problem you reported in #854 caused by the regression in #752 which intended to fix #751

The issue with the new blending technique is that the base material is always set to be 100% opaque. This patch just initializes the base material based on the marker RGBA values.

### DIFF
--- a/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
+++ b/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
@@ -95,6 +95,27 @@ void MeshResourceMarker::onNewMessage(const MarkerConstPtr& old_message, const M
 
   scene_node_->setVisible(false);
 
+  // Get the color information from the message
+  float r = new_message->color.r;
+  float g = new_message->color.g;
+  float b = new_message->color.b;
+  float a = new_message->color.a;
+
+  Ogre::SceneBlendType blending;
+  bool depth_write;
+
+  if (a < 0.9998)
+  {
+    blending = Ogre::SBT_TRANSPARENT_ALPHA;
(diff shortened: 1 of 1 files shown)


jbohren:
@hersh here's what the mesh marker test looks like with this patch:

mesh_marker_test

jbohren:
Based on these results, it looks like it still needs some additional logic to ignore the color information. Shouldn't be too hard to fix.

hersh:
I'm not sure what you mean by "ignore the color information", but the only problem I see is that the alpha=0 rows have a single copy of the original mesh in them.

Thanks for working on this, and including the pictures.  I'm really busy right now, but there's always time for looking at a picture. :)

jbohren:
> I'm not sure what you mean by "ignore the color information", but the only problem I see is that the alpha=0 rows have a single copy of the original mesh in them.

Well the tinted meshes appear "washed out" compared to the correct image you posted on #854.

hersh:
I see, the colorization swamps the native mesh textures, true.

jbohren:
This is superseded by #893

PR #886, opened 2015-05-08
also process empty pointclouds

v4hn:
Otherwise, given a stream of clouds with some of them empty, the last
non-empty message will still be displayed until a the next non-empty cloud
comes in.

### DIFF
--- a/src/rviz/default_plugin/point_cloud_common.cpp
+++ b/src/rviz/default_plugin/point_cloud_common.cpp
@@ -917,11 +917,6 @@ void PointCloudCommon::addMessage(const sensor_msgs::PointCloudConstPtr& cloud)
 
 void PointCloudCommon::addMessage(const sensor_msgs::PointCloud2ConstPtr& cloud)
 {
-  if (cloud->width * cloud->height == 0)
-  {
-    return;
-  }
-
   processMessage(cloud);
 }
 


tlind:
+1 Interestingly, I just wanted to submit a [similar pull request](https://github.com/srl-freiburg/rviz/commit/c2b922282b00c4c44b68f595ab12d341a72b6796) for this issue which I had fixed in our fork a few weeks ago. Our motivation is that we are using clouds to display points inside a safety zone of the robot. If no points lie in that zone, we want the previously displayed points to be cleared by an empty PointCloud2 message.

Note: The LaserScan display internally also uses the PointCloud2 display. For this to also work with empty LaserScan messages, [this pull request](https://github.com/ros-perception/laser_geometry/pull/13) in laser_geometry must also be merged, otherwise RViz crashes with an exception even before reaching the `return;` statement removed by 727e0de.

wjwwood:
Looks good to me, thanks guys.

v4hn:
On Sun, May 17, 2015 at 05:37:20PM -0700, Timm Linder wrote:
> Our motivation is that we are using clouds to display points inside a safety zone of the robot.
> If no points lie in that zone, we want the previously displayed points to be cleared by an empty PointCloud2 message.

That's what we made our students do. And that's how we discovered the bug. :)

PR #887, opened 2015-05-08
find_package python version detected by catkin

v4hn:
Otherwise it might happen that catkin (and the rest of the workspace) uses 2.x
and rviz detects & tries to use 3.x. This can produce some nasty collisions.

See rospack, roslz4, qt_gui_cpp and others for similar invokation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,8 +91,6 @@ endif()
 
 find_package(OpenGL REQUIRED)
 
-find_package(PythonLibs REQUIRED)
-
 find_package(Qt4 REQUIRED COMPONENTS QtCore QtGui QtOpenGL)
 
 find_package(catkin REQUIRED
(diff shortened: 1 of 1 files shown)


wjwwood:
Does this need conditional logic for when `PYTHON_VERSION_MAJOR` and `PYTHON_VERSION_MINOR` are not set? I ran a simple cmake project with the added line from this pr and got this:

```
CMake Error at /usr/local/Cellar/cmake/3.2.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find PythonLibs (missing: PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS)
  (Required is at least version ".")
  /usr/local/Cellar/cmake/3.2.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:374 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.2.1/share/cmake/Modules/FindPythonLibs.cmake:205 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:4 (find_package)
```

v4hn:
It *would* need additional logic if RViz weren't a catkin project.

These variables are set after catkin is find_package'd.
(which is also why I moved the find_package further down in the CMakeLists.

wjwwood:
Cool, works for me.

PR #890, opened 2015-05-19
Added "/" to avoid the following warning when using namespaces:

ferherranz:
Added "/" to avoid the following warning when using namespaces:
Ignoring initial pose in frame "map"; initial poses must be in the global frame, "/map"

It allows multi_robot configurations

### DIFF
--- a/src/rviz/default_plugin/tools/initial_pose_tool.cpp
+++ b/src/rviz/default_plugin/tools/initial_pose_tool.cpp
@@ -62,7 +62,7 @@ void InitialPoseTool::updateTopic()
 
 void InitialPoseTool::onPoseSet(double x, double y, double theta)
 {
-  std::string fixed_frame = context_->getFixedFrame().toStdString();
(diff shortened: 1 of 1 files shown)


trainman419:
This seems like a hack that makes this work for your particular scenario, but likely breaks some other behavior.

Can you do the same thing by setting your fixed frame to `/map` when using rviz, instead of making a code change?

wjwwood:
Thanks for the suggestion and the pull request.

The correct thing to put there would be `/map`; implicitly adding a `/` is not feasible in my opinion, because it would allow a user to intentionally pass a relative frame name and have it unintentionally interpreted as a global frame, potentially not producing an error message or warning message, but instead a logical error, which is hard to notice and track down.

If you think that this is really needed, then please expand more on why you think that, including arguments about how it would be used by different users and rational on how your fix addresses the different use cases. For example, what would happen with your current pull request if `context_->getFixedFrame().toStdString()` already had a leading `/`?

I'm going to close this, if you update it with more information I'm happy to reopen the pull request.

PR #893, opened 2015-06-04
Fix mesh alpha tinting

loganE:
Reinstated original form of mesh tinting and removed color pass technique that was introduced in #752. The color pass technique was overriding mesh texture and alpha values. It appears that the issue present in #751 was actually a result of a piece of code that would make a mesh pure white if the message sent rgba values of 0000. I can't pinpoint why, but I estimate that somehow that was erroneously turning collada markers pure white. Tested change with rviz/src/test/mesh_marker_test.cpp and ran several tests with collada markers, using mesh_use_embedded_materials true and false. Transparency and mesh tinting appear to work as desired.

### DIFF
--- a/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
+++ b/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
@@ -82,8 +82,7 @@ void MeshResourceMarker::reset()
     }
   }
   materials_.clear();
-  // the actual passes are deleted by the material
-  color_tint_passes_.clear();
+
 }
 
 void MeshResourceMarker::onNewMessage(const MarkerConstPtr& old_message, const MarkerConstPtr& new_message)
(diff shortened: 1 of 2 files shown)


jbohren:
@loganE Can you post screenshots of the two tests?

loganE:
Here are attached images of a mesh marker test, and a test involved collada markers with different colors and transparencies. 
collada

meshtransp

wjwwood:
@loganE thanks for the contribution, I've reviewed the code and it seems to make sense to me.

@jbohren I'm guessing you '+1' this pr since you closed yours?

@hersh if you have a spare moment, can you look at the results of this pr and give your opinion.

If @hersh and @jbohren give positive feedback, I'll merge this and work it into the next release of indigo and jade.

hersh:
@loganE thanks for working on this.

I ran it myself and got a matching image to the white/red/pink image above.  However looking at the screenshot from the original version:

image

you can see there is a normal-colored (white with black decorations) copy of the mesh off to the left side.  That is the rgba=0,0,0,0 version.  That was added as a special case as @loganE  mentioned.

The special case is to make mesh markers visible which have uninitialized color values.

When I originally added the mesh-tinting code, mesh markers completely ignored the color value.  Therefore everyone who used mesh markers just left the color unset (0,0,0,0).  I added the special case to make all those uses still work.

I don't actually know what code may be out there counting on the special case.  Maybe there are none.  Maybe a lot.

If it's as easy as just adding back a single "if" block like "if color==0,0,0,0 { color = 1,1,1,1; }" then I would say please do it.  If that breaks the collada colors, then I'm not sure what to do.  Probably then we try it without the special case and see if anyone complains. :-/

wjwwood:
@hersh I can make an announcement about it on ros-users, that way at least some affected people could be notified.

jbohren:
@wjwwood I'll defer to @hersh on this one, but this PR does satisfy my mesh-transparency use cases.

loganE:
@wjwwood I have no objections to putting that transparency fail-safe code back in and giving it a quick test. If it fails we'll leave it out. If it works I'll update the PR. I'll look at it first thing tomorrow.

wjwwood:
Ok, thanks @loganE.

loganE:
@wjwwood I reinserted the line of code that forces opacity and color on mesh markers with nul rgba values. I have attached images from my testing that shows all markers appear as desired.
meshmarknew
colladanew

wjwwood:
Looks like we have an additional pr2 base in the first image, which seems to line up with the original version. @hersh I still haven't had the time to really figure out what we should be looking for here, if you have time I'll defer to your opinion on the results.

hersh:
I am happy with the test image showing the additional PR2 base (the one with color=0,0,0,0).  The other image I have just the concern that in #752 the new image looked like this:

image

Since I don't know what was done to generate the two images, I'm not sure if the new one (with all-black, all-red, and all-blue) is correct but coming from a different source, or if it is incorrect.

@loganE do you know if you are using the same marker source for the black/red/blue image as @brindza used on #752?

loganE:
@hersh I was using @brindza's repo for collada markers, however I had been tinkering with the meshes themselves, hence the difference. I just recloned the original, unmodified version and the textures are indeed present on the bases. There is no texture specified in the arrow dae file, so it is only the color specified by the rgba that is displayed, not the tri-color as shown above. I'm not sure if this is due to a change in the tri-arrow dae, or if this is working as-advertised.

loganE:
Inadvertently left out depth-write argument in the previously added rgba "forced color and opacity" if statement. Created a slight anomaly with those objects. Tri-arrow object and all others were unaffected. Amended the code.

hersh:
@loganE if you get the tri-colored arrows and the textured PR2 bases, then this change looks good to me.

wjwwood:
Great I'll merge it as a bug fix and mark it for forward porting to Jade.

wjwwood:
Thanks for the work guys!

Issue #895, opened 2015-06-25
librviz Python Bindings - Segmentation Fault when using method "setStatusBar()"


futureneer:
Hello,
I am getting a segmentation fault when trying to hide the status bar in the rviz VisualizationFrame widget.  Here is a minimal example which produces the segfault:

```python
#! /usr/bin/env python
import sys
import rviz
from python_qt_binding.QtGui import *
from python_qt_binding.QtCore import *


class TestWidget(QWidget):
    def __init__(self):
        QWidget.__init__(self)

        # config_path = '~/.rviz/default.rviz'
        config_path = ''

        self.frame= rviz.VisualizationFrame()
        self.frame.setSplashPath('')
        self.frame.setMenuBar(None)
        self.frame.setStatusBar(None)
        self.frame.initialize()

        reader = rviz.YamlConfigReader()
        config = rviz.Config()
        reader.readFile(config,config_path)
        self.frame.load(config)

        self.manager = self.frame.getManager()

        layout = QVBoxLayout()
        layout.addWidget( self.frame )

        top_button = QPushButton( "Top View" )
        layout.addWidget( top_button )

        self.setLayout(layout)

if __name__ == '__main__':
    app = QApplication( sys.argv )
    test_widget = TestWidget()
    test_widget.resize( 500, 500 )
    test_widget.show()
    app.exec_()
```

By commenting out the line:
```python
    # self.frame.setStatusBar(None)
```

the code runs just fine.

Let me know what you think, and thanks!!

Kel

wjwwood:
Thanks for the detailed report. Nothing comes to mind off the top of my head. Have you tried sending an empty string instead of None?

I'd have to reproduce this and look through the tracebacks with gdb. I'm not sure when I'll have time to do that.

futureneer:
Hi, thanks for the reply. I did as you said, and it says it doesn't take strings. Incidentally, the setHideButtonVisibility method also causes a segfault as well. Not sure if they are related.

I guess I could create an empty QStatusBar and put that in, since that's the type.  I'll try it.

futureneer:
No dice, the empty status bar doesn't do it.  Its not a critical thing, I'm just trying to create a clean interface as possible, and though I could free up some real estate by getting rid of the status bar.  I'll deal if its a pain.

wjwwood:
Looks like it might be related to the function which updates the fps:

```
[...]

PR #896, opened 2015-06-25
override setStatusBar to avoid trying to use it when it is changed

wjwwood:
Addresses #895


wjwwood:
Ok, I've update this one and I think it's fixed now.

Issue #877, opened 2015-04-14
pointcloud2: support xyzrgba type or display an error that it's not supported


jbohren:
Currently if you try to display a pointcloud with xyzrgba datatype, it won't detect that it can draw the color information.

wjwwood:
rgba is now supported with #898

Issue #885, opened 2015-05-05
Pointclouds with RGBA fields cannot be colored using RGB


calderpg:
Pointclouds created using PCL::PointXYXRGBL and some other PCL types have a field named "rgba", rather than the "rgb" value expected by Rviz. (For alignment issues, it's always rgba, except for the apparently special case of PCL::PointXYZRGB which reports it as "rgb"). Without the "rgb" field, Rviz does not provide a RGB color transformer, so you cannot display the real colors of the points in Rviz.

As far as I can tell, the problematic check is here:

https://github.com/ros-visualization/rviz/blob/indigo-devel/src/rviz/default_plugin/point_cloud_common.cpp#191

Instead of `name == "rgb"` this should be `name == "rgb" || name == "rgba"`.

Obviously, this doesn't actually add support for drawing transparency, but rendering the RGB color for all poinclouds that support it is very useful.

wjwwood:
Thanks for the report. If you think the above solution would work, please make a pull request and perhaps provide a bag file so it can easily be tested.

Otherwise you can wait for someone else to get to it, but it might take a while.

wjwwood:
Closed by #885

PR #898, opened 2015-07-22
Issue #885: display Pointcloud2 with rgba field and handle per pixel alpha.

ThomasMoinel:
@calderpg point in #885 a select handler class which wasn't the right place to look but has a bug too. So I fix it with commit f331536 for rgb and add a case for rgba.

2db2f79 fix the display of the pointcloud by handling rgba directly in rgb RGB8Transformer.
I duplicate the _for_ to avoid garbage value in the first octet (alpha) of rgb to be used as alpha when alpha property or pointcloud is under 1.

Finally, 6cd4607 enable a per pixel alpha for rgba. But this seems a bit odd way to do. I'll be glad to hear comments or suggestions.

## Test
To test setup a PointCloud2 rviz pluggin with size property set to at least 0.2. 
There will be 2 points in (0,0,0) in red alpha=255 and (0.5,0,0) in green alpha=100. 
Add a Selection Panel to test f331536.

```
cat << EOF | rostopic pub /points sensor_msgs/PointCloud2
header: 
  seq: 0
  stamp:
    secs: 0
    nsecs: 0
  frame_id: map
height: 1
width: 2
fields: 
  - 
    name: x
    offset: 0
    datatype: 7
    count: 1
  - 
    name: y
    offset: 4
    datatype: 7
    count: 1
  - 
    name: z
    offset: 8
    datatype: 7
    count: 1
  - 
    name: rgba
    offset: 12
    datatype: 7
    count: 1
is_bigendian: False
point_step: 16
row_step: 32
#       x              y             z             b    g    r    a
data: [ 0, 0, 0,  0,   0, 0, 0, 0,   0, 0, 0, 0,    0,   0, 255, 255,
        0, 0, 0, 63,   0, 0, 0, 0,   0, 0, 0, 0,    0, 255,   0, 100]
is_dense: False
---
EOF
```
Test with name field to 'rgb' and 'rgba'.

Without these patches, the selection of both rgb and rgba displays a black color in the selection panel. With this fix it now displays the right color plus the alpha value normalize in [0,1], even for rgb.

Without theses patches, rgba point will be white and the alpha does nothing. With this fix, the rgba point will be in green with transparency according to alpha value.

Fill free to test other color/alpha values and play with the Alpha property value of PointCloud2 properties.

### DIFF
--- a/src/rviz/default_plugin/point_cloud_common.cpp
+++ b/src/rviz/default_plugin/point_cloud_common.cpp
@@ -188,17 +188,21 @@ void PointCloudSelectionHandler::createProperties( const Picked& obj, Property* 
           {
             continue;
           }
-          if( name == "rgb" )
+          if( name == "rgb" || name == "rgba")
           {
-            uint32_t val = valueFromCloud<uint32_t>( message, f.offset, f.datatype, message->point_step, index );
+            float float_val = valueFromCloud<float>( message, f.offset, f.datatype, message->point_step, index );
+            // Convertion hack because rgb are stored int float (datatype=7) and valueFromCloud can't cast float to uint32_t
+            uint32_t val = *((uint32_t*) &float_val);
             ColorProperty* prop = new ColorProperty( QString( "%1: %2" ).arg( field ).arg( QString::fromStdString( name )),
-                                                     QColor( val >> 16, (val >> 8) & 0xff, val & 0xff ), "", cat );
+                                                     QColor( (val >> 16) & 0xff, (val >> 8) & 0xff, val & 0xff), "", cat );
             prop->setReadOnly( true );
+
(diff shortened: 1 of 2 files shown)


ThomasMoinel:
It also fix #877.

wjwwood:
Thanks @ThomasMoinel, after my initial review I think this change set looks good. I need to spend some more time thinking about https://github.com/ros-visualization/rviz/commit/6cd4607a9ecd927260d7f9a9e99c72fdd20326a3 and I need to try it out on my machine. I'll get to that asap.

wjwwood:
Also, your commits seem to not be associated with your GitHub account. You might want to make sure you have that email address associated with your GitHub account.

ThomasMoinel:
Thanks for noticing the missing email association. I had it to my github account.

wjwwood:
Seems to work for me. Thanks!

PR #907, opened 2015-08-06
remove shortkeys from shortkey_to_tool_map_

### LINKED ISSUE #880: ToolManager: shortkey_to_tool_map_ not updated when removing tools
I'm part of a DRC team, and we have our own rviz fork. Since we have our own tools that use librviz, we don't use the tools that are initially inserted when you create a new visualization manager. So, we do 

```
manager_ = new rviz::VisualizationManager( render_panel_ );
manager_->getToolManager()->removeAll();
```

and then we add our own tools. I have just rebased our fork, and found that the following line caused issues:

https://github.com/ros-visualization/rviz/blame/indigo-devel/src/rviz/tool_manager.cpp#L140

It will insert a new uninitialized Tool* into  shortkey_to_tool_map_ if it doesn't have that key in the map already, and causes the application to segfault a few lines after that when trying to set it as the current tool.

I'd suggest changing that line to the following:

```
Tool* tool = NULL;
  // need to check if it exists in the map, otherwise it will create an uninitialized Tool* with the event->key() inside shortkey_to_tool_map_ and it will segfault later on when trying to access it
  if( shortkey_to_tool_map_.find(event->key()) != shortkey_to_tool_map_.end() )
    tool = shortkey_to_tool_map_[ event->key() ];
```

Now, that by itself doesn't solve the issue that I'm having: shortkey_to_tool_map_ is never updated outside AddTool. After I remove all tools, shortkey_to_tool_map_ will retain all keys and invalid Tool*, causing it to crash.

hdeeken:
this should fix #880

### DIFF
--- a/src/rviz/tool_manager.cpp
+++ b/src/rviz/tool_manager.cpp
@@ -137,7 +137,12 @@ void ToolManager::handleChar( QKeyEvent* event, RenderPanel* panel )
   }
 
   // check if the incoming key triggers the activation of another tool
-  Tool* tool = shortkey_to_tool_map_[ event->key() ];
+  Tool* tool = NULL;
+  if( shortkey_to_tool_map_.find(event->key()) != shortkey_to_tool_map_.end() )
+  {
+    tool = shortkey_to_tool_map_[ event->key() ];
+  }
+
   if( tool )
   {
     // if there is a incoming tool check if it matches the current tool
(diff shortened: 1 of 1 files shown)


wjwwood:
@hdeeken, if you get a chance please try this out w.r.t. your issue #880. I'm gong to go ahead and merge this since I'm pretty sure it addresses the problem.

PR #909, opened 2015-08-17
Enable compilation with Qt5 (conditional)

simonschmeisser:
This PR adds the option to compile rviz against qt5. I want to embed it into a Qt5 application and thus needed this. I also ported it to using automoc. The changes are fairly minimal, but I can also do something even less invasive and go back to the manual moc if preferred.

Known problems:
- There is no way to distinguish between plugins compiled for qt4 and plugins compiled for qt5, loading qt4 ones will crash rviz

This is work in progress, just want to get feedback early on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,17 @@
-language: cpp
+sudo: required
+dist: trusty
+language: generic
 compiler:
   - gcc
-  - clang
 install:
-  - export CI_ROS_DISTRO=hydro
+  - export CI_ROS_DISTRO=indigo
   - echo $CI_ROS_DISTRO
 # Add ROS repositories
-  - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu precise main" > /etc/apt/sources.list.d/ros-latest.list'
+  - sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu trusty main" > /etc/apt/sources.list.d/ros-latest.list'
   - wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
   - sudo apt-get update
 # Install and initialize rosdep
-  - sudo apt-get install python-rosdep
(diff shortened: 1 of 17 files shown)


hersh:
This sounds like a fine idea.  Qt5 is the future, right?

On loading plugins, how about adding a check like this?
```
ldd ~/rviz-ws/devel/lib/libdefault_plugin.so | grep libQtGui.so.4
```
That results in:
```
libQtGui.so.4 => /usr/lib/x86_64-linux-gnu/libQtGui.so.4 (0x00007f1a3b6bf000)
```
for me.  I don't have any Qt5 plugins handy, but my assumption is that would not be there, and a corresponding string with Qt<something>5 would be there.

As a first pass you could shell out to the "ldd" command.  Perhaps there is a library giving access to the capabilities of ldd which you could move to in the future.  Of course you may also have to dig into the ROS pluginlib code to find the library files before they get loaded.

simonschmeisser:
Known issues so far:
- should there be a splash screen?
- disabled display with panel will disable resizing of containing dock panel
- QWidget::paintEngine: Should no longer be called

wjwwood:
I've been looking into this in more detail, thanks again @simonschmeisser for giving us a head start!

For anyone using OS X, I had to apply this fix with the Homebrew installed qt5:

https://github.com/Homebrew/homebrew/issues/29938#issuecomment-104722532

wjwwood:
@simonschmeisser I've made some progress on my branch `wjwwood-qt5` which is based on this PR.

> should there be a splash screen?

I fixed the splash screen.

> disabled display with panel will disable resizing of containing dock panel

Can you explain this one a little more? I was able to enable/disable the display panel and still resize things.

wjwwood:
I'm going to close this in favor of https://github.com/ros-visualization/rviz/pull/980. I did my best to preserve attribution for you @simonschmeisser, thanks for getting us started!

PR #914, opened 2015-09-01
disable Display if associated Panel becomes invisible

v4hn:
Otherwise the state between Panel & Display becomes inconsistent.

Fixed symptom: When loading a configuration that contains a disabled
CameraDisplay, during the configuration of the panel(the camera render
widget), the panel is set visible for a very short period of time. Because of
the missing logic, the CameraDisplay is enabled together with the panel, but
the Display remains enabled after the Panel is set invisible. One ends up with
an enabled
(and subscribed) CameraDisplay without the corresponding RenderWidget, even so
the configuration specified that the Display is _not_ enabled.

### DIFF
--- a/src/rviz/display.cpp
+++ b/src/rviz/display.cpp
@@ -389,6 +389,10 @@ void Display::associatedPanelVisibilityChange( bool visible )
   {
     setEnabled( true );
   }
+  else
+  {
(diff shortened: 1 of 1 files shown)


wjwwood:
Makes sense to me.

PR #915, opened 2015-09-04
Smooth updates for map display plugin

stonier:
Map displays previously only updated when receiving a message. This means that
if your fixed frame was base_link, the costmaps would not move appropriately
around the robot unless a message was received in order to update the transform
that should be applied to the scene. For global costmaps, this is a slow
update and for static maps, this never happened.

This fixes that by hooking into rviz' periodic call to continuously update the
transform to be applied to the scene.

### DIFF
--- a/src/rviz/default_plugin/map_display.cpp
+++ b/src/rviz/default_plugin/map_display.cpp
@@ -706,6 +706,10 @@ void MapDisplay::setTopic( const QString &topic, const QString &datatype )
   topic_property_->setString( topic );
 }
 
+void MapDisplay::update( float wall_dt, float ros_dt ) {
+  transformMap();
+}
+
 } // namespace rviz
 
 #include <pluginlib/class_list_macros.h>
--- a/src/rviz/default_plugin/map_display.h
+++ b/src/rviz/default_plugin/map_display.h
@@ -104,6 +104,7 @@ protected:
 
   virtual void subscribe();
   virtual void unsubscribe();
+  virtual void update( float wall_dt, float ros_dt );
 
   /** @brief Copy msg into current_map_ and call showMap(). */ 
   void incomingMap(const nav_msgs::OccupancyGrid::ConstPtr& msg);


wjwwood:
Thanks @stonier, this makes total sense to me. I'd like to run it myself though before merging. I'll try to do that asap.

stonier:
Can easily test with turtlebot stage:

```
sudo apt-get install ros-indigo-turtlebot-stage
roslaunch turtlebot_stage  turtlebot_in_stage.launch --screen
```

You'll notice some other elements could use this kind of feature as well. Pose for example. There it is the move base goal pose which only gets published once, and doesn't update beyond that.

wjwwood:
@stonier FYI, on a fresh system when I installed `ros-indigo-turtlebot-stage` and run that command I get this:

```
% roslaunch turtlebot_stage  turtlebot_in_stage.launch --screen
... logging to /home/william/.ros/log/7fded784-5b12-11e5-b6d0-001c42833f2f/roslaunch-ubuntu-6013.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

Traceback (most recent call last):
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/__init__.py", line 298, in main
    p.start()
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/parent.py", line 260, in start
    self._start_infrastructure()
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/parent.py", line 209, in _start_infrastructure
    self._load_config()
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/parent.py", line 124, in _load_config
    roslaunch_strs=self.roslaunch_strs, verbose=self.verbose)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/config.py", line 451, in load_config_default
    loader.load(f, config, verbose=verbose)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/xmlloader.py", line 730, in load
    self._load_launch(launch, ros_config, is_core=core, filename=filename, argv=argv, verbose=verbose)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/xmlloader.py", line 702, in _load_launch
    self._recurse_load(ros_config, launch.childNodes, self.root_context, None, is_core, verbose)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/xmlloader.py", line 666, in _recurse_load
    val = self._include_tag(tag, context, ros_config, default_machine, is_core, verbose)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/xmlloader.py", line 95, in call
    return f(*args, **kwds)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/xmlloader.py", line 587, in _include_tag
    inc_filename = self.resolve_args(tag.attributes['file'].value, context)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/xmlloader.py", line 183, in resolve_args
    return substitution_args.resolve_args(args, context=context.resolve_dict, resolve_anon=self.resolve_anon)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/substitution_args.py", line 299, in resolve_args
    resolved = _resolve_args(resolved, context, resolve_anon, commands)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/substitution_args.py", line 312, in _resolve_args
    resolved = commands[command](resolved, a, args, context)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/substitution_args.py", line 137, in _find
    return _find_executable(resolve_without_path, a, [args[0], path], context)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/roslaunch/substitution_args.py", line 173, in _find_executable
    full_path = _get_executable_path(rp.get_path(args[0]), path)
  File "/usr/lib/python2.7/dist-packages/rospkg/rospack.py", line 199, in get_path
    raise ResourceNotFound(name, ros_paths=self._ros_paths)
ResourceNotFound: turtlebot_navigation
ROS path [0]=/opt/ros/indigo/share/ros
ROS path [1]=/opt/ros/indigo/share
ROS path [2]=/opt/ros/indigo/stacks
```

Installing `ros-indigo-turtlebot-navigation` seems to resolve the issue, so you might have a missing dependency somewhere.

wjwwood:
Ok, this works for me.

I'm not opposed to merging this, but as you said there are many other elements that don't support this correctly. The planner's cost cloud, the planner's path, and the goal pose do not update. In general I'd say that putting the fixed frame to be `base_link` isn't a good idea. Can you explain why you want to do this rather than using `map` as the fixed frame and setting the target frame to be `base_link`?

stonier:
The turtlebot stage dependency error is already fixed, just been waiting a move from shadow to public.

On the second point, I'm feeling rather naive now - been using rviz for years but never found that target frame option! And it does explain why I couldn't work out why nobody else had brought this issue to light :neckbeard:

Thanks for the tip.

stonier:
I think it's still a good idea to merge/fix them (as and when people do) . If only for the fact that it is a very broken look for a new user to discover when they change to a moving frame.

wjwwood:
I agree.

Issue #913, opened 2015-08-26
rviz 1.11.8 does not load stl


nxdefiant:
rviz 1.11.8 does not display my stl model, in the console I get:

> [ INFO] [1440539561.574488292]: rviz version 1.11.8
> [ INFO] [1440539561.574592016]: compiled against OGRE version 1.9.0 (Ghadamon)
> [ INFO] [1440539561.722585436]: Stereo is NOT SUPPORTED
> [ INFO] [1440539561.722696805]: OpenGl version: 3 (GLSL 1.3).
> [ERROR] [1440539562.762987357]: The STL file 'package://wild_thumper/meshes/wild_thumper_assembly_corrected.stl' is malformed. According to the binary STL header it should have '23240' triangles, but it has too much data for that to be the case.
> [ERROR] [1440539562.763706699]: Failed to load file [package://wild_thumper/meshes/wild_thumper_assembly_corrected.stl]

..and a few following warnings about the same file. The same stl workes with rviz 1.11.7.
Since the model is not mine, I can not attach it here. It is downloaded from [grabcad.com](https://grabcad.com/library/wild-thumper-correct-dimensions-1) and converted to stl with freecad.

wjwwood:
@caguero Any idea?

wjwwood:
@nxdefiant can you please post the STL version? I believe it will be ok for you to link it here as long as you attribute it to grabcad.com.

Failing that, can you share the STL with me privately? It's going to be very hard to reproduce your problem without the actual STL which fails or without more detailed instruction on how to create the STL.


Do you know if you converted it to an ascii STL or a binary STL using freecad?

nxdefiant:
Here it is: [wild_thumper_assembly_corrected.stl](http://vontaene.de/ros/wild_thumper_assembly_corrected.stl)

Should be binary stl, but I don't think I had the option to choose.

wjwwood:
@nxdefiant thanks a lot for the test file. After comparing it to our test files, I found that one of our new checks had a bug in it where the number of triangles was interpreted as a 16-bit unsigned integer instead of a 32-bit unsigned integer.

https://github.com/ros-visualization/rviz/pull/920 should fix this and adds a test case to cover it.

I'll try to get this fix out in a release at the beginning of next week.

nxdefiant:
Confirmed, thanks

Liangconghui:
hi, wjwwood.
 I encountered the same problem.
Could you let me know how to solve this problem.
[...]

PR #920, opened 2015-09-19
fix reading of number of triangles in STL files

wjwwood:
Fixes: #913


wjwwood:
Back ported to `hydro-devel` in 9fef372

PR #921, opened 2015-09-21
Updated warning message to indicate triangle count is 32bits

Jmeyer1292:
As a follow up to #920, this PR adds a small change to one of the printed messages that displays when a file contains less data than is suggested by its triangle count. The message now correctly indicates that the triangle count is a 32 bit integer.


wjwwood:
Good catch, thanks. I'll merge when CI catches up.

PR #922, opened 2015-09-26
python_bindings: sip: Use CATKIN_PACKAGE_LIB_DESTINATION instead of h…

aballier:
…ardcoded lib.

Fixes build with libdir != lib.
https://bugs.gentoo.org/show_bug.cgi?id=561480


wjwwood:
lgtm

wjwwood:
Thanks for the patch.

PR #927, opened 2015-10-09
build: Use PKG_CONFIG_EXECUTABLE

marv:
... instead of using a hard-coded pkg-config to make cross-compiling
possible where the pkg-config binary is host-prefixed (e.g.
armv7-unknown-linux-pkg-config when cross-compiling for armv7)


wjwwood:
LGTM, thanks for the patch.

PR #928, opened 2015-10-12
Fixed Qt assertions triggered in debug build of Qt.

hersh:
I ran rviz with a debug build of Qt, and it crashed with an assertion failure.  This fixes that failure.

A normal (release) build of Qt has the assertions disabled, so this doesn't happen in normal use.

The assertion fails when any of these values are zero and the code continues.  You end up with a first = 5, last = 4 situation inside QAbstractItemModel::beginRemoveRows(), which triggers an assertion that first <= last.


wjwwood:
lgtm

PR #931, opened 2015-10-16
Enable antialiasing

bchretien:
This PR is an attempt to truly enable antialiasing. The only piece of code that seemed related to it was [here](https://github.com/ros-visualization/rviz/blob/4abbbd9a413ee26942d1c7d838975d58b3cc0434/src/rviz/ogre_helpers/render_system.cpp#L241-L242), but changing the value did not seem to have any effect. I added the entry to the proper dictionary, and used a value of 4 which is quite common (cf. [Gazebo](https://bitbucket.org/osrf/gazebo/src/38fbac52640382b2f7ff749aca3580cca7fd7994/gazebo/rendering/WindowManager.cc?at=default&fileviewer=file-view-default#WindowManager.cc-91)). Talking about Gazebo, they also [added antialiasing to their `Camera` class](https://bitbucket.org/osrf/gazebo/src/38fbac52640382b2f7ff749aca3580cca7fd7994/gazebo/rendering/Camera.cc?at=default&fileviewer=file-view-default#Camera.cc-1234:1253), which for instance shares code similar to [`SelectionManager::setDepthTextureSize`](https://github.com/ros-visualization/rviz/blob/4abbbd9a413ee26942d1c7d838975d58b3cc0434/src/rviz/selection/selection_manager.cpp#L351-L357) for creating `Ogre::RenderTexture` objects. This may not be relevant at all to add FSAA here, but it may be the perfect time to check similar parts of the code where antialiasing could improve the rendering.

As for the previous `"FSAA"` key in the other dictionary, if it is in fact useless, it may be worth removing, but since Gazebo also has it, I only updated its value.

Comparison before/after:

* Without FSAA, full image:
without_aa_full

* With FSAA = 4, full image:
with_aa_full

* Without FSAA, zoomed:

without_aa_zoom

* With FSAA = 4, zoomed:

with_aa_zoom

### DIFF
--- a/src/rviz/ogre_helpers/render_system.cpp
+++ b/src/rviz/ogre_helpers/render_system.cpp
@@ -239,7 +239,7 @@ void RenderSystem::setupRenderSystem()
   //  renderSys->setConfigOption("RTT Preferred Mode", "FBO");
 
   // Set the Full Screen Anti-Aliasing factor.
-  renderSys->setConfigOption("FSAA", "2");
+  renderSys->setConfigOption("FSAA", "4");
 
   ogre_root_->setRenderSystem(renderSys);
 }
@@ -348,6 +348,9 @@ Ogre::RenderWindow* RenderSystem::makeRenderWindow( intptr_t window_id, unsigned
 #endif
 
   params["externalGLControl"] = true;
+
+  // Enable antialiasing
+  params["FSAA"] = "4";
 
 // Set the macAPI for Ogre based on the Qt implementation
 #ifdef QT_MAC_USE_COCOA


wjwwood:
Cool thanks.

It looks like the Gazebo guys only enable FSAA (full screen AA) when ogre >= 1.8:

https://bitbucket.org/osrf/gazebo/src/38fbac52640382b2f7ff749aca3580cca7fd7994/gazebo/rendering/Camera.cc?at=default&fileviewer=file-view-default#Camera.cc-1237:1239

It would be best if this code still worked against Ogre 1.74 even though we don't use it anymore since Hydro.

I think it might also be wise to expose this as a configuration of rviz some how. Some people might prefer less AA in exchange for better performance, especially since rviz is purely aesthetic unlike the camera in Gazebo which is used for simulated images.

bchretien:
In Gazebo, the Ogre >= 1.8 check is only done for `createManual()`, not for the part of the code I modified (see [here](https://bitbucket.org/osrf/gazebo/src/38fbac52640382b2f7ff749aca3580cca7fd7994/gazebo/rendering/WindowManager.cc?at=default&fileviewer=file-view-default#WindowManager.cc-91)). As for making that a configurable parameter, I'm not familiar enough with RViz's code base, so if you can give me a few tips on how to do that properly, that would probably help :)

wjwwood:
I tested this on my local machine it works great. Thanks! I'll follow up with a pr that adds a command line option to disable it.

bchretien:
@wjwwood thanks!

wjwwood:
I made the follow up pr here: https://github.com/ros-visualization/rviz/pull/950

PR #932, opened 2015-10-17
Don't attempt to `transformMap` until one is actually received.

NikolausDemmel:
Without this patch, the map display shows frame transformation errors
from with an empty frame_id, since it attempts to transform the 
default-initialized occupancy grid variable.


wjwwood:
Thanks. I'll merge when CI completes.

NikolausDemmel:
Cheers!

PR #934, opened 2015-10-21
Update map_display.cpp

DLu:
Change OK status to Warning for downsampled maps.


wjwwood:
Seems reasonable to me.

Issue #935, opened 2015-10-21
Possible issue with interactive marker control with local (moving) coordinate frame


jjnak:
Hi,

When using interactive markers with MOVE_3D and MOVE_ROTATE_3D option, I found that when the marker is clicked, the marker jumps. This happens when the marker is attached to a local coordinate frame (specified header.frame_id in InteractiveMarker) which is different from Fixed Frame specified in rviz. The amount of jump is related to the difference between the fixed frame and the local coordinates.

My guess is that this is possibly handled in moveViewPlane() in interactive_maker_control.cpp in default_plugin in rviz. With the MOVE_PLANE option, this jump does not occur. Also, when the ring and arrow are clicked with ROTATE_AXIS and MOVE_AXIS option, the jump does not happen. 

I would most appreciate it if you could investigate this issue and provide a possible solution to it. I'm using ROS Hydro on Ubuntu 12.04. 

The same issue has been posted in the ROS forum 
http://answers.ros.org/question/138616/rviz-tf-and-interactivemarker-pose-with-respect-to-which-frame/

Best regards,

-Jun

jjnak:
Hi,

Also, this could be related to beginMouseMovement() in handleMouseEvent(). So, it would be great if you could look into the coordinate transform between the current frame and the fixed frame.

Thank you,

-Jun

jjnak:
Hi,

I further looked into this issue, and seemingly setting 
parent_to_grab_position_  = Ogre::Vector3(0,0,0);

solved this problem and looks like it is working fine. So, either by setting parent_to_grab_position_ to zero, or changing the following two functions as follows might be a possible solution (removing  parent_to_grab_position_ from these).

 // find plane on which mouse is moving
    Ogre::Plane plane( event.viewport->getCamera()->getRealDirection(),
                       parent_position_at_mouse_down_);

    parent_->setPose( mouse_position_on_plane,
                      parent_->getOrientation(),
                      name_ );

Your confirmation would be much appreciated!

Best regards,
-Jun

jjnak:
Hi,

Further to my previous post, modification of moveViewPlane() in the following way achieved the behavior that I had expected:

    Ogre::Vector3 camera_direction_ = event.viewport->getCamera()->getRealDirection();
    Ogre::Plane plane( reference_node_->getOrientation().Inverse()*camera_direction_,
                       parent_position_at_mouse_down_);

and

    parent_->setPose( mouse_position_on_plane ,
[...]

PR #937, opened 2015-10-28
Updated moveViewPlane in interactive_marker_control.cpp

jjnak:
As I have reported in #935 in Issues, there seem to be two issues using MOVE_3D option with interactive markers when the marker is attached to a local coordinate frame. One is the marker jumps when clicked when the local coordinate frame is different from Fixed Frame in rviz. The other issue is that the direction of the maker becomes unintuitive when the local frame is rotated relative to the Fixed Frame.

In this update, I tried to fix these issues by modifying moveViewPlane function in interactive_marker_control.cpp by

1) removing parent_to_grab_position_, and
2) rotate the plane associated with the camera view.

### DIFF
--- a/src/rviz/default_plugin/interactive_markers/interactive_marker_control.cpp
+++ b/src/rviz/default_plugin/interactive_markers/interactive_marker_control.cpp
@@ -618,8 +618,9 @@ void InteractiveMarkerControl::moveZAxisWheel( const ViewportMouseEvent& event )
 void InteractiveMarkerControl::moveViewPlane( Ogre::Ray &mouse_ray, const ViewportMouseEvent& event )
 {
   // find plane on which mouse is moving
-  Ogre::Plane plane( event.viewport->getCamera()->getRealDirection(),
-                     parent_position_at_mouse_down_ + parent_to_grab_position_);
+  Ogre::Vector3 camera_direction_ = event.viewport->getCamera()->getRealDirection();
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for the pull request, I'll have a look at this and the other pr as soon as I'm able. Thanks for your contribution and patience!

wjwwood:
@jjnak Can you have give `indigo-devel` a try to see if your issue persists (I merged https://github.com/ros-visualization/rviz/pull/961 which I think might address your issue in part). At the very least I think there is some interaction between this PR and https://github.com/ros-visualization/rviz/pull/961 and some action may be needed here before I can merge it.

Thanks!

jjnak:
@wjwwood Thank you for looking into this issue. Unfortunately, at this moment, I don't have an access to ROS at hand to test this. The way I tried to fix this issue could be a bit heuristic (rotation of the view plane). Thus, as @rhaschke mentinoed in #961, my fix could have introduced some incorrect results in the internal computation of the view plane. When I have a chance, I'll try to test it.

Thank you!

jjnak:
Sorry for my delayed response. I've tried indigo-devel and it works very well. Thanks a lot for fixing this issue.

jjnak:
Sorry for my delayed response. I've tried indigo-devel and it works very well. Thanks a lot for fixing this.

Issue #834, opened 2015-01-07
rviz does not save which tf frames were selected for viewing


jmtatsch:
also, when resetting the clock it forgets the current selection as well.

wjwwood:
@jmtatsch can you give more details, like a step by step example on how to reproduce the problem, what exactly you mean by selected (boxes checked versus the selection tool), and pictures or a short video would be helpful in demonstrating the issue.

Also, what OS and ROS version are you using?

What is the exact version of rviz that you are using?

Without more information, I would not expect anyone could help you with this issue.

jmtatsch:
OS: OS X 10.10.1
ROS indigo
rviz version 1.11.3 compiled against OGRE version 1.9.0
![screen](https://cloud.githubusercontent.com/assets/3080674/5658758/d5de49f0-9707-11e4-9790-611f31bb48b6.png)

wjwwood:
Thanks for the updated information, I'll keep it here on the list and we'll get to it when we can.

ipa-lth:
I just recognized the same issue.
The relevant information (which frame is checked) is stored in the .rviz file, however. It seems like the tf plugin is just not parsing (or using) anything after "All enabled".

If you want to reproduce the problem:
1. Uncheck some frames in the tf plugin
2. Save you configuration
3. Load the previous saved configuration again. (All frames will be checked again.) 

System:
Ubuntu 14.04
ROS Indigo
RViz version 1.11.7 (indigo). Compiled against OGRE version 1.8.1 (Byatis).

zweistein:
Same issue on: Ubuntu 14.04; ROS Indigo, Rviz 1.11.8 and OGRE 1.8.1

wjwwood:
Closing with https://github.com/ros-visualization/rviz/pull/946, please comment if this is still and issue.

PR #938, opened 2015-10-29
On fixing TF loading

arntanguy:
Hi, issue #834 reports that rviz does not correctly save/load the TFs in the configuration file.
This bug is rather annoying to me, as I'm dealing with loads of tfs being published in large clutters and starting to move instantly, so I don't have time to select which ones I'd like to see.

In its current state, the TF and TF tree are correctly saved into the configuration file. However, the 
tf plugin doesn't load them. So I attempted to solve it by loading from the configuration file, however,
there are some issues that require discussion:

- I read the tf configuration from the Config class. For each tf in the configuration file, I create the corresponding FrameInfo, and set whether it's enabled. This works well, provided that the tfs are being published prior to the configuration file loaded.
If the tf are not being published and the configuration file is loaded, then the continuous calls to updateFrames will cause calls to deleteFrame and the newly loaded configuration will be deleted for all frames that are not currently being published. Once the missing tfs get published, their enabled state will be the default one instead of the configuration one.
To deal with it, I see two options:
  - we'd have to never delete frames that have been stored in the configuration file. This of course is not optimal, as it would be impossible to delete frames from the configuration file without manually editing it. It, however, has a side advantage that I really like, it's that all tfs that have been saved will be displayed within the configuration tree, regardless of whether they are published or not. This would allow to preselect which tfs we'd like to see before having the whole clutter published (very useful when dealing with moving robots with loads of tfs)
  - or, instead of loading as in my commit, we could modify updateFrames to check whether a tf is within the configuration or not, and set the enabled state from the configuration if needed. This would of course implies that the configuration (Config class) has to be updated upon changing the enabled state of a tf from the interface. This has the drawback that the user can't start interacting with the TF configuration before they get published. 

- At the moment I do not load the hierarchical tf tree, for the same kind of reason.
A lot of overly complicated logic is happening in updateFrames, and some would contradict with loading from the configuration file.

Let me know what you think about how it ought to behave and be implemented.
Best regards,
Arnaud

### DIFF
--- a/src/rviz/default_plugin/tf_display.cpp
+++ b/src/rviz/default_plugin/tf_display.cpp
@@ -213,6 +213,27 @@ void TFDisplay::onInitialize()
   axes_node_ = root_node_->createChildSceneNode();
 }
 
+void TFDisplay::load(const Config& config) {
+  // TODO parse config Tree
+  // FIXME deleteFrames deletes the tfs that are not published upon loading,
+  //       thus their state is never loaded correctly
+
+  Display::load(config);
+
+  Config c = config.mapGetChild("Frames");
+  for( Config::MapIterator iter = c.mapIterator(); iter.isValid(); iter.advance() )
+  {
+    QString key = iter.currentKey();
+    if(key != "All Enabled") {
+      const Config& child = iter.currentChild();
+      bool enabled = child.mapGetChild("Value").getValue().toBool();
+
+      FrameInfo * info = createFrame(key.toStdString());
+      info->enabled_property_->setBool(enabled);
+    }
+  }
+}
+
 void TFDisplay::clear()
 {
   // Clear the tree.
--- a/src/rviz/default_plugin/tf_display.h
+++ b/src/rviz/default_plugin/tf_display.h
@@ -73,6 +73,7 @@ public:
 protected:
   // Overrides from Display
   virtual void onInitialize();
+  virtual void load(const Config& config);
   virtual void fixedFrameChanged();
   virtual void reset();
 


hersh:
This has bugged me for a few years now.  If this fixes it nicely, I'm all for it.

One concern is that it be tested in use cases that don't exactly match @arntanguy 's use case, to make sure it doesn't end up behaving badly for other people.

For instance, what about the case where you save a file with one set of TF frames, then load the file in a ROS environment where a totally different set of TF frames is going around?  Will you end up with a superset of all the old *and* all the new frames?  Is that OK?  I guess to reset it, you could just delete the display and re-add it.

There is an identical situation in everything that has robot joints and links.  In that situation, the data is again saved correctly but not loaded.  The reason there is that until the robot URDF is loaded we don't actually know what joints and links exist, so we ignore them when loading the config.

I had thought maybe a single change could fix both situations by (perhaps) saving unused config entries until they become relevant (when corresponding-named Properties are created) and load them then.

In any case, the current pull request could be merged and then revisited when/if the larger fix were implemented.

arntanguy:
Well, the current code in this pull request doesn't fix it nicely, it is more meant as a discussion starter.
I've restricted the code in the pull request to a minimal suboptimal solution, that, as far as I know has no side effects:
- If the tfs are already published, and the configuration file is loaded, then those in the configuration will be loaded with their restored state, those in the configuration but not published will be loaded, but immediately destroyed, and the ones not in the configuration file will be loaded with default configuration.

No side effects, no artificial superset. But from a user's perspective, the behavior isn't logical. This forces the TFs to be published before the configuration file is loaded to get the expected behavior. I don't really think this is acceptable.

Currently I'm using the first trick I mentioned never to delete the FrameInfo for TFs that are in the configuration file.
While this works fine, and allows me to get the work I need to be doing done without worrying when the TFs are being published, I don't think this is an acceptable solution for all use-cases and end-users.

Fiddling with the Properties as @hersh proposes seems a much better option. However, since
this has quite a strong potential for side-effects, and I'm not familiar enough with the properties model for rviz, I'm not really comfortable fiddling with it by myself.

wjwwood:
Closing in favor of https://github.com/ros-visualization/rviz/pull/946.

PR #944, opened 2015-11-23
Minor map color documentation fix in map_display.cpp

stwirth:
Color for lethal obstacle values for the obstacle map display is yellow, not purple.


wjwwood:
Thanks.

PR #946, opened 2015-12-03
Fix for tf display not loading frame config

bponsler:
This is a fix for issue [#834](https://github.com/ros-visualization/rviz/issues/834) using the changes from PR [#938](https://github.com/ros-visualization/rviz/pull/938) by @arntanguy with the following modifications:

1. Rather than adding a new frame and setting the enabled state of the frame immediately when reading the config, the enabled state for each specified frame is stored in a map.
2. Whenever a frame is being added to the display it looks up its enabled state in the map, and if it was configured as disabled the frame is set to be disabled.

Because no new frames are created when loading the config this properly handles the case where frames specified in the config file are not published at the time the config is read mentioned in PR [#938](https://github.com/ros-visualization/rviz/pull/938). 

This also handles the case (mentioned by @hersh) where the set of frames specified in the config file are different than the set of frames being actively published. Since no new frames are created based on the config file only the intersection of those two sets will have their enabled state modified, and only the frames being actively published will be added to the display.

### DIFF
--- a/src/rviz/default_plugin/tf_display.cpp
+++ b/src/rviz/default_plugin/tf_display.cpp
@@ -206,11 +206,34 @@ TFDisplay::~TFDisplay()
 
 void TFDisplay::onInitialize()
 {
+  frame_config_enabled_state_.clear();
+
   root_node_ = scene_node_->createChildSceneNode();
 
   names_node_ = root_node_->createChildSceneNode();
   arrows_node_ = root_node_->createChildSceneNode();
   axes_node_ = root_node_->createChildSceneNode();
+}
+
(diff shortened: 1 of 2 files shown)


arntanguy:
Thanks @bponsler. I've just given this a quick test on some of my complex use cases with a humanoid robot, and real-life situations with intense use of the tf graph. As far as I can see, this seems to work as expected and provide a decent simple solution to this longstanding problem.

wjwwood:
I tested this out locally and it works great for me too. I think the code changes make sense too. So I'm going to merge this and close #938 in favor of this.

Thanks @bponsler and @arntanguy for your work on this!

PR #948, opened 2015-12-05
Remove default_plugin from exported libraries

jspricke:
The plugin is handled by pluginlib. Exporting it creates an unneeded dependency in rvizConfig.cmake.


wjwwood:
Makes sense, thanks.

Issue #945, opened 2015-12-02
rviz 1.11.8+ cannot load STL


seanyun:
The stl files which work fine with rviz `1.11.7` are not working with `1.11.8`+.
Here's the error from `1.11.10` (same as any version `1.11.8`+):

```
[ERROR] [1449097635.190619726]: The STL file 'package://####_description/meshes/####_link.stl' is malformed. According to the binary STL header it should have '208' triangles, but it has too much data for that to be the case.
[ERROR] [1449097635.190681335]: Failed to load file [package://####_description/meshes/####_link.stl]
[ WARN] [1449097635.190821463]: OGRE EXCEPTION(6:FileNotFoundException): Cannot locate resource package://####_description/meshes/####_link.stl in resource group Autodetect or any other group. in ResourceGroupManager::openResource at /build/buildd/ogre-1.8-1.8.1+dfsg/OgreMain/src/OgreResourceGroupManager.cpp (line 756)
[ERROR] [1449097635.191032928]: Could not load model 'package://####_description/meshes/####_link.stl' for link '####_link': OGRE EXCEPTION(6:FileNotFoundException): Cannot locate resource package://####_description/meshes/base_link.stl in resource group Autodetect or any other group. in ResourceGroupManager::openResource at /build/buildd/ogre-1.8-1.8.1+dfsg/OgreMain/src/OgreResourceGroupManager.cpp (line 756)
```

I commented out the real names because the stl files are proprietary. So Unfortunately I can't post the files.

wjwwood:
@seanyun would it be possible for you to share one of these STL files with me confidentially? That might help me understand why this is happening.

seanyun:
I'm sorry that I can't send the file now. I'll talk about it with my company and let you know.
Meanwhile, I played with the rviz source file, and it seems the difference of the sizes is just '1'.

[ERROR] [1449100802.924760814]: The STL file 'package://####.stl' is malformed. num triangles: 208  binary_stl_header_len: 84 num_bytes: 10485 expected_size: 10484
[ERROR] [1449100802.951429978]: The STL file 'package://####.stl' is malformed. num triangles: 1896  binary_stl_header_len: 84 num_bytes: 94885 expected_size: 94884
[ERROR] [1449100802.980868243]: The STL file 'package://####.stl' is malformed. num triangles: 1896  binary_stl_header_len: 84 num_bytes: 94885 expected_size: 94884

seanyun:
[...]

PR #951, opened 2015-12-10
Allow STL files bigger than expected (printing a warning message).

caguero:
See issue https://github.com/ros-visualization/rviz/issues/945.


wjwwood:
Thanks @caguero!

@seanyun can you try this to see if it at least loads your STL's? Strictly speaking I think your STL's are not compliant to the STL standard (if I read it correctly).

seanyun:
I'm sorry about the late response. This patch works. I got a bunch of warnings but the graphical model shows up in the right way. Thx!

wjwwood:
Ok, I'll merge this, but I'd still like to know why your models have extra data at the end. Do they open in other tools like Blender and Meshlab?

seanyun:
I can open the STL files in Blender smoothly. I don't see any warning or errors in Blender. 
The STL files were generated by the SolidWorks auto-export.

JamesNewton:
I seem to be having this issue with the attached file. Version 1.12.14.
[LowRes_DexterHDIKinematic_Link1_Bottom.stl.zip](https://github.com/ros-visualization/rviz/files/5079926/LowRes_DexterHDIKinematic_Link1_Bottom.stl.zip)

Issue #679, opened 2013-09-05
Ability to rename saved views is gone in Hydro rviz.


hershwg:
This change c6dacb158cd0589ba9534198b423f03575991fcb (make read-only properties appear disabled) breaks the ability to rename saved Views.

Open the "Views" panel, click "Save", then click on the newly-saved view entry which will be named after the current ViewController, often "Orbit".

First problem is that the entry has a gray text color, so looks bad.

The next problem is that the entry is not selectable, so there is no way to rename or remove it.  In the old days (before c6dacb158cd0589ba9534198b423f03575991fcb), the saved view entries were selectable but read-only.  They were drawn in black text and were remove-able and rename-able.

In my opinion the change c6dacb158cd0589ba9534198b423f03575991fcb could just be reversed.

PR #958, opened 2016-01-04
Make read-only properties read-only

rhaschke:
A read-only property shouldn't be allowed to be edited ;-)
Fixed by not creating an editor in that case.

Related #654 which attempted to solve that issue by disabling the property,
which in turn was reverted in #679 due to its side effects.


wjwwood:
Any idea how this relates to https://github.com/ros-visualization/rviz/pull/687?

wjwwood:
lgtm

I was able to reproduce by using the `move_base_amcl_10cm.launch` example in the `navigation_stage` package. Before the patch I could edit the position fields for the costmaps which are of display type "map". After the patch I could not.

PR #961, opened 2016-01-05
fix #959: jumping marker in MOVE_3D mode

### LINKED ISSUE #959: jumping interactive marker
Using a `MOVE_ROTATE_3D` control for an interactive marker *and* defining the marker relative to a non-root frame *and* defining it frame-locked, will make the marker jump (by the inverse transform from root frame to ref frame) as soon as the this control is used. Other controls defined for the same interactive marker, behave correctly.

The behavior can be reproduced with the following changes to `interactive_marker_tutorials/src/simple_marker.cpp`:

```
diff --git a/interactive_marker_tutorials/src/simple_marker.cpp b/interactive_marker_tutorials/src/simple_marker.cpp
index bc7c758..172aec6 100644
--- a/interactive_marker_tutorials/src/simple_marker.cpp
+++ b/interactive_marker_tutorials/src/simple_marker.cpp
@@ -50,8 +50,8 @@ int main(int argc, char** argv)
 
   // create an interactive marker for our server
   visualization_msgs::InteractiveMarker int_marker;
-  int_marker.header.frame_id = "base_link";
-  int_marker.header.stamp=ros::Time::now();
+  int_marker.header.frame_id = "tool0";
+// make it frame-locked! int_marker.header.stamp=ros::Time::now();
   int_marker.name = "my_marker";
   int_marker.description = "Simple 1-DOF Control";
 
@@ -70,6 +70,7 @@ int main(int argc, char** argv)
   visualization_msgs::InteractiveMarkerControl box_control;
   box_control.always_visible = true;
   box_control.markers.push_back( box_marker );
+  box_control.interaction_mode = visualization_msgs::InteractiveMarkerControl::MOVE_ROTATE_3D;
 
   // add the control to the interactive marker
   int_marker.controls.push_back( box_control );
```

This issue was reported before [here](http://answers.ros.org/question/138616/rviz-tf-and-interactivemarker-pose-with-respect-to-which-frame/).

rhaschke:
Initial fix for #959: jumping interactive marker in MOVE_3D mode.
I need to check whether other modes are affected too.

### DIFF
--- a/src/rviz/default_plugin/interactive_markers/interactive_marker_control.cpp
+++ b/src/rviz/default_plugin/interactive_markers/interactive_marker_control.cpp
@@ -619,7 +619,7 @@ void InteractiveMarkerControl::moveViewPlane( Ogre::Ray &mouse_ray, const Viewpo
 {
   // find plane on which mouse is moving
   Ogre::Plane plane( event.viewport->getCamera()->getRealDirection(),
-                     parent_position_at_mouse_down_ + parent_to_grab_position_);
+                     grab_point_in_reference_frame_);
 
   // find intersection of mouse with the plane
   std::pair<bool, Ogre::Real> intersection = mouse_ray.intersects(plane);
@@ -628,7 +628,7 @@ void InteractiveMarkerControl::moveViewPlane( Ogre::Ray &mouse_ray, const Viewpo
   Ogre::Vector3 mouse_position_on_plane = mouse_ray.getPoint(intersection.second);
 
   // move parent so grab position relative to parent coincides with new mouse position.
-  parent_->setPose( mouse_position_on_plane - parent_to_grab_position_,
+  parent_->setPose( mouse_position_on_plane - grab_point_in_reference_frame_ + parent_position_at_mouse_down_,
                     parent_->getOrientation(),
                     name_ );
 }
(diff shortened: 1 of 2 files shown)


rhaschke:
+1
All other modes worked (and still work) fine. I tested with standard `basic_controls` example,
defining `base_link` as a non-root frame in rviz. The only issue was with MOVE_3D and MOVE_ROTATE_3D when moving.

rhaschke:
@wjwwood Could you please review this fix for #959 and consider it for merging?

wjwwood:
Yeah, I'm currently working through the backlog. I want to get through them and make a release this week, so that I can fork for Qt5 and Kinetic.

wjwwood:
Ok, I was able to reproduce this and see that this fixes the issue. Thanks!

I'm going to merge this, but @rhaschke if you have time, can you comment on how https://github.com/ros-visualization/rviz/issues/935 and https://github.com/ros-visualization/rviz/pull/937 related to this issue. They seem very similar, but I haven't looked into those just yet.

rhaschke:
@wjwwood, indeed #935 and #937 describe exactly the same issue - I didn't see them when browsing the issue history. @jjnak found a quite similar solution to mine - removing the variable `parent_to_grab_position_`. However, his modification of the view plane computation results in strange results: If the reference frame is rotated, mouse motion in the view plane result in translations in depth instead of xy of view plane.

jjnak:
Sorry for my delayed response. I've tried the current indigo-devel and it works very well. @rhaschke and @wjwwood, thanks a lot!

PR #962, opened 2016-01-18
Updated the marker display to allow namespaces to be enabled/disabled from the config

bponsler:
The marker display does not currently load the enabled/disabled state of namespaces from its config. Those values are simply not used. These changes load the enabled/disabled state of namespaces from the config into a map which gets used to disable namespaces when they are added.

This is essentially the same fix as commit 0baa9f351cb73f880141ecfabd61aaa224de7458 but applied to the marker display instead of the tf display.


wjwwood:
lgtm

I was able to reproduce this with the `basic_shapes` demo in the `visualization_marker_tutorials` package.

PR #963, opened 2016-01-19
Change costmap lethal color different from illegal values.

dornhege:
Lethal obstacles are now purple to make them different from illegal negative values that might also have been yellow.

Note that this reverts the previous commit that changed the comment from purple to yellow and instead changes the implementation. If there is a concern that the display is changed, I could also do this the other way around and make negative values in purple, but I believe that this is how it was intended originally.

### DIFF
--- a/src/rviz/default_plugin/map_display.cpp
+++ b/src/rviz/default_plugin/map_display.cpp
@@ -182,10 +182,10 @@ unsigned char* makeCostmapPalette()
   *palette_ptr++ = 255; // green
   *palette_ptr++ = 255; // blue
   *palette_ptr++ = 255; // alpha
-  // lethal obstacle values (100) in yellow
+  // lethal obstacle values (100) in purple
   *palette_ptr++ = 255; // red
-  *palette_ptr++ = 255; // green
-  *palette_ptr++ = 0; // blue
+  *palette_ptr++ = 0; // green
+  *palette_ptr++ = 255; // blue
   *palette_ptr++ = 255; // alpha
   // illegal positive values in green
   for( int i = 101; i <= 127; i++ )


wjwwood:
Can I get some input on this from some of the navigation maintainers? @mikeferguson @jonbinney @DLu

jonbinney:
This sounds reasonable, although it might be a fairly surprising change for people who have been using indigo for a while. Should this go into jade?

wjwwood:
This would be the first difference between Jade and Indigo, so selfishly I would prefer to push it to Kinetic if it is not to be put into Indigo (I will probably have to fork for Kinetic anyways).

jonbinney:
Waiting for Kinetic akes sense. Although I think this is a good change, it doesn't feel like a super-time-critical thing.

DLu:
At one point I had a version of this that had dropdowns where you could select all the colors used. That would probably be the ideal.

DLu:
What about a plugin architecture with the color schemes each being their own plugin that defined a function like 

    void getColor(unsigned char cost, unsigned char* red, unsigned char* green, unsigned char* blue, unsigned char* alpha);

wjwwood:
My opinion is the plugin architecture is a bit extreme, but I would consider a pull request that added it. I'd also consider a fixed list in a drop down.

However, I'd appreciate +1/-1 on merging this at least for now.

jonbinney:
I'm -1 for merging this into indigo; +1 for merging a similar change into kinetic.

wjwwood:
Ok, I'll mark this for inclusion into Kinetic only. Speak up if you disagree.

wjwwood:
I'm going to close this in favor of https://github.com/ros-visualization/rviz/pull/994

PR #965, opened 2016-02-01
Fix CMake warning about FindEigen.cmake being deprecated. 

tobiasberling:
CMake prints the following warning:
```
-- checking for module 'eigen3'
--   found eigen3, version 3.2.0
CMake Warning at /opt/ros/jade/share/cmake_modules/cmake/Modules/FindEigen.cmake:62 (message):
  The FindEigen.cmake Module in the cmake_modules package is deprecated.

  Please use the FindEigen3.cmake Module provided with Eigen.  Change
  instances of find_package(Eigen) to find_package(Eigen3).  Check the
  FindEigen3.cmake Module for the resulting CMake variable names.

  CMakeLists.txt:130 (find_package)


-- Found Eigen: /usr/include/eigen3  
-- Eigen found (include: /usr/include/eigen3)
```

Replacing `Eigen` with `Eigen3` as proposed by the warning message works for me (Ubuntu 14.04.3 LTS).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,7 @@ endif()
 
 find_package(PythonLibs "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" REQUIRED)
 
-find_package(Eigen REQUIRED)
+find_package(Eigen3 REQUIRED)
 
 catkin_python_setup()
 


wjwwood:
Hi, thanks for the pull request, but I don't think this is sufficient because you need to change the usage of that `find_package()` as well, e.g.: https://github.com/ros-visualization/rviz/pull/965/files#diff-af3b638bc2a3e6c650974192a53c7291R170

For more info see: http://wiki.ros.org/jade/Migration#Eigen_CMake_Module_in_cmake_modules

PR #966, opened 2016-02-01
Add package path cache to load_resource.cpp getPath()

tobiasberling:
When having a lot of display plugins in one package, `getPath` is called for the same package name over and over again. To speed things up, the result from `ros::package::getPath` is now cached in memory.

In my case, it took about 15 seconds to load the add display popup. Opening it a second time within 30 seconds is fast. Looking into the implementation of QPixmapCache showed that there is a reset timer that clears the cache every 30 seconds (see [QTBUG-31452](https://bugreports.qt.io/browse/QTBUG-31452)). That explains the fast reopening. 

Logging calls to `getPath` showed that the package name was not changing that often but only the resource name. The package path got requested from `ros::package::getPath` every time. This seems to have some delay which sums up.  After adding caching, the number of calls to `ros::package::getPath` in my case were reduced from 234 to 10 and the dialog is opening instantaneously now.

### DIFF
--- a/src/rviz/load_resource.cpp
+++ b/src/rviz/load_resource.cpp
@@ -33,6 +33,7 @@
 #include <ros/package.h>
 #include <ros/ros.h>
 
+#include <QCache>
 #include <QPixmapCache>
 #include <QPainter>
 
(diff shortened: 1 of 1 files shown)


wjwwood:
Using a `QCache` which can reset based on a timer doesn't really make sense to me. I'd say we should either not cache or cache the results using a normal `std::map` and then add a refresh button which reopens the dialog after clearing the map.

What do you think?

rhaschke:
This issue was solved upstream in roslib/rospack (currently only in jade-devel):
https://github.com/ros/ros/pull/97
https://github.com/ros/rospack/pull/49

wjwwood:
I'm going to close this based on @rhaschke's feedback. Please comment here if you disagree.

PR #970, opened 2016-02-10
Convert to class

cmansley:
ViewportMouseEvent is forward declared as a class in other headers, which causes warnings in clang with -Wmismatched-tags


wjwwood:
Thanks.

PR #971, opened 2016-02-20
Add checkbox to hint for unreliable (UDP) topic transport for DisplayFilter based plugins.

mikeodr:
Added "Unreliable hint" option to base MessageFilterDisplay template
class.

This allows the user to choose the option to use UDP unreliable [TransportHints](http://docs.ros.org/hydro/api/roscpp/html/classros_1_1TransportHints.html) for high throughput topics in a bandwidth constrained environment where TCP retransmits cause more congestion.

screenshot from 2016-02-19 19 41 23

Verified that saving the rviz config shows `Unreliable: true` for topics selected with the option.
Also verified in wireshark that UDP is correctly used for topics selected and publishers that support it.

### DIFF
--- a/src/rviz/message_filter_display.h
+++ b/src/rviz/message_filter_display.h
@@ -142,15 +142,11 @@ protected:
 
       try
       {
-        ros::TransportHints transport_hint;
+        ros::TransportHints transport_hint = ros::TransportHints().reliable();
         // Determine UDP vs TCP transport for user selection.
         if (unreliable_property_->getBool())
         {
           transport_hint = ros::TransportHints().unreliable();
-        }
-        else
-        {
-          transport_hint = ros::TransportHints().reliable();
         }
         sub_.subscribe( update_nh_, topic_property_->getTopicStd(), 10, transport_hint);
         setStatus( StatusProperty::Ok, "Topic", "OK" );
--- a/src/rviz/message_filter_display.h
+++ b/src/rviz/message_filter_display.h
@@ -58,6 +58,10 @@ public:
       topic_property_ = new RosTopicProperty( "Topic", "",
                                               "", "",
                                               this, SLOT( updateTopic() ));
+      unreliable_property_ = new BoolProperty( "Unreliable", false,
+                                               "Prefer UDP topic transport",
+                                               this,
+                                               SLOT( updateTopic() ));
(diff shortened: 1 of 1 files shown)


wjwwood:
Cool, the code changes look good. I'll try to test it out this weekend or the beginning of next week.

Thanks.

src/rviz/message_filter_display.h
```diff
@@ -137,7 +142,17 @@ class MessageFilterDisplay: public _RosTopicDisplay
...
       try
       {
-        sub_.subscribe( update_nh_, topic_property_->getTopicStd(), 10 );
+        ros::TransportHints transport_hint;
+        // Determine UDP vs TCP transport for user selection.
+        if (unreliable_property_->getBool())
+        {
+          transport_hint = ros::TransportHints().unreliable();
+        }
+        else
```
mikepurvis: Else clause is superfluous, I think. Also, do you have to instantiate another one and copy? What about just:

```
       if (unreliable_property_->getBool())
       {
         transport_hint.unreliable();
       }
```

mikeodr: I don't see a constructor in the header for `TransportHints` but yes, I can default to reliable and do the check as you propose.

src/rviz/message_filter_display.h
```diff
@@ -58,6 +58,10 @@ Q_OBJECT
       topic_property_ = new RosTopicProperty( "Topic", "",
                                               "", "",
                                               this, SLOT( updateTopic() ));
+      unreliable_property_ = new BoolProperty( "Unreliable", false,
```
mikeodr: I don't see a delete for `topic_property_` and that concerns me, or am I missing something?

mikeodr: More specifically is Qt doing my magic to clean this up?

wjwwood: I believe this code handles the deleting: https://github.com/ros-visualization/rviz/blob/indigo-devel/src/rviz/properties/property.cpp#L84-L98

It's definitely a round about way to do it.

wjwwood:
This is great! Thanks for the pr. I was also able to see the laser data from `stage` come over UDP using wireshark.

PR #972, opened 2016-02-24
Adding raw mode for map vizualization

aaronhoy:
This allows map pixel values to be passed directly through and drawn as-is in grayscale.
This will be especially useful in combination with ros-planning/navigation#422


mikeferguson:
+1

src/rviz/default_plugin/map_display.cpp
```diff
@@ -212,6 +213,22 @@ unsigned char* makeCostmapPalette()
   return palette;
 }
 
+unsigned char* makeRawPalette()
+{
+  unsigned char* palette = new unsigned char[256*4];
```
wjwwood: This is never `delete`'ed. I see that the other `make*Palette()` functions do the same... I guess this is fine as is, but really they should all be updated to use static arrays or something like that.

wjwwood:
lgtm

PR #976, opened 2016-03-13
Unreliable option on image, camera and map

joao-avelino:
When streaming videos it's usually a good idea to do it using UDP. This is specially true if we are connected through Wi-Fi or other "unreliable" connection.

Since we were having problems visualizing the streams coming from our robot over Wi-Fi, we decided to make the **Unreliable** (UDP) option also available for **Image** and **Camera** displays. We also made it available for Map.

### DIFF
--- a/src/rviz/default_plugin/map_display.cpp
+++ b/src/rviz/default_plugin/map_display.cpp
@@ -112,6 +112,11 @@ MapDisplay::MapDisplay()
                                                   "Orientation of the map. (not editable)",
                                                   this );
   orientation_property_->setReadOnly( true );
+
+  unreliable_property_ = new BoolProperty( "Unreliable", false,
+                                           "Prefer UDP topic transport",
+                                           this,
+                                           SLOT( updateTopic() ));
 }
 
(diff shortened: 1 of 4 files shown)


wjwwood:
@joao-avelino have you seen https://github.com/ros-visualization/rviz/pull/971?

joao-avelino:
Hey there, @wjwwood!

Yes, I've seen #971, and this changes are based on it. #971 doesn't make the option available for **Image** and **Camera** displays, since they don't inherit from "MessageFilterDisplay" where the "unreliable option" is implemented in #971.

That's also true for the **Map Display**.

NikolausDemmel:
I'm not sure what the Relation to MessageFilterDisplay is (i.e. should Camerw/Image/Map use MessageFilterDisplay), but the patch seems simple enough as it is.

I haven't tested it yet, but I would want to use it next time I am in the field with flaky WiFi, so +1.

wjwwood:
@joao-avelino got it. I tested this and it looks good. Thanks!

PR #977, opened 2016-03-14
Add "Duplicate" button to duplicate display with same config

wkentaro:
Demo: https://youtu.be/FxUOrkBzVzs

**1**

**2**

**3**

### DIFF
--- a/src/rviz/displays_panel.cpp
+++ b/src/rviz/displays_panel.cpp
@@ -58,6 +58,10 @@ DisplaysPanel::DisplaysPanel( QWidget* parent )
   QPushButton* add_button = new QPushButton( "Add" );
   add_button->setShortcut( QKeySequence( QString( "Ctrl+N" )));
   add_button->setToolTip( "Add a new display, Ctrl+N" );
+  duplicate_button_ = new QPushButton( "Duplicate" );
+  duplicate_button_->setShortcut( QKeySequence( QString( "Ctrl+D" )));
+  duplicate_button_->setToolTip( "Duplicate a display, Ctrl+D" );
+  duplicate_button_->setEnabled( false );
   remove_button_ = new QPushButton( "Remove" );
   remove_button_->setShortcut( QKeySequence( QString( "Ctrl+X" )));
   remove_button_->setToolTip( "Remove displays, Ctrl+X" );
@@ -69,6 +73,7 @@ DisplaysPanel::DisplaysPanel( QWidget* parent )
 
   QHBoxLayout* button_layout = new QHBoxLayout;
   button_layout->addWidget( add_button );
+  button_layout->addWidget( duplicate_button_ );
   button_layout->addWidget( remove_button_ );
   button_layout->addWidget( rename_button_ );
   button_layout->setContentsMargins( 2, 0, 2, 2 );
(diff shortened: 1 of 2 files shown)


wkentaro:
ping

NikolausDemmel:
This seems like a very useful addition. Havent tested it, but high level review lgtm.
+1

wjwwood:
@wkentaro thanks for the pr, this is a great feature!

I didn't, however, get a new display name (copy of ...) and so I opened a pull request for that: https://github.com/wkentaro/rviz/pull/1

Also I'd like to ask you to consider changing the use of the term "copy" with "duplicate". Since you're not exactly copying it in the sense of copy and then paste zero to many times. You're really duplicating an item exactly one time.

So if you choose to do "duplicate" instead of "copy", then I'd recommend changing the new display name to be "duplicate of X", change the button to "duplicate" rather than "copy" and consider changing the hotkey of ctrl-c to ctrl-d (or something like that).

Thanks again.

wkentaro:
> I didn't, however, get a new display name (copy of ...) and so I opened a pull request for that: wkentaro#1

I rethought about the name of copied display, and changed it to be same as the original one.
Which one do you prefer?

wkentaro:
> Also I'd like to ask you to consider changing the use of the term "copy" with "duplicate". Since you're not exactly copying it in the sense of copy and then paste zero to many times. You're really duplicating an item exactly one time.
So if you choose to do "duplicate" instead of "copy", then I'd recommend changing the new display name to be "duplicate of X", change the button to "duplicate" rather than "copy" and consider changing the hotkey of ctrl-c to ctrl-d (or something like that).

It's reasonable, I use "duplicate" term.

wkentaro:
@wjwwood I have updated the commit.

wjwwood:
@wkentaro looks good thanks. I'm testing it out now.

As for the new name, I think having something different is a good idea, but I think a suffix or prefix that is too long might not be best, e.g. "duplicate of X" or "X duplicate" or even "X dupe". I thought about suffixing numbers, e.g. "X (1)", then "X (2)", but the logic to get this right is tricky (and to avoid overmatching something the user created). Something simple could be to use the mathematical prime symbol to indicate a duplicate: https://en.wikipedia.org/wiki/Prime_(symbol) That would be easy to implement as you just add a single quote to the end of the name.

Anyone else have thoughts?

NikolausDemmel:
I woudn't mind having the duplicates with the exact same name. That is what rviz does already when adding new displays. It defaults to the same name no matter if it exists already or not.

But I am not strongly attached to that and all your suggestions seem fine to me, with a preference for a short suffix. I actually think the `X (1)` thing would be nicest, but I see that it is not trivial to implement like the others. For all options one would need a recursive check for the suffixed-name's existence though.

wjwwood:
Yeah, there's more issues with the changing of names (as @wkentaro may have discovered and finally decided not to do that). For example, nothing stops you from duplicating `foo` as `foo'`, then duplicating it again. Now you have `foo`, `foo'`, and another `foo'`. Also you can duplicate multiple items so you could duplicate `foo` so you also have `foo'`, select both and duplicate again. Now you have `foo`, `foo'`, `foo'`, and `foo''`.

I'm thinking of going back to just duplicating it and not changing the name.

I also have a change locally that will ensure the duplicated item(s) are selected after duplication.

wjwwood:
@wkentaro I opened https://github.com/wkentaro/rviz/pull/2 against this pr.

I might do the merge myself later today.

wjwwood:
@wkentaro I merged this manually with a few changes about what is selected after duplicating and deleting displays. Thanks again!

PR #980, opened 2016-03-22
port rviz to Qt5 with continued support for Qt4

wjwwood:
I rebased and squashed @simonschmeisser's commits in https://github.com/ros-visualization/rviz/pull/909. I'll also be cherry-picking over my fixes for the splash screen and stuff like that.

This will replace https://github.com/ros-visualization/rviz/pull/909.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 2.8.12)
 project(rviz)
 
 if (POLICY CMP0042)
@@ -91,7 +91,18 @@ endif()
 
 find_package(OpenGL REQUIRED)
 
-find_package(Qt4 REQUIRED COMPONENTS QtCore QtGui QtOpenGL)
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+option(UseQt5 "UseQt5" OFF)
+if (UseQt5)
+  find_package(Qt5Widgets REQUIRED)
+  find_package(Qt5Core REQUIRED)
+  find_package(Qt5OpenGL REQUIRED)
+else (UseQt5)
+  find_package(Qt4 REQUIRED COMPONENTS QtCore QtGui QtOpenGL)
+endif(UseQt5)
+  
 
 find_package(catkin REQUIRED
   COMPONENTS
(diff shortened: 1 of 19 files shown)


rhaschke:
@wjwwood 
I think rvizConfig.cmake should advertize the used Qt version and `QT_DIR`, e.g. as `RVIZ_QT_VERSION` and `RVIZ_QT_DIR` in order to allow plugins to build against the same Qt version.
It might be also helpful to have an additional auto-generated cmake file, that can be used to pull in the Qt dependencies on demand. Obviously catkin_package doesn't expose the Qt dependency.

wjwwood:
@rhaschke that's a good idea, @dirk-thomas suggested something very similar just today. I will make sure to include that.

NikolausDemmel:
> Obviously catkin_package doesn't expose the Qt dependency.

Why is this the case? Do you not need qt headers to build something against librviz?

rhaschke:
>> Obviously catkin_package doesn't expose the Qt dependency.

> Why is this the case? Do you not need qt headers to build something against librviz?

This way, one can pull in a different, but compatible Qt version?

NikolausDemmel:
> This way, one can pull in a different, but compatible Qt version? I have to admit, that this might be a rare use case ;-)

Is this a common pattern in catkin / cmake? I was under the impression that with catkin one is expected to export all build-export dependencies in `catkin_package`. I just noticed that rviz actually exports none of its non-catkin-pkg-dependencies with `DEPENDS`. @wjwwood, do you know if this is intentional, or simply something no-one ever added?

> Hence advertising the Qt version and/or its cmake source dir is mandatory to allow for ABI compatibility.

True. This is a general problem not specific to Qt. The same could be said for other libraries with components like Boost, OpenCV, Ogre. Maybe Qt is just a very prominent example where this occurs often and where people tend to have multiple versions installed in parallel (in particular thanks to qmake and QtCreator being rather good at handling multiple installed versions of Qt in parallel.

`RVIZ_QT_VERSION` and `RVIZ_QT_DIR` sound good to me, but I was woundering if that is an ad-hoc solution for this case, and if there might be a cmake pattern for this kind of thing already in use elsewhere that we might consider adopting. But feel free to ignore my ramblings :).

wjwwood:
While it is true that rviz should probably export Qt headers and libraries (unless explicitly supporting the different but compatible Qt pattern @rhaschke mentioned), if your plugin is using Qt in its headers at all (which I think it the case for most plugins) you should be finding and using Qt yourself, a.k.a. depend on what you use or include what you use. So the problem might have not been noticed.

My intention is just to add the cmake variables to let someone know what rviz is using, but I'll also have a look at what libraries rviz is exporting at the same time.

rhaschke:
Another reason, why `catkin_package` shouldn't simply export the Qt system deps, is that `find_package(Qt)` pulls in a lot of cmake magic to handle moc processing etc. So you better find Qt yourself.

wjwwood:
I was just looking and there doesn't seem to be a consistent `_DIR` variable between versions of Qt. So I'm just going to export the version number. I think that should be sufficient for most cases. See: https://github.com/ros-visualization/rviz/pull/985

PR #982, opened 2016-03-30
generalize WrenchVisual

rhaschke:
This PR attempts to increase code-reusability for `rviz::WrenchStampedVisual`, which is currently used by the wrench display plugin.
The visual itself doesn't handle or use the timestamp. Hence I suggest to:
- change the API of `setMessage()` to accept a `geometry_msgs::Wrench` 1c341c0
- rename `setMessage()` to `setWrench()` 1c341c0
- add an additional method `setWrench(Vector3 force, Vector3 torque)` 24e4bab

Furthermore, I modified the code to allow for
- hiding the markers e78ebee
- auto-hiding the force or torque marker if their magnitude becomes too small (which created strange artifacts in existing code) 7a51333

### DIFF
--- a/src/rviz/default_plugin/wrench_display.cpp
+++ b/src/rviz/default_plugin/wrench_display.cpp
@@ -149,7 +149,7 @@ void WrenchStampedDisplay::processMessage( const geometry_msgs::WrenchStamped::C
     }
 
     // Now set or update the contents of the chosen visual.
-    visual->setMessage( msg );
+    visual->setWrench( msg->wrench );
     visual->setFramePosition( position );
     visual->setFrameOrientation( orientation );
     float alpha = alpha_property_->getFloat();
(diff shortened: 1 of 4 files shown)


wjwwood:
@rhaschke I'll have to consider this for Kinetic-only. I think there's a good change others are already using the API as-is in Indigo/Jade.

rhaschke:
@wjwwood Targeting kinetic is fine. To maintain API stability, I can add the original setMessage() function, which would be a simple wrapper then. 
If you agree to this idea, I could prepare an API-compatible version targeted to indigo, depreacating the critical API functions for later removal in kinetic?

wjwwood:
@rhaschke that would work fine. I would consider releasing for Indigo if this only added to the current API.

rhaschke:
@wjwwood I fixed my changes to retain API and ABI compatibility.
If you have merged this into Indigo, I will create another PR for Kinetic, merging and cleaning up this, i.e. removing the deprecated function and renaming the class WrenchStampedVisual to WrenchVisual.

wjwwood:
lgtm, I'm going to squash it somehow.

PR #983, opened 2016-03-30
set rviz_DEFAULT_PLUGIN_LIBRARIES during configure time

wjwwood:
Attempt to fix `catkin_make` builds.


wjwwood:
See: https://github.com/ros-planning/moveit_ros/issues/654#issuecomment-203407388

PR #988, opened 2016-04-07
Updated the marker display and tf plugins to not reset enabled namespaces/frames

bponsler:
Updated the marker display and tf plugins to update the map of enabled namespaces and frames whenever those frames are enabled/disabled using the check boxes. Also updated the plugins so that the map of enabled namespaces and frames does not get erased whenever the plugin is reset.

This allows the currently selected namespaces/frames to remain selected after the Reset button is pressed.


stuglaser:
I tried it out, and the behavior looks right to me.  Thanks for the fix!

wjwwood:
Thanks guys, lgtm me too.

PR #992, opened 2016-04-08
make Qt5 the default for Kinetic, fix issues

wjwwood:
This pr will make Qt5 the default option (rather than Qt4) in ROS Kinetic. It will also address some issues with Python support.


wjwwood:
@ros-pull-request-builder please test this

wjwwood:

wjwwood:
I'll have to disable PySide2 support for now as I had to do a lot of hacks to get Shiboken2 to do the code generation and even then the generated code results in invalid C++ code (which fails to compile). I'm going to defer fixing that until a future Kinetic release, so I've disabled PySide2 support for now. My work so far is in this branch: https://github.com/ros-visualization/rviz/tree/wjwwood-qt5-pyside2

/cc @dirk-thomas

PR #1007, opened 2016-05-13
Create multiple textures if map doesn't fit in a single texture

DLu:
If you attempt to display a large `rviz::Map`, and your graphics card is not super-powerful, it will downsample your map. 
```
[ WARN] [1487954079.336034022]: OGRE EXCEPTION(3:RenderingAPIException): Zero sized texture surface on texture MapTexture0 face 0 mipmap 0. Probably, the GL driver refused to create the texture. in GLTexture::_createSurfaceList at /build/buildd/ogre-1.8-1.8.1+dfsg/RenderSystems/GL/src/OgreGLTexture.cpp (line 420)
[ WARN] [1487954079.336585615]: Failed to create full-size map texture, likely because your graphics card does not support textures of size > 2048.  Downsampling to [2048 x 1314]...
```
world1

You can get around the limitations of your graphics card by creating multiple objects, each with their own texture, which I call Swatches. The result:
world2

Here is my sample data:
[the_world.png](https://cloud.githubusercontent.com/assets/1016143/23312007/20bec2d6-fa86-11e6-8683-2f760f0ea0ee.png)

[the_world.yaml.txt](https://github.com/ros-visualization/rviz/files/799934/the_world.yaml.txt)

### DIFF
--- a/src/rviz/default_plugin/map_display.cpp
+++ b/src/rviz/default_plugin/map_display.cpp
@@ -158,6 +158,11 @@ Swatch::Swatch(MapDisplay* parent, unsigned int x, unsigned int y, unsigned int 
 
 }
 
+Swatch::~Swatch()
+{
+    delete manual_object_;
+}
+
 void Swatch::updateAlpha(const Ogre::SceneBlendType sceneBlending, bool depthWrite, AlphaSetter* alpha_setter)
 {
   Ogre::Pass* pass = material_->getTechnique( 0 )->getPass( 0 );
@@ -478,7 +483,7 @@ void MapDisplay::updateAlpha()
   AlphaSetter alpha_setter( alpha );
 
   for (unsigned i=0; i < swatches.size(); i++) {
-    swatches[i].updateAlpha(sceneBlending, depthWrite, &alpha_setter);
+    swatches[i]->updateAlpha(sceneBlending, depthWrite, &alpha_setter);
   }
 }
 
@@ -489,13 +494,13 @@ void MapDisplay::updateDrawUnder()
   if( alpha_property_->getFloat() >= 0.9998 )
   {
     for (unsigned i=0; i < swatches.size(); i++)
-      swatches[i].material_->setDepthWriteEnabled( !draw_under );
+      swatches[i]->material_->setDepthWriteEnabled( !draw_under );
   }
 
   int group = draw_under ? Ogre::RENDER_QUEUE_4 : Ogre::RENDER_QUEUE_MAIN;
   for (unsigned i=0; i < swatches.size(); i++){
-    if( swatches[i].manual_object_ )
-      swatches[i].manual_object_->setRenderQueueGroup( group );
+    if( swatches[i]->manual_object_ )
+      swatches[i]->manual_object_->setRenderQueueGroup( group );
   }
 }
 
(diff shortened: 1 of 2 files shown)


DLu:
Thoughts, @dgossow @wjwwood ?

DLu:
Ping. This has been working fairly well for me, but I'd like to hear back from the experts.

wjwwood:
Sorry @DLu, I haven't had time to spend on rviz lately. I'll hopefully have some time to catch up in the coming weeks.

DLu:
I merged in some bug fixes. This has been running stable for me for the last year.

DLu:
Updated above description with additional information, sample data and screenshots.

wjwwood:
Thanks for the additional details @DLu, can you imagine anyway this might affect users of maps that are small or on the edge of needing this feature? Put another way, does this affect people who never ran into this issue or does this new code only come into play when the texture is too large?

If it does affect them, would it be prudent/possible to have an option in the display settings to explicitly disable this feature (revert to existing behavior always)?

DLu:
The way this PR works is that it attempts to load the texture into a single object initially. If that succeeds, it continues with the single object, just like before. Otherwise, it splits the texture along the longest dimension. 

So it should not affect any normal users. Just the fools who use enormous maps. 

My only fear is that I didn't do stuff right with the memory management. I was copying and pasting some patterns without really knowing what the underlying mechanism was, particularly the "manual object" stuff.

daizr:
@DLu thanks for this feature works fine for me w/ 15815 x 19384 ma and 2 swatches/tiles

wjwwood:
Ok, so it probably doesn't make sense to disable it.

I'll review your changes in more detail to double check the memory handling. I'll do this right before merging, but I'm going to work through all the pr's at the same time, in preparation for Lunar. That should be in the next few weeks. Thanks for your patience.

DLu:
Thanks @wjwwood. I just want it out there in the world, I'm not too particular as to the when.

src/rviz/default_plugin/map_display.cpp
```diff
@@ -59,9 +59,162 @@
...
+
+  // don't show map until the plugin is actually enabled
+  manual_object_->setVisible( false );
+
+
+}
+
+Swatch::~Swatch()
+{
+    delete manual_object_;
```
wjwwood: I think this should probably use http://www.ogre3d.org/docs/api/1.8/class_ogre_1_1_scene_manager.html#a75d15c3ac882f78f0001d9d007f403be

src/rviz/default_plugin/map_display.cpp
```diff
@@ -59,9 +59,162 @@
...
+  for(unsigned int yy=y_; yy<y_+height_;yy++){
+    int index = yy * fw + x_;
+    int pixels_to_copy = std::min((int)width_, N-index);
+    memcpy(ptr, &parent_->current_map_.data[ index ], pixels_to_copy); 
+    ptr+=pixels_to_copy;
+    if(index+pixels_to_copy>=N) break;
+  }
+
+  Ogre::DataStreamPtr pixel_stream;
+  pixel_stream.bind( new Ogre::MemoryDataStream( pixels, pixels_size ));
```
wjwwood: This looks ok, since `pixel_stream` is a shared pointer of Ogre's own making. The memory for the `MemoryDataStream` should be reclaimed correctly.

src/rviz/default_plugin/map_display.cpp
```diff
@@ -518,6 +580,59 @@ void MapDisplay::incomingUpdate(const map_msgs::OccupancyGridUpdate::ConstPtr& u
...
+  float resolution = current_map_.info.resolution;
+
+  int sw = width;
+  int sh = height;
+  int n_swatches = 1;
+
+  for(int i=0;i<4;i++){
+    ROS_INFO("Creating %d swatches", n_swatches);
+    for (unsigned i=0; i < swatches.size(); i++){
+        delete swatches[i];
```
wjwwood: This is the only place that swatches are deleted, but the destructor of this class doesn't call this function directly nor does it call it indirectly that I can see.

wjwwood:
Also, I cannot get it to fail on my GTX 1070. Apparently that is too awesome. I'll see if I can come up with an even larger image.

wjwwood:
closing in favor of https://github.com/ros-visualization/rviz/pull/1095

PR #1008, opened 2016-05-14
cmake qt5 cleanup

rhaschke:
Unify QT usage in cmake files for both Qt4 and Qt5 version by setting _common_ variables once.


wjwwood:
Thanks for the patch, I think it would a good improvement. Looks like CI is failing though, and I don't have time to address the issue at the moment, so it would be a while before I can fix it up.

rhaschke:
Oops. I missed one dependency. Sorry.

wjwwood:
Thanks, lgtm.

PR #1009, opened 2016-05-16
change rviz to no longer use opencv

wjwwood:
This would avoid issues with linking against OpenCV on Jessie which uses Qt4. It looks like the process may have already been done, but the dependencies were never removed?:

From https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/default_plugin/depth_cloud_mld.cpp#L137-L139:

>     // The following computation of center_x,y and fx,fy duplicates
    // code in the image_geometry package, but this avoids dependency
    // on OpenCV, which simplifies releasing rviz.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,6 @@ find_package(catkin REQUIRED
   angles
   cmake_modules
   geometry_msgs
-  image_geometry
   image_transport
   interactive_markers
   laser_geometry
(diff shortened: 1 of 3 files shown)


wjwwood:
@tfoote @vrabaud this resolves the Qt5/Qt4 issues with rviz on Jessie in my local VM. Turns out rviz doesn't use `image_geometry` any more and was unnecessarily linking against it...

See https://github.com/ros/rosdistro/issues/11488#issuecomment-218596122

tfoote:
that's convenient!

PR #1010, opened 2016-05-17
Fix compile error due to the user-defined string literals feature

kartikmohta:
While compiling rviz from source on Arch with gcc 6.1.1, got errors like the following:
```cpp
In file included from /opt/ros/kinetic/include/tf/transform_datatypes.h:44:0,
                 from /opt/ros/kinetic/include/tf/time_cache.h:38,
                 from /opt/ros/kinetic/include/tf/tf.h:43,
                 from /opt/ros/kinetic/include/tf/transform_listener.h:38,
                 from [..]/src/rviz/src/rviz/default_plugin/effort_display.cpp:4:
[..]/src/rviz/src/rviz/default_plugin/effort_display.h: In destructor ‘virtual tf::MessageFilterJointState::~MessageFilterJointState()’:
[..]/src/rviz/src/rviz/default_plugin/effort_display.h:39:37: error: unable to find string literal operator ‘operator""fmt’ with ‘const char [28]’, ‘long unsigned int’ arguments
   ROS_DEBUG_NAMED("message_filter", "MessageFilter [target=%s]: "fmt, getTargetFramesString().c_str(), __VA_ARGS__)
                                     ^
```
With the new C++11 user-defined string literals feature, there has to be a space between string concatenations like the ones used here. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51282.

### DIFF
--- a/src/rviz/default_plugin/effort_display.h
+++ b/src/rviz/default_plugin/effort_display.h
@@ -36,13 +36,13 @@ namespace tf
 # undef TF_MESSAGEFILTER_DEBUG
 #endif
 #define TF_MESSAGEFILTER_DEBUG(fmt, ...) \
-  ROS_DEBUG_NAMED("message_filter", "MessageFilter [target=%s]: "fmt, getTargetFramesString().c_str(), __VA_ARGS__)
+  ROS_DEBUG_NAMED("message_filter", "MessageFilter [target=%s]: " fmt, getTargetFramesString().c_str(), __VA_ARGS__)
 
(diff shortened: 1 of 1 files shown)


de-vri-es:
One more time then:

Note that GCC 6 compiles with C++11 support enabled by default, so this is required to compile rviz from source on for example Arch Linux. Since this is also true for the inidigo version, I think it is a good idea to backport it to indigo-devel.

bchretien:
:+1: I added the patch to the PKGBUILD for now.

PR #1011, opened 2016-05-17
always (re)create default_plugin_location.cmake

rhaschke:
I had some issues switching between `kinetic` and `indigo` branch in my workspace: Both use different names of the `default_plugin` lib, written by the custom target `create_default_plugin_location_file`.
However, cmake's `add_custom_command` only evokes the file-writing command if the output file to be created is not yet there (or older than dependencies). So, it's never rewritten as there is no dependency.
Using `COMMAND` directly in `add_custom_target` solves that issue.
The same patch should be applied to `kinetic`.

### DIFF
--- a/src/rviz/default_plugin/CMakeLists.txt
+++ b/src/rviz/default_plugin/CMakeLists.txt
@@ -74,16 +74,12 @@ install(TARGETS default_plugin
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION})
 
 # Generate to the devel space so the extras file can include it from the devel space.
-add_custom_command(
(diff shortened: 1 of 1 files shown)


wjwwood:
Are you also switching Qt versions when you do that or are you using Qt 4 with Kinetic or Qt 5 with Indigo?

I'll take the improvement, but I'd shy away from building the indigo and kinetic version in the same workspace without cleaning first.

rhaschke:
I'm running Indigo on Trusty only. But I switch rviz' kinetic-devel and indigo-devel branches frequently and use them with either Qt4 or Qt5 (changing the UseQt5 option). Doing so, in a workspace works just fine: rviz rebuilds against the other Qt version when switched and all dependencies fetch the correct Qt version from rviz and rebuild too. I never observed that multiple Qt versions were linked together.

PR #1012, opened 2016-05-19
in the new c++11 a space between string concatenation is mandatory

sekuch:


de-vri-es:
Note that GCC 6 compiles with C++11 support enabled by default, so this is required to compile rviz from source on for example Arch Linux. Since this is also true for the inidigo version, I think it is a good idea to backport it to `indigo-devel`.

de-vri-es:
Oh, #1010 already exists too

PR #1013, opened 2016-05-20
Prevent format sring being interpreted as user-defined string literal.

de-vri-es:
C++11 introduced user defined string literals, and one of the macros used by the `effort_display` plugin is now getting parsed as one. The fix is easy: separate the offending string literal and identifier with a space.

GCC 6 compiled with C++11 support on by default, which means that this is not just an improvement for the future but a requirement to compile from source on for example Arch Linux. This is also true when compiling the indigo version, so back-porting to indigo-devel would be prudent I think (I tried a simple cherry pick, it applies without problems).


wjwwood:
https://github.com/ros-visualization/rviz/pull/1012?

de-vri-es:
Ah, yes, that is the same. I'll close this then :p

PR #1016, opened 2016-05-29
Add a feature to display pose markers(axes or arrow)

longjie:
Hi,
This PR is for adding pose markers (axes or arrow) to Path display.
I personally wanted this feature to visualize the planned path including the pose of each waypoint.
Please merge if anyone needs this and the patch is acceptable.
Thank you,


NikolausDemmel:
I didn't look at the patch, but just wanted to say that this addition would be very useful to us. If fact, we are using a visualization node together with the Marker Display to achieve something similar for Path messages. Having this ability directly in the Path Disaply would be much nicer. So thanks for the contribution!

wjwwood:
I tried it out locally and it seems to work as advertised. Thanks for the contribution!

hiddewie:
Great! Thanks @NikolausDemmel and @wjwwood for the quick feedback!

Issue #743, opened 2014-03-27
Fullscreen CameraDisplay


jbohren:
For a long time now I've liked to have a fullscreen RViz camera display. I took a look at the calls that the Oculus plugin makes to make its window fullscreen, but the same doesn't seem to work with ImageDisplay or CameraDisplay. 

I feel like this shouldn't be hard, but I'm not too familiar with the RViz, Qt, and Ogre APIs, so I don't know what the best approach is. If it's something simple, and someone could give me some guidance, I could do it and post a PR.

PR #1017, opened 2016-05-29
Add fullscreen option.

de-vri-es:
This PR adds an option to enter fullscreen mode, hiding all interface elements and showing nothing but the rendered view. It makes rviz much nicer to use for showing information in demo setups.

I added the option under the `view_menu_` in code, but it turns out it's actually called "Panels" in the GUI. That might make it the wrong place for this option. Alternatively the menu could be renamed to view, or a new view menu could be added.

F11 also works, and is the only way other than restarting rviz to leave fullscreen mode. Might make sense to also use Escape for that?

One point of note: I'm using `setHideButtonVisibility(bool)` which is a public member of `VisualizationFrame`. I couldn't find anything actually calling the member, but if it is exposed to plugins they could potentially interact badly with the fullscreen option. It seems unlikely though, so maybe it's better to make that member private?

### DIFF
--- a/src/rviz/visualization_frame.h
+++ b/src/rviz/visualization_frame.h
@@ -168,6 +168,9 @@ Q_SIGNALS:
   /** @brief Emitted during file-loading and initialization to indicate progress. */
   void statusUpdate( const QString& message );
 
+  /** @brief Emitted when the interface enters or leaves full screen mode. */
+  void fullScreenChange( bool hidden );
+
 protected Q_SLOTS:
   void onOpen();
   void onSave();
@@ -218,6 +221,12 @@ protected Q_SLOTS:
    * The sender() of the signal should be a QAction whose text() is
    * the name of the panel. */
   void onDeletePanel();
+
+  /** @brief Set full screen mode. */
+  void setFullScreen( bool full_screen );
+
+  /** @brief Exit full screen mode. */
+  void exitFullScreen();
 
 protected Q_SLOTS:
   /** @brief Set loading_ to false. */
@@ -299,7 +308,6 @@ protected:
   QMenu* view_menu_;
   QMenu* delete_view_menu_;
   QMenu* plugins_menu_;
-  QList<QAction*> view_menu_actions_;
 
   QToolBar* toolbar_;
 
@@ -350,6 +358,9 @@ protected:
   ros::WallTime last_fps_calc_time_;
 
   QString error_message_; ///< Error message (if any) from most recent saveDisplayConfig() call.
+
+  /// Indicates if the toolbar should be visible outside of fullscreen mode.
+  bool toolbar_visible_;
 };
 
 }
--- a/src/rviz/panel_dock_widget.cpp
+++ b/src/rviz/panel_dock_widget.cpp
@@ -100,13 +100,13 @@ void PanelDockWidget::setCollapsed( bool collapse )
   {
     if ( isVisible() )
     {
-      QDockWidget::setVisible( false );
+      PanelDockWidget::setVisible( false );
       collapsed_ = collapse;
     }
   }
   else
   {
-    QDockWidget::setVisible( true );
+    PanelDockWidget::setVisible( true );
     collapsed_ = collapse;
   }
 }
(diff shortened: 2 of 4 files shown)


de-vri-es:
Oh, I tested on a system runing xmonad without window decorations, so I'm not 100% that `showFullScreen()` removes the window decoration. I kind of assumed it does.

de-vri-es:
Updated to use signals/slot for notifying `PanelDockWidgets` of full screen changes to decouple them from the `VisualizationFrame` again.

wjwwood:

wjwwood:

VictorLamoine:
I tested this PR and it works perfectly.

Minor issue: the maximized window state is not retrieved when exiting full screen (the window is not maximized anymore).

Really useful for demos/screenshots.

:+1:

de-vri-es:
I'll see if there is a sensible way to retain maximized state. Thanks for testing!

de-vri-es:
By the way, escape does work to exit full screen mode.

wjwwood:
Did you ever find a way to preserve the maximized state?

src/rviz/visualization_frame.h
```diff
@@ -350,6 +358,9 @@ protected Q_SLOTS:
   ros::WallTime last_fps_calc_time_;
 
   QString error_message_; ///< Error message (if any) from most recent saveDisplayConfig() call.
+
+  /// Indicates if the toolbar should be visible outside of fullscreen mode.
+  bool toolbar_visibile_;
```
VictorLamoine: Typo, `visible`

VictorLamoine:
1 - Retrieve if the window was [maximized]( http://doc.qt.io/qt-5.7/qwidget.html#maximized-prop) before going full screen
2 - When returning from full screen, call [`showMaximized`](http://doc.qt.io/qt-4.8/qwidget.html#showMaximized) instread of `showNormal` if the window was maximized before.

The problem is that the `maximized` property does not work with X11. I tested and it is always zero.

de-vri-es:
There's also `QtWidget::saveGeometry()` and `QtWidget::restoreGeometry()`. Not sure how reliable that will be, but it's worth a try. Note that I'm currently on holiday, so earliest I can try it is somewhere next week.

VictorLamoine:
Fixes #743

de-vri-es:
Digging through the Qt documentation once more, I found this method of setting/disabling fullscreen:
```
if (full_screen)
  setWindowState(windowState() | Qt::WindowFullScreen);
else
  setWindowState(windowState() & ~Qt::WindowFullScreen);
show();
```

This preserves all other flags such as minimized (not applicable here I think) and maximized state. I tested with a `xfwm4` as window manager, and it worked as expected. Window state is preserved between entering and leaving fullscreen.

Updated and rebased the PR.

But note that I don't think it fixes #743. All widgets except for the 3D display are hidden in full-screen mode. That probably includes camera displays.

wjwwood:
Cool, I'll have another pass at it asap.

src/rviz/visualization_frame.h
de-vri-es: Thanks btw, fixed.

de-vri-es:
No rush, I haven't been rushing either, but have you had the time already?

wjwwood:
I'll come around to it before the Lunar release of rviz.

de-vri-es:
Heh, shooting for the moon. Anyway, thanks :)

VictorLamoine:
I tested again and it still works great.

The issue with the full-screen mode not being kept is still here but does not happen the first time. I think this really is minor and should not prevent the merging of this PR.

Here is a nice video:
https://drive.google.com/open?id=0B4_0QfSy1tFqa29fMHZCTWwxbHc

de-vri-es:
> The issue with the full-screen mode not being kept is still here but does not happen the first time. I think this really is minor and should not prevent the merging of this PR.

That is weird. I tested with `xfwm` and the previous version did mess up maximized state after leaving full screen, but the current did not.

Though I agree its something that shouldn't be a big problem for merging. Since it's a problem with Qt / X11 I actually think it's better to ignore it. The proper place to fix it is there. Any workaround we introduce may end up breaking if Qt fixes the issue on their end, and we end up with legacy code which nobody really remembers the purpose of.

wjwwood:
Finally got around to testing this. Looks good, thanks!

wxmerkt:
Awesome, thanks! Will this be backported to indigo and released into the debians?

wjwwood:
@wxmerkt I was just going to put it into Kinetic and Lunar, but since you asked nicely (and it seems low risk) I'll back port it to Indigo too.

de-vri-es:
hmm, maybe I shouldn't have removed the branch? Yay, github has a restore branch button.

Thanks for merging.

wjwwood:
@de-vri-es that's ok, I'll manually handle cherry-picking for forward/back porting.

wxmerkt:
Thank you @wjwwood :)

wjwwood:

PR #1019, opened 2016-06-03
Fix compilation warnings

de-vri-es:
This PR fixes compilation warnings (all of them with gcc 6.1.1 20160501).

I was a bit confused by the existence of `setUserData` but the absence of `getUserData`. It seems like a pointless member functions without a matching getter. I changed `setUserData` to not use the deprecated `Ogre::MoveableObject::SetUserAny`, and when I wanted to update the getter I noticed it wasn't there.

### DIFF
--- a/src/rviz/display.h
+++ b/src/rviz/display.h
@@ -113,7 +113,11 @@ public:
    *  @param topic The published topic to be visualized.
    *  @param datatype The datatype of the topic.
    */
-  virtual void setTopic( const QString &topic, const QString &datatype ) { }
+  virtual void setTopic( const QString &topic, const QString &datatype )
+  {
+    (void) topic;
+    (void) datatype;
+  }
 
   /** @brief Return true if this Display is enabled, false if not. */
   bool isEnabled() const;
@@ -124,7 +128,11 @@ public:
   /** @brief Called periodically by the visualization manager.
    * @param wall_dt Wall-clock time, in seconds, since the last time the update list was run through.
    * @param ros_dt ROS time, in seconds, since the last time the update list was run through. */
-  virtual void update( float wall_dt, float ros_dt ) {}
+  virtual void update( float wall_dt, float ros_dt )
+  {
+    (void) wall_dt;
+    (void) ros_dt;
+  }
 
   /** @brief Called to tell the display to clear its state */
   virtual void reset();
--- a/src/rviz/external/resource_retriever.h
+++ b/src/rviz/external/resource_retriever.h
@@ -1,0 +1,7 @@
+#ifndef RVIZ_EXTERNAL_RESOURCE_RETRIEVER_H
+#define RVIZ_EXTERNAL_RESOURCE_RETRIEVER_H
+
+#pragma GCC system_header
+#include <resource_retriever/retriever.h>
+
+#endif
--- a/src/rviz/mesh_loader.cpp
+++ b/src/rviz/mesh_loader.cpp
@@ -28,7 +28,7 @@
  */
 
 #include "mesh_loader.h"
-#include <resource_retriever/retriever.h>
+#include "external/resource_retriever.h"
 
 #include <boost/filesystem.hpp>
 
--- a/src/rviz/ogre_helpers/grid.cpp
+++ b/src/rviz/ogre_helpers/grid.cpp
@@ -240,7 +240,7 @@ void Grid::create()
 
 void Grid::setUserData( const Ogre::Any& data )
 {
-  manual_object_->setUserAny( data );
+  manual_object_->getUserObjectBindings().setUserAny( data );
 }
 
 } // namespace rviz
--- a/src/rviz/ogre_helpers/line.cpp
+++ b/src/rviz/ogre_helpers/line.cpp
@@ -145,7 +145,7 @@ const Ogre::Quaternion& Line::getOrientation()
 
 void Line::setUserData( const Ogre::Any& data )
 {
-  manual_object_->setUserAny( data );
+  manual_object_->getUserObjectBindings().setUserAny( data );
 }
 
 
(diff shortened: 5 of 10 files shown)


de-vri-es:
Note that all jenkin builds are marked unstable until the warnings are fixed. I think that policy is fine, but then there should be no warnings on the devel branch (which this PR addresses).

CMakeLists.txt
```diff
@@ -204,7 +204,7 @@ include_directories(SYSTEM
   ${OPENGL_INCLUDE_DIR}
   ${PYTHON_INCLUDE_PATH}
 )
-include_directories(src ${catkin_INCLUDE_DIRS})
+include_directories(src SYSTEM ${catkin_INCLUDE_DIRS})
```
wjwwood: I don't want to use `SYSTEM`, it's known to cause all sorts of problems with the include order on some OS's.

Please use the diagnostics pragma's to disable the particular warnings when including the problem headers. See:

https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html

Here's an example of it in use:

https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/src/functions.cpp#L26-L40

de-vri-es: Could you elaborate on that? I have never had problems with marking include directories as system directories and it is the cleanest way for the compiler to ignore errors in external headers. It is the least compiler specific and it does not depend on the exact contents of the external headers, which is not under control of this package (so can change and fix or add new warnings).

wjwwood: Sure, the `SYSTEM` directive ends up using `-isystem` as the option to the compiler. The `-isystem` paths are always looked at after `-I`, and if a path is specified with both, the `-I` entry is ignored:

From https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html:
> All directories named by -isystem are searched after all directories named by -I, no matter what their order was on the command line. If the same directory is named by both -I and -isystem, the -I option is ignored.

So that means that when you use it with header paths that may be coming from a workspace, like `~/my_ws/install/include`, and from somewhere else you have an `-I/opt/ros/kinetic/include`, then you can get in a situation where you use the headers from `/opt/ros` rather than the headers from `~/my_ws/install` even when the workspace should take precedence. Also `catkin_INCLUDE_DIRS` is sorted by workspace prefix to avoid a similar problem, so adding these include unevenly may result in unexpected, and hard to debug, behavior when using a workspace.

Also, on OS X, the `isystem` flag is used to control which SDK you're using and it caused problems in the past to have more instances of that, though I think CMake/Apple fixed that at some point.

I think in this case it would ok to use it since we're putting all headers under `SYSTEM`, but because of the subtle issues with `isystem` I don't want to introduce it in case someone adds more headers later without understanding the pitfall. If you don't want to take the time to suppress the warns with the `#pragma`'s, then I'll take care of it. Thanks for the fixes you've provided so far!

de-vri-es: I see, I did not take into account multiple workspaces. That issue is unfortunate, as it doesn't seem like something that can easily be worked around.

My main hesitation with using pragmas is that they should be maintained properly to stay in sync when new warnings are added by the external dependency. Also, removing unneeded pragmas becomes difficult since it would mean reintroducing warnings when compiling with older versions of the external dependency. On the other hand `-isystem` more clearly communicates the intent ("these headers are not mine, don't blame me") and doesn't require maintenance.

Is there perhaps a pragma for gcc and clang which has the same effect on an included header as `-isystem`? That would fix the problem without adding a maintenance burden while avoiding include path order issues.

de-vri-es: Hmm, there is this:

```
#pragma GCC system_header
    This pragma takes no arguments. It causes the rest of the code in the current file to be treated as if it came from a system header. See System Headers. 
```
But the rest of the file is too broad. If that could be turned off again it would be perfect...

wjwwood: You can create a "proxy" header which has that pragma and includes the other header.

de-vri-es: Yeah, I was thinking that too. On the one hand it seems a bit overkill, but on the other hand a separate wrapper header would be the best way anyway if the offending include is used multiple times. And then `#pragma GCC system_header` is viable option again.

I'll go for that I suppose =]

de-vri-es: There is still a larger maintenance burden of course, since external dependencies can still introduce warnings in other headers. I'm now wondering again if including everything external with `SYSTEM` isn't actually the best option.

An additional complication is that I don't get any warnings from external dependencies locally, even without the `CMakeLists.txt` modifications and a clean configure+build. I can go by the log from Jenkins of course, but it does illustrate the problem with manually disabling warnings from specific includes.

wjwwood: Well, as I said, I'm not interested in maintaining the use of `SYSTEM`, but I'll take on the work of maintaining suppression of warnings from other headers. I can't take care of this today, but perhaps this weekend I can add what's needed to your PR here and merge it.

de-vri-es: Fair enough. I'll remove the `CMakeLists.txt` changes commit.

wjwwood: Thanks.

wjwwood:
Yes, they should probably export the flag in that case. As of Kinetic they're allowed to use C++11 if they want.

de-vri-es:
Oh, and of course another option to get rid of the `resource_retriever` warning is to compile with C++11 support. Then the crufty wrapper header can be removed, and even better, we get C++11 :)

wjwwood:
Ok, I undid the fake system header commit and turned on c++11 (faf97be) and then merged to master.

wjwwood:
Thanks for the contribution and your patience.

de-vri-es:
You too ;)

And great to have C++11 in rviz

PR #1021, opened 2016-06-07
Tweak focal shape size marker depending on focal distance

### LINKED ISSUE #1020: focal_shape_ size should depend on distance_property_
When using RViz and zooming on very small objects in RViz the `focal_shape_` (the yellow squeezed sphere on the screenshot) becomes very big and clutters the view.
The inverse happens when zooming out, we can't see the focal shape anymore.

focal_shape_

We should resize the sphere as the user zooms in/out so that it always remains visible but never too big.

https://github.com/ros-visualization/rviz/blob/indigo-devel/src/rviz/default_plugin/view_controllers/orbit_view_controller.cpp#L80

VictorLamoine:
Fixes #1020 

This PR adds two options:
- Setting the focal shape size.
- A zoom invariant mode so that the focal shape size does not change when zooming in/out.

~~The only problem with this PR is the initial shape size; I found no way of knowing the focal distance before the user interacts (zooms in or out) so the shape size at the start is a little bit arbitrary.~~

Fixed size:
---
old

Zoom invariant:
---
new

### DIFF
--- a/src/rviz/default_plugin/view_controllers/orbit_view_controller.cpp
+++ b/src/rviz/default_plugin/view_controllers/orbit_view_controller.cpp
@@ -39,6 +39,7 @@
 #include "rviz/display_context.h"
 #include "rviz/geometry.h"
 #include "rviz/ogre_helpers/shape.h"
+#include "rviz/properties/bool_property.h"
 #include "rviz/properties/float_property.h"
 #include "rviz/properties/vector_property.h"
 #include "rviz/uniform_string_stream.h"
@@ -51,6 +52,8 @@
 static const float PITCH_START = Ogre::Math::HALF_PI / 2.0;
 static const float YAW_START = Ogre::Math::HALF_PI * 0.5;
 static const float DISTANCE_START = 10;
+static const float FOCAL_SHAPE_SIZE_START = 0.05;
+static const bool FOCAL_SHAPE_FIXED_SIZE = true;
 
 namespace rviz
 {
@@ -60,6 +63,11 @@ OrbitViewController::OrbitViewController()
 {
   distance_property_ = new FloatProperty( "Distance", DISTANCE_START, "Distance from the focal point.", this );
   distance_property_->setMin( 0.01 );
+
+  focal_shape_size_property_ = new FloatProperty( "Focal Shape Size", FOCAL_SHAPE_SIZE_START, "Focal shape size.", this );
+  focal_shape_size_property_->setMin( 0.001 );
+
+  focal_shape_fixed_size_property_ = new BoolProperty ( "Focal Shape Fixed Size", FOCAL_SHAPE_FIXED_SIZE, "Focal shape size.", this );
 
   yaw_property_ = new FloatProperty( "Yaw", YAW_START, "Rotation of the camera around the Z (up) axis.", this );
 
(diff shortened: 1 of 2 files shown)


trainman419:
When zooming or panning an empty world, the focal shape is often my only visual cue about what the current zoom level is. Having the focal shape scale dynamically counteracts this.

A simpler approach that maintains the functionality that I use on a regular basis would be to make the size of the focal shape settable by the user.

de-vri-es:
Or both, and a checkbox to correct for zoom level (the size should then be the size at 100% zoom)

wjwwood:
@VictorLamoine makes sense to me, but I also see @trainman419's point of view. How hard would it be to add an option to the display menu? Are you interested in doing that?

Also, I added gif's of your before and after to your OP.

VictorLamoine:
I don't know how hard it can be because I don't know the code, but I will at least give it a try!

I will try to add an option in the display menu:
- Fixed sized focal shape
- Zoom-invariant focal shape

For both of them we will be able to choose a size for the shape.

VictorLamoine:
> How hard would it be to add an option to the display menu?

You probably meant the `Views` panel, right?

I added a user option to be able to choose the shape size, I still need to add a menu allowing to choose between fixed size/zoom invariant.

wjwwood:
I was originally meaning the display panel, which has the displayed types and the grid and stuff like that, but I think the View panel makes more sense now that you mention it.

VictorLamoine:
Ready for review!

view_panel

wjwwood:
Thanks! I'll have a look at it this weekend.

wjwwood:

wjwwood:

VictorLamoine:
Friendly ping :pig:

wjwwood:
Just tried it out, but I've noticed that the changes in the settings do not take effect until you change the zoom:

rviz_focal_point

VictorLamoine:
There are two things that can be improved:

- Behavior at start-up: the `distance_property_` is not set until the user zooms in or out so there is no way of setting the right focal shape size on start-up
- When the user changes the focal shape size it is not updated until the user zooms in/out.

For the first item we need to find a way to correctly initialize the distance property value.

For the second item we could probably connect the `Focal Shape Size` `valueChanged()` signal to a `updateFocalShapeSize` slot. (same applies to the check box).

I will look at that and get back to you!

wjwwood:
> I will look at that and get back to you!

Sounds like a plan. Let me know if I can help with questions or testing.

VictorLamoine:
All problems are solved; updating the focal shape in mouse events solves the two problems.

VictorLamoine:
@wjwwood friendly ping :pig: Is this repository on "pause"?

wjwwood:
Sorry for the delay, I'll just tried it out real quick and it looks great now. The only thing is that on my machine the pip is now teal or blueish. Did you change that too?

wjwwood:
It doesn't look like it... If I don't use the navigation tutorial to launch rviz it is yellow... That must be a different bug.

wjwwood:
Looks like this happens on kinetic-devel too, so it won't block this pr.

Thanks again for the contribution, it's a great feature!

wjwwood:
https://github.com/ros-visualization/rviz/issues/1056

PR #1023, opened 2016-06-10
Add libX11 to librviz target_link_libraries

kartikmohta:
With strict ld flags, I get the following errors:
```
render_system.cpp: undefined reference to `XOpenDisplay'
render_system.cpp: undefined reference to `XCreateSimpleWindow'
```
There is a dependency on libX11 in ogre_helpers/render_system.cpp.

### DIFF
--- a/src/rviz/CMakeLists.txt
+++ b/src/rviz/CMakeLists.txt
@@ -4,6 +4,8 @@ if(NEW_YAMLCPP_FOUND)
   add_definitions(-DRVIZ_HAVE_YAMLCPP_05)
 endif(NEW_YAMLCPP_FOUND)
 
+find_package(X11 REQUIRED)
+
 add_subdirectory(default_plugin)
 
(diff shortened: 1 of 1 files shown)


src/rviz/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ if(NEW_YAMLCPP_FOUND)
   add_definitions(-DRVIZ_HAVE_YAMLCPP_05)
 endif(NEW_YAMLCPP_FOUND)
 
+find_package(X11 REQUIRED)
```
wjwwood: This needs to be wrapped in a `if(UNIX AND NOT APPLE)` block, as the `find_package(X11 ...)` is not required on OS X.

wjwwood:

wjwwood:

wjwwood:
lgtm, thanks for the contribution.

PR #1024, opened 2016-06-10
Allow disabling RViz splash-screen

VictorLamoine:
I hate splash screens.

`setSplashPath` explains that the user can disable the splash screen by setting an empty path:
https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/visualization_frame.h#L87-L91

But the user can't because if the splash screen path argument is empty, the function is not called:
https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/visualizer_app.cpp#L277-L280

This PR fixes the behavior:
```
rosrun rviz rviz -s ""
```


wjwwood:
retest this please

wjwwood:

wjwwood:
Lgtm, thanks!

PR #1029, opened 2016-07-06
Visualize joint axis and display type and limits

lucasw:
Show joint properties in left hand panel and optionally show joint axis as arrow in the 3d view.

### DIFF
--- a/src/rviz/robot/robot_joint.cpp
+++ b/src/rviz/robot/robot_joint.cpp
@@ -412,7 +412,7 @@ void RobotJoint::updateAxes()
 
 void RobotJoint::updateAxis()
 {
-  if( axis_property_->getValue().toBool() )
+  if( show_axis_property_->getValue().toBool() )
   {
     if( !axis_ )
     {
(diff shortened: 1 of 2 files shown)


wjwwood:
Can you post some screenshots which show the behavior you've added. I believe I know what you intend from the code changes, but trying it locally I'm not seeing what I would expect. It would also be helpful if you could describe how you test it out so that we're using the same baseline for comparison.

lucasw:
Load a robot model with joints that have axes to the parameter server, then add a RobotModel display to rviz, then expand `Links`, change `Link Tree Style` to `Joints in Alphabetic Order` or `Tree of links and joints`, then find a continuous, revolute, prismatic, or planar joint, then click the `Show Joint Axis` checkbox to show a green arrow for the joint.

joint_axis

... and it looks like unchecking the box doesn't make the arrow disappear, I need to fix that.

lucasw:
Using rrbot from gazebo_ros_demos:

    roslaunch rrbot_description rrbot_rviz.launch 

joint_axis2

wjwwood:
Ok, I was missing the part where you have to select a different tree view mode. lgtm

PR #1031, opened 2016-07-16
Rotate image with transformation from image frame to target one

wkentaro:
rviz_image_rotate_display


wjwwood:
@wkentaro thanks for the pull request. I recently removed a dependency on OpenCV within rviz. It would be great to preserve this. Is there a way to do this which does not use OpenCV? It seems like a naive 2D rotation could be done using QImage and QTransform.

wkentaro:
I see.

I've worked around a bit in https://github.com/wkentaro/rviz/commits/feature/image-rotate-without-opencv but got segv with current implement.
It segfaults at converting ROS image msg to QImage.

wjwwood:
Unfortunately I'll have to insist that we keep OpenCV out of the dependencies for now. If you or anyone else has time to resurrect this using QImage that would be great. Until then I'll leave this here as untargeted.

PR #1039, opened 2016-08-15
robot: check for root_link_ before use

v4hn:
A couple of lines below the code accesses link->getChildJointNames() without
testing "link". The additional test guarantees the link is set before working
with it.

I noticed that when adding support to MoveIt to correctly initialize an empty
robot model after printing an error.


wjwwood:
lgtm, thanks.

PR #1041, opened 2016-08-22
robot: check for root_link_ before use (kinetic port of #1039)

v4hn:
I noticed that @wjwwood added "needs forward port" to #1039.
If he'd just asked I would have provided the request directly...

### DIFF
--- a/src/rviz/robot/robot.cpp
+++ b/src/rviz/robot/robot.cpp
@@ -682,7 +682,14 @@ void Robot::calculateJointCheckboxes()
 
   // check root link
   RobotLink *link = root_link_;
-  if (link && link->hasGeometry())
+
+  if(!link)
+  {
+    setEnableAllLinksCheckbox(QVariant());
+    return;
+  }
+
+  if (link->hasGeometry())
   {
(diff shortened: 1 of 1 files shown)


trainman419:
It's roughly equal effort for @wjwwood to cherry-pick patches made on an older branch onto the kinetic branch or for him to handle a pull request. I think he usually does the forward ports in batches as preparation for a new release of rviz.

This isn't an underhanded or passive-aggressive request for a forward port; it's a note-to-self that #1039 needs to be cherry-picked in before the next kinect release.

This pull request is still fine if you'd like to see #1039 incorporated into the kinetic branch sooner.

v4hn:
On Wed, Aug 24, 2016 at 02:41:50PM -0700, Austin wrote:
> This pull request is still fine if you'd like to see #1039 incorporated into the kinetic branch sooner.

That was the hope, yes.
On the other hand, one of our patches in MoveIt depends on this request being actually released in RViz.
So we wait for the release either way.

Thanks for the explanation Austin.

wjwwood:
Yeah, sorry I just put those labels on so I can handle it when doing a release. Thanks for doing for me though.

Issue #1025, opened 2016-06-10
Cannot set float parameters in RViz if locale decimal mark is not dot.


VictorLamoine:
I'm using the French locale and our decimal mark is the comma, not the dot. ([like many other countries](https://upload.wikimedia.org/wikipedia/commons/thumb/a/a8/DecimalSeparator.svg/1024px-DecimalSeparator.svg.png?1465589334705))

In RViz when trying to set the near clip distance in the Views panel; I cannot enter values between 0 and 1. (eg: `0,56` becomes `56`).

This can be easily debugged by adding a `cout` in the code ([here](https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/view_controller.cpp#L253-L254)):
```
  std::cout << "Near clip distance = " << n << std::endl;
```
Install the french locale and switch to it:
```
export LANG=fr_FR.UTF-8
```

The issue is solved on my machine when using `export LANG=C` before running RViz in the terminal.

Something like this should help:
```
doublespinbox->setLocale(QLocale(QLocale::English));
```

wjwwood:
@VictorLamoine as always, these things tend to favor the environment in which they were originally developed. I'm not familiar with the locale mechanism used to handle this difference, but I'm happy to review pull requests that improve the portability of how we interpret decimals.

wjwwood:
I think this should be resolved in https://github.com/ros-visualization/rviz/pull/1043. I tried rviz by setting my locale to fr_FR and I had to use `,` as the decimal delimiter.

PR #1043, opened 2016-08-29
bugfix in float_edit for non-english locales 

1r0b1n0:
Hi,

The float property line edit doesn't work on systems with a decimal comma separator locale. For example on Ubuntu 16.04, with ros kinetic and a french locale, it is impossible to enter a decimal value.
This pull requested fixes the problem.

Another possible fix would be to force the QDoubleValidator to use the "C" locale : 

> QDoubleValidator *validator = new QDoubleValidator(this);
> validator->setLocale(QLocale::c());
> setValidator(validator);

### DIFF
--- a/src/rviz/properties/float_edit.cpp
+++ b/src/rviz/properties/float_edit.cpp
@@ -28,6 +28,7 @@
  */
 
 #include <QDoubleValidator>
+#include <QLocale>
 
 #include "rviz/properties/float_edit.h"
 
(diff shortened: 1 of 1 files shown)


wjwwood:
I guess this would close https://github.com/ros-visualization/rviz/issues/1025?

1r0b1n0:
@wjwwood yes this fixes the problem of #1025

wjwwood:
Backport to Indigo: https://github.com/ros-visualization/rviz/pull/1058

PR #1044, opened 2016-09-02
Use urdf::*ShredPtr instead of boost::shared_ptr

jspricke:
urdfdom_headers uses C++ std::shared_ptr. As it exports it as custom
*SharedPtr type, we can use the to sty compatible.

This also adds a proper dependency for urdfdom-headers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,8 @@ find_package(Boost REQUIRED
   thread
 )
 
+find_package(urdfdom_headers REQUIRED)
+
 find_package(PkgConfig REQUIRED)
 
 find_package(ASSIMP QUIET)
@@ -131,6 +133,7 @@ find_package(catkin REQUIRED
   tf
   urdf
   visualization_msgs
+  urdfdom_headers
 )
 
 if(${tf_VERSION} VERSION_LESS "1.11.3")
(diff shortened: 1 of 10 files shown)


wjwwood:
It doesn't look like `urdfdom-headers-dev` is a valid rosdep key, perhaps you meant `lib urdfdom-headers-dev`? Can you update it?

I wonder why the urdf types were forward declared and if that was necessary. Other than that lingering curiosity, the changes look reasonable to me. Thanks!

jspricke:
Fixed the name, thanks for spotting. I was wondering about the forward declarations as well, but they are older then the git, guess it's not needed anymore.

rhaschke:
@jspricke You should explicitly require urdfdom_headers >= 0.4. Before, the typedefs are not available.

jspricke:
Hi Robert, this one is merged already, so feel free to open a new pull
request if needed ;).

wjwwood:
Sorry, I didn't follow up on @rhaschke's comment. Indeed this breaks the build on Wily and Debian Jessie. I guess I need to revert this change or apply a fix like @rhaschke did for `srdfdom` in https://github.com/ros-planning/srdfdom/pull/25.

wjwwood:
I think I'm going to revert this and open an issue to resolve it in the future. I'm not sure it makes sense to replicate the work in https://github.com/ros-planning/srdfdom/pull/25 in rviz and every other ROS package that uses urdf. Perhaps this compatibility shim should be put in into `robot_model`'s `urdf` package so it can be consumed by other ROS packages no matter which version of `urdfdom` is under the hood.

wjwwood:
https://github.com/ros-visualization/rviz/issues/1061

PR #1053, opened 2016-10-11
Fix double free in display dialog

at-wat:
This commit deletes the `AddDisplayDialog` object after use.

Copy operator of `QString` in `rviz::AddDisplayDialog::updateDisplay()` causes double free AFTER CLOSING THE DIALOG.
Since `AddDisplayDialog` object created in `DisplaysPanel::onNewDisplay` never deleted, local variable `topic_output` could be overwritten after leaving local scope.
This seems to have occurred especially in many-core (16-physical-core in my case) environments.


wjwwood:
Indigo backport: https://github.com/ros-visualization/rviz/pull/1057

PR #1062, opened 2016-10-20
WIP: Improve performance by reducing materials update rate

VictorLamoine:
Two cases:
- `updater.getLinkTransforms` fails, it is useless to update the materials because we are going to set all materials to the error material (https://github.com/ros-visualization/rviz/blob/de1e8341cd0517bc1b28ebe11d3972cccd349ec1/src/rviz/robot/robot.cpp#L797)
- `updater.getLinkTransforms` suceed, then we can update the materials.

Ideally in case of success we should update the materials of only the links with an error material because there is no way that I know to dynamically change the materials during RViz execution appart from the alpha settings in the display panel but this already updates all the materials.

This should improve performance a little bit (less material updates)

### DIFF
--- a/src/rviz/robot/robot.cpp
+++ b/src/rviz/robot/robot.cpp
@@ -733,8 +733,6 @@ void Robot::update(const LinkUpdater& updater)
   {
     RobotLink* link = link_it->second;
 
-    link->setToNormalMaterial();
-
     Ogre::Vector3 visual_position, collision_position;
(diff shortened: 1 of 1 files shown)


VictorLamoine:
I'll try to find a better solution and minimize the update rate of the materials.
Useful link:
https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/robot/robot_link.cpp#L875-L897

If I can retrieve the material name I can update only the necessary

PR #1064, opened 2016-10-27
urdfdom compatibility

rhaschke:
Reapplied #1044, using urdf::SharedPtr instead of explicit boost::shared_ptr.
Fixed compatibility issues with older `urdfdom 0.3`, by relying on https://github.com/ros/robot_model/pull/160.

### DIFF
--- a/src/rviz/robot/robot.h
+++ b/src/rviz/robot/robot.h
@@ -39,6 +39,9 @@
 #include <OgreQuaternion.h>
 #include <OgreAny.h>
 
+#include <urdf_model/types.h>
+#include <urdf_world/types.h>
+
 namespace Ogre
 {
 class SceneManager;
@@ -60,13 +63,6 @@ class Axes;
 namespace tf
 {
 class TransformListener;
-}
-
-namespace urdf
-{
-class ModelInterface;
-class Link;
-class Joint;
 }
 
 namespace rviz
@@ -173,12 +169,12 @@ public:
   {
   public:
     virtual RobotLink* createLink( Robot* robot,
-                                   const boost::shared_ptr<const urdf::Link>& link,
+                                   const urdf::LinkConstSharedPtr& link,
                                    const std::string& parent_joint_name,
                                    bool visual,
                                    bool collision);
     virtual RobotJoint* createJoint( Robot* robot,
-                                     const boost::shared_ptr<const urdf::Joint>& joint);
+                                     const urdf::JointConstSharedPtr& joint);
   };
 
   /** Call this before load() to subclass the RobotLink or RobotJoint class used in the link property.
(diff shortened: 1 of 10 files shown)


wjwwood:
Wow thanks @rhaschke, I was expecting to do this today. I appreciate you opening the pr!

The code changes lgtm, I need to do a build on Wily real quick to test it.

rhaschke:
You are welcome. I was working on that anyway. Please also have a look at https://github.com/ros/urdfdom_headers/pull/33. The most annoying incompatibility was the new `urdf_world/types.h`, which doesn't make sense to me at all.

wjwwood:
Is ros/urdfdom_headers#33 necessary for these pr's to get merged? I would think not (since we probably cannot get those changes released into Debian/Ubuntu at this point).

rhaschke:
No, it's not. There is a workaround in the compatiblity header to also define the `ModelInterface` types correctly without `urdf_world/types.h`.

src/rviz/robot/robot.h
```diff
@@ -39,8 +39,7 @@
 #include <OgreQuaternion.h>
 #include <OgreAny.h>
 
-#include <urdf_model/types.h>
-#include <urdf_world/types.h>
+#include <urdf/model.h> // can be replaced later by urdf_model/types.h
```
rhaschke: This can be changed to include urdf_model/types.h as soon as https://github.com/ros/urdfdom_headers/pull/33 is merged.

wjwwood:
I tested this manually on Wily with ros/robot_model#160. Thanks again @rhaschke!

languitar:
Is it possible to also merge this into indigo-devel?

wjwwood:
@languitar maybe, why do you need it though?

languitar:
At least on archlinux I am unable to compile indigo desktop without this fix.

wjwwood:
That's probably because archlinux has a newer version of `urdfdom`/`urdfdom_headers` that we use on Ubuntu Trusty for ROS Indigo.

I'll consider backporting it next time I do a release, but honestly you should update to Kinetic if you're using the latest dependency versions on arch.

languitar:
Sure, I'd like to if everything I wanted to try worked well with that version.

wjwwood:
> Sure, I'd like to if everything I wanted to try worked well with that version.

I assume you mean, something that you're using doesn't support Kinetic?

For me there's three things you could do:

- Use Indigo on Ubuntu Trusty
  - Nobody wants to ask you to do that, but it's the least effort
- Push on the thing you're using to provide support for Kinetic
  - This is more work, but the best solution
- Ask me to backport patches to a stable release (Indigo) in rviz
  - I might be willing to help you out, but in the context of the other options, it seems the least appropriate to me

I'll see what I can do, but please put pressure on the upstream thing to support Kinetic too.

wjwwood:

Issue #1065, opened 2016-11-01
OccupancyGrid display ignores timestamps


ayrton04:
The OccupancyGrid plugin (MapDisplay) is [transforming](https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/default_plugin/map_display.cpp#L707) the map using a time stamp of `ros::Time()` instead of `current_map_.header.stamp`. If this was intentional, perhaps we can add an option to use the header's time stamp? I'd be happy to submit a PR.

wjwwood:
I'm not sure if it was intentional or not. @hersh @DLu can either of you think of a reason the current time was used for the transform in the MapDisplay rather than using the timestamp in the header? Either way, having an option (which defaults to the current behavior) would not do any harm.

NikolausDemmel:
When the `OccupancyGrid` represents a grid map, I would assume that `header.stamp` would correspond to the map-creation time, which for the typical localization scenario would lie far in the past and outside the TF buffer, so I think the current behaviour as a default makes sense for that scenario, but I understand that for different uses of OccupancyGrid this might make less sense.

DLu:
Nothing comes to mind.

wjwwood:
Closed with https://github.com/ros-visualization/rviz/pull/1066

PR #1066, opened 2016-11-02
Adding option to transform map based on header timestamp

ayrton04:
Addresses #1065. Wasn't sure what name to give the property in the display window. I'm open to suggestions if anyone has any!

### DIFF
--- a/src/rviz/default_plugin/map_display.cpp
+++ b/src/rviz/default_plugin/map_display.cpp
@@ -117,6 +117,11 @@ MapDisplay::MapDisplay()
                                            "Prefer UDP topic transport",
                                            this,
                                            SLOT( updateTopic() ));
+
+  transform_timestamp_property_ = new BoolProperty( "Use Timestamp", false,
+                                                    "Use map header timestamp when transforming",
+                                                    this,
+                                                    SLOT( transformMap() ));
 }
 
 MapDisplay::~MapDisplay()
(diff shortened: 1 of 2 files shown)


NikolausDemmel:
LGTM.

The name is reasonable IMHO.

wjwwood:
Cool, I'll try it out locally ASAP. @ayrton04 can you give an example scenario where I could test this out?

ayrton04:
I'm afraid I don't have code I can share, but if you turn a delayed laser scan into an occupancy grid with a timestamp and `frame_id` of the laser scan itself, then it becomes easy to see the effect of the PR.

ayrton04:
I've also found that just running a robot in sim and visualizing the local costmap works well. If it moves around a fair amount when the new parameter is set to false, it becomes much more stable after enabling it.

wjwwood:
I tried using the examples in `navigation_stage` but I was unable to see the change you're talking about when enabling/disabling this option. Maybe you could make a short video/gif and upload it?

ayrton04:
Sure, I'll do that.

wjwwood:
I tried this again but I still couldn't see the difference. I must be doing something that doesn't exhibit the problem in the first place. Since the code lgtm and its hidden behind an option with a default behavior that hasn't changed, I'm going to go ahead and merge this. If someone else notices this doesn't resolve the problem we can fix it up then.

ayrton04:
Apologies, @wjwwood. I am running Xenial/Kinetic, and was hoping to use one of the open-source robot sims to produce an example, but never found a Kinetic-friendly one, and didn't get around to building an Indigo install in Xenial.

PR #1073, opened 2017-01-09
Update display if empty pointcloud2 is published

axelschroth:
Currently, a PointCloud2 display keeps showing the old point cloud if an empty point cloud message is published. I consider this a bug, since rviz still shows old data which is invalid in the meantime.

### DIFF
--- a/src/rviz/default_plugin/point_cloud2_display.cpp
+++ b/src/rviz/default_plugin/point_cloud2_display.cpp
@@ -106,50 +106,51 @@ void PointCloud2Display::processMessage( const sensor_msgs::PointCloud2ConstPtr&
   }
 
   filtered->data.resize(cloud->data.size());
+  uint32_t output_count;
   if (point_count == 0)
   {
-    return;
-  }
-
-  uint8_t* output_ptr = &filtered->data.front();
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for the report and pr, I'll investigate it asap. If you have time to provide it, a simple bag file that reproduces the problem would make it easier for me to test this.

As you can imagine, this is sensitive code (a break in this code would affect a lot of users), so I'll have to be quite cautious with it. If any other point cloud users can +1/-1 this I would appreciate it too.

axelschroth:
Sure, attached you will find a bag publishing some points on /point_cloud followed by an empty point cloud.

[test_pointcloud_2017-01-10-10-20-10.bag.zip](https://github.com/ros-visualization/rviz/files/695730/test_pointcloud_2017-01-10-10-20-10.bag.zip)

wjwwood:
Sorry it took me so long to get back to you. I've reproduced what you've described and that your patch fixes it. Thanks for giving a clear description and a reproducible bag, makes my job a lot easier :smile:!

I'll still do a code review before merging, but it's looking good so far.

wjwwood:

PR #1074, opened 2017-01-12
visualize labeled points in point cloud with different colors

NamanCMU:
Added a small piece of code to visualize points with different labels in the point cloud with different colors (10 for now, can be increased).

To use it, in the PointCloud2 plugin in RVIZ, set ColorTransformer to Intensity and then change Channel Name to label. Note this requires that label is added as an entry in the PointField[] in the sensor_msgs/PointCloud2 entry. All the points with the same label are assigned the same color. This is useful to visualize point clusters representing objects, ground plane with the same color.

### DIFF
--- a/src/rviz/default_plugin/point_cloud_transformers.cpp
+++ b/src/rviz/default_plugin/point_cloud_transformers.cpp
@@ -39,6 +39,9 @@
 #include "rviz/validate_floats.h"
 
 #include "point_cloud_transformers.h"
+#include <sensor_msgs/point_cloud_conversion.h>
+
+#define maxColors 10
 
 namespace rviz
 {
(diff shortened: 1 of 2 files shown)


wjwwood:
I'll review this asap, but in the meantime I wanted to suggest some alternatives you could do without modification.

I believe that you could color by label already if it were a scalar index (an int) rather than a string by using the "rainbow coloring" feature. I'd have to try it with your specific point cloud, but it might be possible with the current features. However, it might not give as much control over the visualization as your patch.

NamanCMU:
yes, the label is an int.
Yeah, I tried using the rainbow coloring feature (although I didn't fully understand its implementation) and here are the two pictures showing point cloud using rainbow coloring feature and the new patch.
Using rainbow feature:
use_rainbow
Using the patch:
labeledpc

wjwwood:
After looking at the code and looking at your point cloud data, I think it is because the function `getRainbowColor` expects the channel's value to be within [0.0 - 1.0], see:

https://github.com/NamanCMU/rviz/blob/950af39b87e2e30eaa5f669b4dc5981d59a974f7/src/rviz/default_plugin/point_cloud_transformers.cpp#L53-L54

The other thing that I noticed is that if I select your pointcloud's points in rviz and look at the value for label, a few are normal integers, but others are large floating point numbers:

screenshot from 2017-01-27 17-05-14

So, I don't know if it is your data or the fact that rainbow coloring expects the channel (in this case "label") to be a floating point number between 0 and 1. Either way, if you have control over your data's creation I'd advise trying out some different values for label and ensure you're putting the correct information into that channel for each point.

As for this patch, I don't think looking for a channel called "label" explicitly is very flexible. I'd rather see additional options for how rviz maps values in a channel to point colors than what's proposed here.

daizr:
@wjwwood i needed something similar but w/ ability to name/alias labels + set colors so did that @ https://github.com/dispatch-ai/rviz-labeled-point-cloud. if you are interested i can submit it for merge.

wjwwood:
@daizr pr's are always welcome. We'll do our best to merge it at some point, but at the very least others can see it and perhaps iterate on it.

dhood:
I don't have access to your specific point cloud, @NamanCMU, but from what wjwwood mentioned it must be using UINT_MAX for some of its labels.

I suspect this is the reason why rainbow colouring wasn't sufficient, because that colouring scheme by default will auto-compute the range of intensity values, which it uses for the mapping of intensities onto the 0-to-1 rainbow range. Your range would have been very large, and so all labels around 0, e.g. 0-20, would have [mapped to roughly the same colour the rainbow scale](https://github.com/NamanCMU/rviz/blob/950af39b87e2e30eaa5f669b4dc5981d59a974f7/src/rviz/default_plugin/point_cloud_transformers.cpp#L175) (red). If instead you deselect "Autocompute Intensity Bounds" in rviz and set the max intensity to the number of labels-1, you will be able to utilise the full rainbow range for your labels.

However, if the range of the labels is still large, then there won't be much colour distinction between similarly-numbered labels. At that point we can introduce additional properties for the Intensity colour transformer (such as modulo like you have incorporated in this PR), but that won't guarantee you colour distinction in space either (if labels 1 and 11 are next to each other), so we are probably better off considering an approach to colouring the label data that allows user modification, e.g. https://github.com/ros-visualization/rviz/pull/1162, which has since been proposed. You are welcome to try out that PR and give feedback, I suspect it will fit your use case nicely especially if you have high range in your labels.

dhood:
I imagine that you have already resolved/worked around your issue, @NamanCMU, so I am going to close this PR. If other users encounter the same situation, I think it's possible to resolve without this PR. For data labeled integers, rainbow colouring should be sufficient, provided that "autocompute intensity bounds" is deselected.

If a use case comes up where the modulo approach to colouring is more appropriate than rainbow, this PR can be revived. Thanks again for proposal!

Issue #1077, opened 2017-01-18
Render window 1/4 size with QT5, OSX, Retina display


jonfink:
System:

- OSX 10.11
- ROS Kinetic
- QT 5.7.1 (homebrew)
- Ogre 1.9.0 (homebrew osrf/simulation tap)
- RVIZ 1.12.4

After launching Rviz, the render window seems to only draw for the lower-left 1/4 of it's total space:

<img width="1258" alt="screen shot 2017-01-18 at 11 48 29 am" src="https://cloud.githubusercontent.com/assets/131946/22073732/1406dd82-dd74-11e6-8eea-f3601dae3068.png">

I've traced this problem back to be related to [Qt5's "support" for retina displays](https://blog.qt.io/blog/2013/04/25/retina-display-support-for-mac-os-ios-and-x11/) (where the rendering units are scaled).

[Realizing that this problem was recently addressed by the Ogre development team](https://bitbucket.org/osrf/gazebo/issues/2070/rendering-1-4-of-scene-with-qt5-on-macos), I tried the same fix locally (since I am using Ogre 1.9):

In `ogre_helpers/render_system.cpp`
```c++
  params["macAPI"] = "cocoa";
  params["macAPICocoaUseNSView"] = "true";
  // Adding this param for retina display scaling
  params["contentScalingFactor"] = "2.0";
```

Obviously, this fix requires a number of checks (for presence of Qt5 + Ogre 1.9 and check for retina display) before a merge request can be considered but I wanted to report here in case others are hitting the same issue.

folkwang:
this happened to me as well on macOS Sierra. My environments are ROS kinetic + Ogre 1.9.0 + rviz version 1.12.4. FYI, Since I am also using an external low DPI display, by dragging rviz window into the external display it would be rendered correctly

wjwwood:
Can either of you try `kinetic-devel` now that https://github.com/ros-visualization/rviz/pull/1078 has been merged to see if it fixes your issue?

wjwwood:
In fact I'm going to close this optimistically, and if it is not fixed I'll reopen this.

Issue #1069, opened 2016-12-15
High DPI issue


boris-il-forte:
From ros answers:
http://answers.ros.org/question/249914/rviz-high-dpi-issue/

I'm posting also here because on ros answers I haven't had any informations about this issue.
I don't know if is an issue about my current configuration or an RVIZ issue, unfortunately I don't have another high dpi screen to test the issue on a different configuration.

wjwwood:
Can you try `kinetic-devel` now that https://github.com/ros-visualization/rviz/pull/1078 has been merged to see if it fixes your issue?

I'm going to close this and if it is not fixed, just comment here and I'll reopen it.

boris-il-forte:
This seemes to be already fixed in the version on kinetic repositories. Maybe was an ogre bug?

orzechow:
I'm having exactly the same issue with RViz version 1.12.15 (kinetic), compiled against Qt 5.5.1 and OGRE 1.9.0 (Ghadamon)

I'm running ROS on a Linux Mint 18 KDE system (based on Ubuntu 16.04)

One thing I noticed: my system has Qt 5.6.1 installed (python-pyqt5 and python3-pyqt5 5.5.1 though). Maybe this is related?

boris-il-forte:
The problem is back again in kubuntu 16.04... Maybe a regression?

PR #1078, opened 2017-01-20
Correctly scale the render panel on high resolution displays

roehling:
On Apple retina displays, the render widget only occupies a quarter of the available screen estate due to the higher DPI resolution. This PR fixes the problem for the render display and the mouse events.


roehling:
Fixes #1077 and (I hope) #1069

wjwwood:

PR #1080, opened 2017-02-12
support multiple material for one link

k-okada:
this is rewrited version of #812 for indigo


peci1:
I have rviz 1.11.19-0trusty-20180715-213359-080 and I still cannot use multiple materials per link if the first visual is a mesh (all subsequent primitive visuals are displayed in the same red collision elemets are displayed). Just reordering the mesh visual to be the last fixes that behavior.

PR #1084, opened 2017-03-06
Visualization collada texture in marker fix

mcopejans:
Fixes COLLADA mesh material colors as described in issue https://github.com/ros-visualization/rviz/issues/751. From the issue thread, it looks like this case has already been fixed by @wjwwood ([752](https://github.com/ros-visualization/rviz/pull/752)), but unfortunately it got unfixed by another PR ([893](https://github.com/ros-visualization/rviz/pull/893)).

- Tested locally with Collada file generated by Blender. Without the fix, this image would be plain white.
image
- Tested with example by @garaemon: https://github.com/garaemon/rviz_collada_marker
- The fix implemented here was originally proposed by @brindza

### DIFF
--- a/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
+++ b/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
@@ -194,8 +194,6 @@ void MeshResourceMarker::onNewMessage(const MarkerConstPtr& old_message, const M
 
    
 
-    // always update color on resource change
-    update_color = true;
 
     handler_.reset(new MarkerSelectionHandler(this, MarkerID(new_message->ns, new_message->id), context_));
     handler_->addTrackedObject(entity_);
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for the pr, especially since I haven't had the time to look into the reports of this:

https://github.com/ros-visualization/rviz/commit/e5ecd455b20aae15b31d29325c93ecb1e5894403#comments

I feel a bit like a leaf in the wind on this issue, since I don't have a good way of sorting out which solutions to merge or not, which probably led to the "unfix" of this issue. What I'm missing from this is confirmation that we're not "unfixing" something else with this pull request. I'd really like some feedback from @loganE or @brindza or even @hersh.

Otherwise, I'll have to sit down and try to run the "demos" we have to see that we're not breaking some other expected behavior. That will probably take some time for me to get to myself. If anyone else has time to look at the demos with this pr and corroborate with @mcopejans that we're making purely forward progress I'd appreciate and I could merge this directly. Specifically I'm talking about the demos in these images:

https://github.com/ros-visualization/rviz/pull/893#issuecomment-110776874

@mcopejans I don't mean to discount the testing you've done either. Perhaps some before and after images of the tests you did would help too:

> Tested with example by @garaemon: https://github.com/garaemon/rviz_collada_marker

mcopejans:
@wjwwood Thanks for your time, here are some before-after images. Let me know if there is anything I can do to test that this PR does not break something else.
Before
image
After
image

Issue #1087, opened 2017-03-19
Path color not saved in config file


theonteg:
As described in the topic of the issue, rviz does not save Path lines color to the config file.

As a result every time i load a config, path lines are the same pink default color.

Is this a bug?

wjwwood:
It sounds like a bug. Unfortunately I don't have time to verify it and/or look into a fix for it.

You might start by looking at the config's section for the path display and see if there is a field like that in there. If not, then you might need to modify the display plugin for path's to store and use it. Otherwise it might just be a bug in the path display plugin. Either way, you can see the code for it here:

- https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/default_plugin/path_display.h
- https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/default_plugin/path_display.cpp

hiddewie:
I've just today also run into this bug.

I looked into the code, and it seems related to the fact that in `PathDisplay` *two* properties are defined with the name `Color`. See https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/default_plugin/path_display.cpp#L99 and https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/default_plugin/path_display.cpp#L71. 

A solution would probably be renaming the pose color as this property should have a different name.

It would be nice to test this when adding properties to a certain display type to check if a property already exists with that name.

romainreignier:
@hiddewie Thanks for taking the time to fix this issue.
I have faced that issue for several weeks but did not take the time to track the issue and find a fix.
:+1:

PR #1089, opened 2017-03-23
Fixed duplicate property name for Path colors

hiddewie:
See issue #1087.


romainreignier:
Successfully tested!
Thanks @hiddewie !
:+1:

wjwwood:
Thanks for the fix!

Looks like the duplicate property name was introduced in this pr: https://github.com/ros-visualization/rviz/pull/1016 and then forward ported to kinetic-devel by me. Since this is a relatively new feature that caused a bug I'll take this pr now. Otherwise I'd have been cautious of changing a property name because it might break existing config files.

I'll mark this for back port to indigo/jade as well. Thanks again!

romainreignier:
Will it be part of the next Kinetic sync? It would be great.

wjwwood:

PR #1094, opened 2017-05-01
ogre 1.10: prevent loading resource under same name twice

de-vri-es:
In Ogre 1.10 it is now an error to load a resource under the same name twice, even when it is the same resource.

This seems a bit odd when resources are automatically loaded from shader programs, considering they are supposed to be able to re-use functions from other files. That's exactly what was happening with some shaders included with rviz, as also described in #1049.

This PR applies the fix described there by @phil0stine.


de-vri-es:
Apologies, I just noticed #1063. I'll close this PR.

PR #1079, opened 2017-02-12
support multiple material for one link

k-okada:
this is rewrited version of #812

we'll get 
screenshot from 2017-02-12 23 43 04
with this urdf
```
<?xml version="1.0"?>
<robot name="hoge">
  <link name="base_link">
    <visual>
      <geometry>
        <cylinder length="4" radius="1"/>
      </geometry>
      <material name="black_metal">
        <color rgba="0.1 0.1 0.1 1"/>
      </material>
    </visual>
    <visual>
      <geometry>
        <cylinder length="2" radius="5"/>
      </geometry>
      <material name="gold">
        <color rgba="1 0.549 0 1"/>
      </material>
    </visual>
  </link>
</robot>
```

PR #1098, opened 2017-05-02
fix type error in newer versions of urdf

wjwwood:
I believe this should fix the compiler errors of the latest rviz release on some platforms which were introduced in: https://github.com/ros-visualization/rviz/pull/1079


wjwwood:

PR #1102, opened 2017-05-07
Use unique material names for robot links.

de-vri-es:
Two places in `robot_link.cpp` forgot to increment the static counter for creating unique material names. Since Ogre 1.10 this is not just a possible visual glitch but also an error resulting in an `Ogre::ItemIdentityException`:

```
  what():  ItemIdentityException: Resource with the name robot link color material 1 already exists. in ResourceManager::add at /build/ogre/src/ogre/OgreMain/src/OgreResourceManager.cpp (line 150)
```

This PR increments the counter, and RViz no longer crashes when attempting to visualize a robot.

### DIFF
--- a/src/rviz/robot/robot_link.cpp
+++ b/src/rviz/robot/robot_link.cpp
@@ -208,7 +208,7 @@ RobotLink::RobotLink( Robot* robot,
   // create material for coloring links
   std::stringstream ss;
   static int count = 1;
-  ss << "robot link color material " << count;
+  ss << "robot link color material " << count++;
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for the fix, sorry about missing this in the review.

This is the second regression of https://github.com/ros-visualization/rviz/pull/1079 so far. Hopefully it is the last...

de-vri-es:
Thanks for the quick merge :)

wjwwood:

PR #1104, opened 2017-05-17
Update variable name

genemerewether:
Somehow, variable names got out of sync. Lines 370 and 371 refer to "parameters" but it is "params" everywhere else.

Thanks for the Retina fixes! Just tested on Sierra and they work great.


wjwwood:
Thanks, might have happened when I merged an older branch that also touched this code.

PR #1105, opened 2017-05-18
Fix Ogre::ItemIdentityException when loading mesh without material

kartikmohta:
When loading meshes without a material defined, rviz crashes with error:
```
  what():  ItemIdentityException: Resource with the name RVIZ/ShadedRed0Robot already exists.
```
Seems like this was the intended purpose of adding the `static int count`.

### DIFF
--- a/src/rviz/robot/robot_link.cpp
+++ b/src/rviz/robot/robot_link.cpp
@@ -620,11 +620,11 @@ void RobotLink::createEntityForGeometryElement(const urdf::LinkConstSharedPtr& l
     offset_node->setPosition(offset_position);
     offset_node->setOrientation(offset_orientation);
 
+    static int count = 0;
     if (default_material_name_.empty())
     {
       default_material_ = getMaterialForLink(link);
 
-      static int count = 0;
       std::stringstream ss;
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for the pull request, can you provide a urdf snippet (and bag file with `/tf` if needed) that reproduces this problem?

kartikmohta:
Here is the urdf + dae which cause the crash: [robot_model.zip](https://github.com/ros-visualization/rviz/files/1012106/robot_model.zip)

Just launching with `roslaunch urdf_tutorial display.launch model:=robot.urdf` causes rviz to crash.

kartikmohta:
I should mention that this is with Ogre v1.10

v4hn:
+1

I just wanted to contribute the exact same patch.
Ogre 1.10 became more strict when the user tries to add existing resources.
Looking at the code, the patch creates the behavior that was originally planned, but wrongly implemented.
Please merge and release soon.

wjwwood:
I confirmed the bug last week and this fix. I just finished reviewing the code (not many changes but I wanted to look at the context to understand the fix) and it lgtm.

PR #1106, opened 2017-05-23
Created a new Preferences window, and added PromptSaveOnExit option.

MasterEric:
For tutorials installed to a read-only location, it may be disorienting for the user to be prompted to save. For this reason, I created a preferences dialog window, with a boolean option for whether to Prompt Save on Exit, defaulting to true. These preferences are saved to any config .rviz files that are created.

### DIFF
--- a/src/rviz/default_plugin/CMakeLists.txt
+++ b/src/rviz/default_plugin/CMakeLists.txt
@@ -22,6 +22,7 @@ set(SOURCE_FILES
   map_display.cpp
   marker_array_display.cpp
   marker_display.cpp
+  marker_utils.cpp
   markers/arrow_marker.cpp
   markers/line_list_marker.cpp
   markers/line_strip_marker.cpp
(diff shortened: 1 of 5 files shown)


wjwwood:
@MasterEric sorry for the long delay in reviewing this. Looks like there are some merge conflicts now. One of us may look at resolving the conflicts soon, but if you have time that would be appreciated.

MasterEric:
@wjwwood The commit has been rebased to resolve conflicts with the base branch.

One thing to note about the preferences window is that I added it with the intent of allowing myself or other users to create additional options in the future.

MasterEric:
@wjwwood It has been 17 days since this pull request was last reviewed. There are currently no conflicts, and if you see no problem with the changes I would like them merged.

wjwwood:
@MasterEric one of us needs to compile it and try it out. We'll get to as soon as possible. Thanks for your patience.

dhood:
Hey @MasterEric, you're right that this has taken a long time to get reviewed; we will try to look at it this week/next. One thing I'm able to say is now that, unfortunately, since the PR will break ABI, it is not likely we will be able to release this into existing distros, but instead will have to hold for the Melodic release even if it gets approved. I'll get back to you with a more thorough review once I've tested it out.

PR #1108, opened 2017-05-25
Add fullscreen option (backport #1017 to indigo)

wxmerkt:
Cherry-picked the commit to indigo-devel as a follow up to #1017


wjwwood:
Thanks but I'll do the back ports myself when I do the next Indigo release. I plan to do that after Lunar has gotten some testing.

wxmerkt:
Alright, thanks!

Issue #1116, opened 2017-06-22
RVIZ open config and save config shortcuts not working


shahvi:
Hi folks,
I observed this on my installation and was wondering if anyone else has seen it too.
In RVIZ, under the file menu, keyboard shortcuts 'Ctrl+O' and 'Ctrl+S' are offered for 'Open Config' and 'Save Config' respectively.  However, I am unable to use either of these shortcuts.  Has anyone else experienced this?
ROS version: Kinetic
OS: Ubuntu 16.04

Cheers,
Vik

Kukanani:
I can confirm that this is an issue on my machine. For example, when pressing Ctrl+S, the Save action does not occur, but instead, the Select tool is activated. Ctrl+O does nothing instead of opening Open.

I have a fix in the works: #95

dhood:
@shahvi this should have been addressed by #1117 (thanks @Kukanani), which has been released now. If you don't see the issue's fixed after updating rviz to >=1.12.11 please re-open.

dhood:
correction: if it's not fixed please comment here and we'll re-open on your behalf

PR #1117, opened 2017-06-28
Add keyboard shortcuts actions to VisualizationFrame directly

Kukanani:
This fixes a known issue (#1116) where keyboard shortcuts to trigger menu actions do not work. 
The action has to be added to the parent frame as well as the menu itself. The Fullscreen action (F11) was already doing that correctly.

I also added a (pretty standard) keyboard shortcut for Save As.

### DIFF
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -447,9 +447,14 @@ void VisualizationFrame::savePersistentSettings()
 void VisualizationFrame::initMenus()
 {
   file_menu_ = menuBar()->addMenu( "&File" );
-  file_menu_->addAction( "&Open Config", this, SLOT( onOpen() ), QKeySequence( "Ctrl+O" ));
-  file_menu_->addAction( "&Save Config", this, SLOT( onSave() ), QKeySequence( "Ctrl+S" ));
-  file_menu_->addAction( "Save Config &As", this, SLOT( onSaveAs() ));
+
+  QAction * file_menu_open_action = file_menu_->addAction( "&Open Config", this, SLOT( onOpen() ), QKeySequence( "Ctrl+O" ));
+  this->addAction(file_menu_open_action);
+  QAction * file_menu_save_action = file_menu_->addAction( "&Save Config", this, SLOT( onSave() ), QKeySequence( "Ctrl+S" ));
+  this->addAction(file_menu_save_action);
+  QAction * file_menu_save_as_action = file_menu_->addAction( "Save Config &As", this, SLOT( onSaveAs() ), QKeySequence( "Ctrl+Shift+S"));
+  this->addAction(file_menu_save_as_action);
+
   recent_configs_menu_ = file_menu_->addMenu( "&Recent Configs" );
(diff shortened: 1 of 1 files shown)


src/rviz/visualization_frame.cpp
```diff
@@ -447,9 +447,14 @@ void VisualizationFrame::savePersistentSettings()
...
-  file_menu_->addAction( "&Open Config", this, SLOT( onOpen() ), QKeySequence( "Ctrl+O" ));
-  file_menu_->addAction( "&Save Config", this, SLOT( onSave() ), QKeySequence( "Ctrl+S" ));
-  file_menu_->addAction( "Save Config &As", this, SLOT( onSaveAs() ));
+
+  QAction * file_menu_open_action = file_menu_->addAction( "&Open Config", this, SLOT( onOpen() ), QKeySequence( "Ctrl+O" ));
+  this->addAction(file_menu_open_action);
+  QAction * file_menu_save_action = file_menu_->addAction( "&Save Config", this, SLOT( onSave() ), QKeySequence( "Ctrl+S" ));
+  this->addAction(file_menu_save_action);
+  QAction * file_menu_save_as_action = file_menu_->addAction( "Save Config &As", this, SLOT( onSaveAs() ), QKeySequence( "Ctrl+Shift+S"));
+  this->addAction(file_menu_save_as_action);
```
wjwwood: That's interesting that this fixes it for you. Can you elaborate on why this is necessary? The Qt examples don't do that, but then again it could be related to the relationship between the UI elements in this particular scenario.

http://doc.qt.io/qt-5/qtwidgets-mainwindows-menus-example.html (I was looking at the section `void MainWindow::createMenus()`)

Maybe the `createMenus()` function is being overloaded and all we need to do is to change the name of this function.

I'm not certain about any of those ideas, I'm just throwing things out there incase you don't have a good explanation for why this fixes the problem and is the "correct" way to do it.

wjwwood:
Also, maybe we should switch to using the standard key sequences provided by Qt:

http://doc.qt.io/qt-5/qkeysequence.html#StandardKey-enum

E.g. `QKeySequence::SaveAs`.

Kukanani:
My guess is that the RenderPanel has focus by default, and as long is it does, all keyboard events are routed there (which is required for tool selection etc.) instead of to the menu object. By adding the event to the parent window, I assume this allows the event to be captured even when the RenderPanel has focus.

Without this fix, if you switch focus to the menu (for example, enter and exit fullscreen with F11), then try Ctrl+S, it works. It's pretty odd behavior, but this hand-wavey explanation is the best hypothesis I have.

wjwwood:
Even with your explanation. It isn't clear to me why it's necessary. You'd think that the Qt example I linked would work even if the menu doesn't have focus (though I'm not entirely clear on that mechanism either). Perhaps there is an issue with the passing of keypress events up from the render window which causes the visualization frame itself to be handled, but not it's children (like the menu). Since this is an incremental improvement, I'll consider it for merge, but I'll keep an eye out for a better solution.

PR #1122, opened 2017-07-10
Avoid unnecessary addPoints in PointCloudCommon::update

ThomasMoinel:
After adding new points to the PointCloud, setRenderMode call regenerateAll which remove all points and add then again (to apply the new render mode). 
As the PointCloud is just created, it is empty. So calling setRenderMode before addPoints  avoid the regeneration and improve performance.

This is an easy performance improvement, there a lot a copying in this plugin that hurt perf, with patterns like allocating a lot of memory to throw it right away from each new incoming pointcloud (or depth_cloud). I might find some time to dig more deeply into it.

### DIFF
--- a/src/rviz/default_plugin/point_cloud_common.cpp
+++ b/src/rviz/default_plugin/point_cloud_common.cpp
@@ -590,8 +590,8 @@ void PointCloudCommon::update(float wall_dt, float ros_dt)
         bool per_point_alpha = findChannelIndex(cloud_info->message_, "rgba") != -1;
 
         cloud_info->cloud_.reset( new PointCloud() );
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks for the pr! The change looks straightforward, but I'm going to delay it until the next release because I want to release it when we have time for some soak time in the testing repository, since this change could have subtle side effects on point cloud rendering (it is a potentially high impact change).

wjwwood:
Thanks again for the patch! I've tested it out myself, and I didn't notice any difference off-hand. I'll merge this and ask for some extra testing from people while it is in the testing repository.

Issue #1120, opened 2017-07-06
Visualization markers ignore color and show everything as white


gaya-:
After the last update on Indigo visualization markers started to appear white.
The message is the same as before and has correct color data.
I suspect it was one of the latest commits about collada file visualization.
Do I understand it correctly that the color information from the Collada file has now higher precedence than the visualization marker message color? If so, what is the reasoning behind that?

wjwwood:
@gaya- sorry for the regression. That code has been tricky, I'll have a look and see if I can just revert those changes for now and do another release.

wjwwood:
@gaya- I'm guessing it is this pr: https://github.com/ros-visualization/rviz/pull/1084

Can you provide a script or bag file (with mesh file) which causes this problem for you?

wjwwood:
Cross-reference to the answers.ros.org: http://answers.ros.org/question/265425/ros-indigo-rviz-maker-turned-white-after-update/

gaya-:
Unfortunately the cad models that created this problem are proprietary so I cannot share them :/
And we have a big deadline in 2 weeks, so no time to look deeper into this.
If no one else comes up with a test case I can look into this after the deadline and see what makes our meshes so special and maybe create a similar mesh to test...

wjwwood:
I just spent some time trying to reproduce this with Indigo, but I cannot. The behavior in https://github.com/ros-visualization/rviz/pull/1084 seems correct to me and seems to still be the case with the latest version of rviz for Indigo. (btw @gaya- exactly which version of rviz are you using?)

I'm going to wait until either @gaya- or someone else can provide me with a bag file or simple Python script to reproduce it.

HenrySleek:
Hey! I have the very same problem.
I am simply using this code:
```
marker.header.frame_id = "base_footprint";
    marker.header.stamp = ros::Time::now();
    marker.ns = "basic_shapes";
    marker.id = 0;
    marker.type = visualization_msgs::Marker::MESH_RESOURCE;
    marker.action = visualization_msgs::Marker::ADD;
    marker.mesh_resource = "package://autonomous_grasping/mesh/Gripper_Assy.stl";
    marker.mesh_use_embedded_materials = false;
    marker.color.r = 0.4;
    marker.color.g = 0.4;
    marker.color.b = 0.4;
    marker.color.a = 1.0;
    marker.scale.x = 0.0012;
    marker.scale.y = 0.0012;
    marker.scale.z = 0.0012;
```
[...]

PR #1129, opened 2017-07-31
Visualization collada texture in marker fix (#1084)

wjwwood:
This is a forward port of #1084 for Kinetic. I'm pretty sure I already forward ported it, but perhaps it my other forward/back porting I accidentally merged over this change in Kinetic. It appears to be intact on `indigo-devel`.

See also: https://github.com/ros-visualization/rviz/issues/1120


dhood:
think the needs-forward-port label was applied here accidentally. It's been so long that I doubt anyone will remember (I forgot to do forward ports last kinetic release), I am just being explicit that I am removing the label without any other action.

PR #1130, opened 2017-08-01
find tinyxml and link against it

mikaelarguedas:
Since https://github.com/ros/pluginlib/pull/59 pluginlib doesnt find_package tinyXML anymore. So downstream packages using tinyxml have to find it themselves.

This builds and passes the tests locally but I didn't run prerelease jobs or run nany manual test for it.

Note @wjwwood & @dhood this will require a new lunar release

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,8 @@ else()
   set(EIGEN3_INCLUDE_DIRS ${EIGEN3_INCLUDE_DIR})
 endif()
 
+find_package(TinyXML REQUIRED)
+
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
I just noticed that there was a TiXmlElement in `robot_model_display.cpp`, given that it's not part of the API I assumed that including it in the cpp file. feel free to relocate if it's better practice

wjwwood:
That's fine, it's not used in the header, so the cpp file is ok.

wjwwood:
Thanks!

PR #1131, opened 2017-08-02
Add dhood as maintainer

dhood:


wjwwood:
@dgossow fyi, we're going to remove you as a maintainer unless you would like to remain listed as one. You're already listed as an author.

PR #1132, opened 2017-08-03
Fix for 1120

### LINKED ISSUE #1120: Visualization markers ignore color and show everything as white

wjwwood:
Fixes #1120.

Basically #1084 caused a regression that caused markers which had a mesh text and color to render white until the color or alpha was updated, after which point it would render correctly.

I also improved the marker mesh test executable to test a few of these things. Here's a screen shot of the behavior with this pr (the correct behavior imo):

screen shot 2017-08-03 at 4 47 51 pm


wjwwood:
I should note that the first row of items in the above matrix of test markers only has something in the first column. The other columns are empty, though I might update the tool again to fill the spots out, just for completeness.

wjwwood:
41bf5e2 adds the missing mesh markers to the matrix, but it doesn't affect the screen shot as they are invisible due to alpha of 0.

wjwwood:

Issue #1137, opened 2017-08-09
RViz dies when receiving an incomplete Odometry message


sharst:
Steps to reproduce: 
* Add odometry display
* Point to a topic
* Publish an odometry message with a (0x, 0y, 0z, 0w)-quaternion on this topic.

Rviz dies with the following traceback:

```
[ INFO] [1502309518.274159660]: rviz version 1.12.10
[ INFO] [1502309518.274207080]: compiled against Qt version 5.5.1
[ INFO] [1502309518.274219210]: compiled against OGRE version 1.9.0 (Ghadamon)
[ INFO] [1502309518.411463334]: Stereo is NOT SUPPORTED
[ INFO] [1502309518.411575145]: OpenGl version: 3 (GLSL 1.3).
rviz: /build/ogre-1.9-mqY1wq/ogre-1.9-1.9.0+dfsg1/OgreMain/include/OgreAxisAlignedBox.h:252: void Ogre::AxisAlignedBox::setExtents(const Ogre::Vector3&, const Ogre::Vector3&): Assertion `(min.x <= max.x && min.y <= max.y && min.z <= max.z) && "The minimum corner of the box must be less than or equal to maximum corner"' failed.
Aborted (core dumped)
```

It would be nicer if this error were caught and a warning pointing to the origin of this problem be presented to the user.

wjwwood:
That would be great. I don't have time to do that right now, but I'm always happy to take a pull request.

skohlbr:
Pretty sure this sensitivity to non-normalized (in most cases probably all-zero) quaternions appears  also in other Displays used in rviz. I know I saw a OgreAxisAlignedBox segfault when we rviz received a "bad" quaternion in some other message (cannot remember which one unfortunately).

dhood:
thanks for pointing that out @skohlbr. @sharst has fixed this particular crash in https://github.com/ros-visualization/rviz/pull/1139 so I'll close this issue; I've opened https://github.com/ros-visualization/rviz/issues/1142 to track that there are other parts in the codebase that could use this improvement (contributions welcome!)

PR #1139, opened 2017-08-10
Check for odometry quaternion normalization before display

sharst:
Fixes issue  #1137

### DIFF
--- a/src/rviz/default_plugin/odometry_display.cpp
+++ b/src/rviz/default_plugin/odometry_display.cpp
@@ -253,6 +253,15 @@ bool validateFloats(const nav_msgs::Odometry& msg)
   return valid;
 }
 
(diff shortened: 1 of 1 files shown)


dhood:
Ideally we would factor out this logic so that it can be reused in other places (as with [validate_floats.h](https://github.com/ros-visualization/rviz/blob/7970ba08cee3810cfa1609c3b0f5136970eb2f7c/src/rviz/validate_floats.h)). @sharst are you interested in doing that? otherwise this is still an incremental improvement.

wjwwood:
@dhood should we merge this as an incremental improvement and ticket the refactor? I'm just curious because I am looking at doing a release soon.

dhood:
Yep. Follow-up improvements ticketed in https://github.com/ros-visualization/rviz/issues/1142

Thanks again for the fix @sharst

PR #1143, opened 2017-08-21
Indigo backports

wjwwood:
Backporting:

- https://github.com/ros-visualization/rviz/pull/1139
- https://github.com/ros-visualization/rviz/pull/1141


wjwwood:
Actually https://github.com/ros-visualization/rviz/pull/1139 doesn't make sense (it didn't apply cleanly, but I tried to fix it up). I will drop it from the backports.

dhood:
Oops, I didn't check if there'd been refactoring before putting the 'needs back port' label on 1139. I'll just put that label on trivial back ports in the future, yeah?

wjwwood:
@dhood if you think the fix is worth the more complex back porting you're welcome to try and do it. I was just grabbing the trivial ones. It might be appropriate to put the label on non-trivial backports if they are important.

PR #1144, opened 2017-08-23
Add vertex color support to mesh_loader

emersondispatch:


wjwwood:
Thanks for the patch. We'll have a look asap.

dhood:
@emersondispatch are you able to provide an example model for us to test with?

src/rviz/mesh_loader.cpp
```diff
@@ -447,6 +465,7 @@ void loadMaterials(const std::string& resource_path,
 
     Ogre::Technique* tech = mat->getTechnique(0);
     Ogre::Pass* pass = tech->getPass(0);
+    pass->setVertexColourTracking(Ogre::TVC_DIFFUSE);
```
dhood: this is impacting other meshes, e.g. `rosrun rviz mesh_marker_test` and you'll see some of the meshes flicker because of this change. I don't have a suggestion yet for a fix

dhood:
I've put the "changes requested" label on this because of https://github.com/ros-visualization/rviz/pull/1144#discussion_r181537693 which we don't have time to investigate ourselves. I've put "help wanted" in case @emersondispatch or someone else has time to look at it.

PR #1145, opened 2017-08-27
MONO8 transformer for point cloud plugin

wkentaro:
## Why?

I need to overlay color point cloud on mono one.
For example, grasp point on sensor input.

rviz_point_cloud_mono8

### DIFF
--- a/src/rviz/default_plugin/point_cloud_transformers.cpp
+++ b/src/rviz/default_plugin/point_cloud_transformers.cpp
@@ -385,6 +385,57 @@ bool RGB8PCTransformer::transform(const sensor_msgs::PointCloud2ConstPtr& cloud,
   return true;
 }
 
+bool MONO8PCTransformer::transform(const sensor_msgs::PointCloud2ConstPtr& cloud, uint32_t mask, const Ogre::Matrix4& transform, V_PointCloudPoint& points_out)
+{
+  if (!(mask & Support_Color))
+  {
+    return false;
+  }
+
+  const int32_t rgb  = findChannelIndex(cloud, "rgb");
+  const int32_t rgba = findChannelIndex(cloud, "rgba");
+  const int32_t index = std::max(rgb, rgba);
+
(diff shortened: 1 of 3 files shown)


wjwwood:
Thanks for the pr, we'll have a look asap.

src/rviz/default_plugin/point_cloud_transformers.cpp
```diff
@@ -385,6 +385,57 @@ bool RGB8PCTransformer::transform(const sensor_msgs::PointCloud2ConstPtr& cloud,
...
+  {
+    for (V_PointCloudPoint::iterator iter = points_out.begin(); iter != points_out.end(); ++iter, rgb_ptr += point_step)
+    {
+      uint32_t rgb = *reinterpret_cast<const uint32_t*>(rgb_ptr);
+      float r = rgb_lut[(rgb >> 16) & 0xff];
+      float g = rgb_lut[(rgb >> 8) & 0xff];
+      float b = rgb_lut[rgb & 0xff];
+      float mono = 0.2989 * r + 0.5870 * g + 0.1140 * b;
+      iter->color.r = iter->color.g = iter->color.b = mono;
+      iter->color.a = 1.0f;
```
dhood: AFAICT the defaults for green and blue channels [will be 1.0](https://github.com/ros-visualization/rviz/blob/7da51fc417e93025508f95be2dd15eca59476651/src/rviz/default_plugin/point_cloud_common.cpp#L831), so I would think we need to set them to `mono` here as well to get the greyscale result that you show in the screenshot from the initial post. However, your image seems to show the pointcloud displaying as expected only setting the red. I want to confirm that only setting the red is sufficient before approving this PR (@wkentaro I know it's been a while since this PR was opened but if you had a specific reason for only setting the red channel I am interested).

dhood: Ah, I misread the line and it _is_ setting all three.

PR #1153, opened 2017-09-27
Fixes "Parse error at BOOST_JOIN error" under Xenial.

BillWSY:
When compiling under Ubuntu Xenial I countered "Parse error at 'BOOST_JOIN'" error. This fix should not break building on Trusty, but also allows rviz to build under Xenial.

Related:
https://github.com/ros-planning/moveit_ros/issues/653
https://github.com/ros-planning/moveit_ros/pull/666

### DIFF
--- a/src/rviz/robot/robot.h
+++ b/src/rviz/robot/robot.h
@@ -39,7 +39,9 @@
 #include <OgreQuaternion.h>
 #include <OgreAny.h>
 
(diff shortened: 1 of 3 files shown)


dhood:
Thanks for the patch and links; this is a duplicate of https://github.com/ros-visualization/rviz/pull/1149 so I will close it in favour of that PR

PR #1158, opened 2017-10-05
Taking roi in cameraInfo in consideration when using camera_display plugin

Shin-nn:
This should fix camera_display when using roi.

### DIFF
--- a/src/rviz/default_plugin/camera_display.cpp
+++ b/src/rviz/default_plugin/camera_display.cpp
@@ -490,12 +490,23 @@ bool CameraDisplay::updateCamera()
 #endif
 
   //adjust the image rectangles to fit the zoom & aspect ratio
-  //corners are computed according to roi
+  double x_corner_start, y_corner_start, x_corner_end, y_corner_end;
 
-  double y_corner_start = (-2.0*info->roi.y_offset/info->height+1.0)*zoom_y;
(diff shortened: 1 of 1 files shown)


dhood:
This looks good, thanks for the patch.

Thinking about behaviour stability: this _will_ change the behaviour for users with cameras with ROI set, but given that those users currently have the subimage stretched to the full camera width, I think this will be a welcome improvement for those users.

There's a chance this will negatively impact the following user: ROI set in the camera info, but image being published at full size. This previously would have rendered the image appropriately but now will distort the image according to the ROI. However, the camera info ought to match the camera image, so this user _shouldn't_ exist, so I do not consider this a reason to prevent this change.

@wjwwood if you have a chance to confirm that this behavioural change is appropriate to push into kinetic that would be appreciated.

_camera display with roi (selected in red) without this patch (top right, stretched)_
camera_roi_without_1158

_camera display with roi (selected in red) with this patch (top right, not stretched, appropriate size)_
camera_roi_with_1158

The images are from testing with turtlebot in gazebo and interactive ROI selection; launch/config file: [camera_roi.zip](https://github.com/ros-visualization/rviz/files/1569244/camera_roi.zip)

dhood:
@wjwwood I've requested your review, not to hurry it up, just so it's less likely to get lost over the holidays

dhood:
@wjwwood I'm planning to push the behaviour fix into kinetic FYI

PR #1161, opened 2017-10-10
allow inheriting classes access to more state

daizr:
needed when adding roi selection overlay to image display here https://github.com/dispatch-ai/rviz-image-roi


dhood:
replaced by https://github.com/ros-visualization/rviz/pull/1221/files exposing less state ( @daizr appears to have deleted their fork)

PR #1162, opened 2017-10-10
add label point cloud transformer

daizr:
from https://github.com/dispatch-ai/rviz-labeled-point-cloud which was based on https://github.com/ros-visualization/rviz/pull/1074


dhood:
I haven't reviewed the code in great detail yet, but have tested it out as a user. One of the issues I've encountered is that if I want to swap label colours for, say, 3 and 5, then as soon I modify 5's LabelProperty to refer to label ID 3, if a point cloud comes in with 5 in it, it will add another LabelProperty for label ID 5 at the end of the list (since it ensures that one exists for 5 at all times). I continue with my swap of 3 and 5 and am then left wondering which of the two LabelProperty's for ID 5 actually applies. This duplication even happens if I go to edit the label ID for 5's LabelProperty and then back out: I end up with two LabelProperty's for 5.

What do you think about having the label ID as immutable? This will solve the above problem, but will be inconvenient for users that have set up their LabelProperty's by alias and want to swap label IDs, e.g. because for one reason or another the point cloud labels get shuffled. In that case the user could still edit the rviz config file, swapping the label IDs for LabelProperty's.

PR #1165, opened 2017-10-20
Avoid check macro collision.

mikepurvis:
Fixes the issue described here: https://github.com/mikepurvis/ros-install-osx/pull/84#issuecomment-338221161


CMakeLists.txt
```diff
@@ -89,6 +89,7 @@ endif(NOT DEFINED OGRE_OV_LIBRARIES_ABS)
 if(APPLE)
   FIND_LIBRARY(Cocoa_LIBRARIES Cocoa)
   set(rviz_ADDITIONAL_LIBRARIES ${Cocoa_LIBRARIES})
+  add_definitions(-D__ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES=0)
```
wjwwood: Can you add a comment here describing why this is needed? If I read this a few years from now, I might just remove it because I have not idea what's its purpose is (well I wouldn't but someone might).

mikepurvis:
Commented as suggested.

PR #1168, opened 2017-11-02
Disable dock widget text eliding

### LINKED ISSUE #1148: Allow to use scroll buttons in QTabWidgets created by RViz
If you stack a lot of widgets inside RViz, the labels becomes hard to read like this:
capture du 2017-09-11 11-17-43

Solutions:
- Use shorter tab names (+ a tooltip)
- Expand the tab so that labels becomes easy to read again:
capture du 2017-09-11 11-17-54
- Allow to use [scroll buttons](http://doc.qt.io/qt-5/qtabwidget.html#usesScrollButtons-prop) in the QTabWidget:
  - fusion-tabbar (without scroll buttons)
  - fusion-tabbar-truncated (with scroll buttons)

Last solution is by far the best in my opinion.
I think one solution is to subclass the tab bar and add a right click action to choose between scroll buttons, no scroll buttons.

VictorLamoine:
Fixes #1148 

Diff without trailing whitespace changes:
https://github.com/ros-visualization/rviz/compare/kinetic-devel...VictorLamoine:QTabWidget_elide?w=1

The new slot `onDockPanelChange` checks for every QTabBar there is in the main window and makes sure text is not elided inside it.

This slot is triggered
- When a config file is loaded
- When a panel is created
- When a panel is moved

# Before
default_behaviour

# After
no_elide

### DIFF
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -48,6 +48,7 @@
 #include <QLabel>
 #include <QToolButton>
 #include <QHBoxLayout>
+#include <QTabBar>
 
 #include <boost/algorithm/string/split.hpp>
 #include <boost/algorithm/string/trim.hpp>
@@ -409,7 +410,7 @@ void VisualizationFrame::loadPersistentSettings()
       last_config_dir_ = last_config_dir.toStdString();
       last_image_dir_ = last_image_dir.toStdString();
     }
-    
+
     Config recent_configs_list = config.mapGetChild( "Recent Configs" );
     recent_configs_.clear();
     int num_recent = recent_configs_list.listLength();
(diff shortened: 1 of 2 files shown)


VictorLamoine:
Rebased against current `kinetic-devel`.

dhood:
for future reference I tested with this config:
[test_tab_eliding.rviz.txt](https://github.com/ros-visualization/rviz/files/1885932/test_tab_eliding.rviz.txt)

theobslhc:
I think this PR introduces a Qt5 dependency, but branch **kinetic-devel** is supposed to be Qt4-compatible.

`src/rviz/visualization_frame.cpp`, line 1208
```cpp
QList<QTabBar *> tab_bars = findChildren<QTabBar *>(QString(), Qt::FindDirectChildrenOnly);
```
The `Qt::FindDirectChildrenOnly` flag was introduced in Qt5.

Everything seems to build fine with Qt4 and without the commits.
Is there any solution that would not rely on `Qt:FindDirectChildrenOnly` flag? (I might try to look into that out myself soon - will update)

rhaschke:
@theobslhc, thanks for reporting this issue. Please file a new issue for this, referring to this PR.
@VictorLamoine Do you have an idea to handle this issue?

theobslhc:
See issue [#1369](https://github.com/ros-visualization/rviz/issues/1369).

PR #1174, opened 2017-12-04
add option to disable keyboard focus being set on mouse move

simonschmeisser:
currently all mouse events get forwarded to onRenderWindowMouseEvents,
there keyboard focus is requested (setFocus) unconditionally. This commit
adds a new property "focus_on_mouse_move_" that enables setting keyboard
focus on mouse move events. Setting focus on click or wheel is handled by
QObject::setFocusPolicy

Subclasses/Users of rvis::RenderPanel can disable either setting focus on
mouse move by using the boolean property and disable keyboard focus on click,
on wheel or totally via setFocusPolicy

### DIFF
--- a/src/rviz/render_panel.cpp
+++ b/src/rviz/render_panel.cpp
@@ -49,6 +49,7 @@ RenderPanel::RenderPanel( QWidget* parent )
   : QtOgreRenderWindow( parent )
   , mouse_x_( 0 )
   , mouse_y_( 0 )
+  , focus_on_mouse_move_( true )
   , context_( 0 )
   , scene_manager_( 0 )
   , view_controller_( 0 )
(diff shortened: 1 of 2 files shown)


simonschmeisser:
ping?

dhood:
thanks for the PR, we'll take a look at it as soon as possible

wjwwood:
@simonschmeisser if you want/need this in `kinetic-devel` then we'll have to find a way to do it ABI compatible. Feel free to start a discussion about that if you want after I merge this. Thanks for pr!

simonschmeisser:
@wjwwood thanks for merging it to melodic. I'm fine that way. We will upgrade asap anyway and have a workaround for the time being

PR #1182, opened 2018-01-03
Warn on unnormalised quaternions instead of rejecting them

dhood:
Since https://github.com/ros-visualization/rviz/pull/1180 fixes the only confirmed issue with invalid quaternions, don't reject them anymore.

Null quaternions will be set to identity by `FrameManager::transform` where appropriate. We do not output a warning for these because they are common in uninitialised ROS messages and are more likely indicative of users just being lazy than an error in the publisher.

Non-null, unnormalised quaternions will generate a warning, as it may be from an error in the code of the publisher (see https://github.com/ros-visualization/rviz/pull/1179#issuecomment-354910545). It would be most convenient for this warning to be displayed as a status in the GUI so it can be coupled with the specific offender. However, #1167 has revealed that there are many such cases (particularly markers), and having so many warnings throughout the display may cause more serious warnings to go unnoticed.

Instead, a console warning is generated, with `ONCE` since there's such a high prevalence of unnormalised quaternions around these days (it would be more appropriate to log a warning once for each invalid quaternion but that'd be less straightforward). Users can access info about all offending messages by setting the `quaternions` sublogger to debug. If it's more appropriate to potentially flood with console warnings then we can remove the ONCE filter; this seemed like a "gentler" approach.

### DIFF
--- a/src/rviz/validate_quaternions.h
+++ b/src/rviz/validate_quaternions.h
@@ -32,6 +32,7 @@
 
 #include <geometry_msgs/PoseStamped.h>
 #include <OgreQuaternion.h>
+#include <ros/ros.h>
 #include <tf/LinearMath/Quaternion.h>
 
 #include <boost/array.hpp>
@@ -46,7 +47,11 @@ inline bool validateQuaternions( float w, float x, float y, float z )
     // Allow null quaternions to pass because they are common in uninitialized ROS messages.
     return true;
   }
-  return std::abs( w * w + x * x + y * y + z * z - 1.0f ) < 10e-3f;
+  float norm2 = w * w + x * x + y * y + z * z; 
+  bool is_normalized = std::abs( norm2 - 1.0f ) < 10e-3f;
+  ROS_DEBUG_COND_NAMED( !is_normalized, "quaternions", "Quaternion [x: %.3f, y: %.3f, z: %.3f, w: %.3f] not normalized. "
+                        "Magnitude: %.3f", x, y, z, w, std::sqrt(norm2) );
+  return is_normalized;
 }
 
 inline bool validateQuaternions( double w, double x, double y, double z )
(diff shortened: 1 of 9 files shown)


src/rviz/validate_quaternions.h
```diff
@@ -41,12 +42,30 @@ namespace rviz
...
 {
-  return std::abs( w * w + x * x + y * y + z * z - 1.0 ) < 10e-3;
+  if ( 0.0 == x && 0.0 == y && 0.0 == z && 0.0 == w )
+  {
+    // Allow null quaternions to pass because they are common in uninitialized ROS messages.
+    return true;
+  }
+  double norm2 = w * w + x * x + y * y + z * z; 
+  bool is_normalized = std::abs( norm2 - 1.0 ) < 10e-3;
+  ROS_DEBUG_COND_NAMED( !is_normalized, "quaternions", "Quaternion [x: %.3f, y: %.3f, z: %.3f, w: %.3f] not normalized. "
```
dhood: the reason for logging here as opposed to the callers is because it can be a nested quaternion that causes a message to be invalid, so it's not so straightforward to have the caller print the norm. Still, it seems a bit awkward to have logging here so we could just remove it altogether without too much impact (trade off of less info)

rhaschke:
> Since #1180 fixes the only confirmed issue with invalid quaternions, don't reject them anymore.

The covariance display is not the only one, directly using quaternions from ROS msg as an Ogre orientation. As you can see in https://github.com/ros-visualization/rviz/pull/1179/commits/3dc2e74dec48123c716e5262c79a6999bf11ab7c, there are several other locations that might potentially cause issues:
* [interactive markers](https://github.com/ros-visualization/rviz/pull/1179/commits/3dc2e74dec48123c716e5262c79a6999bf11ab7c#diff-fad4b11cd27fd95ba15ff410fff31ee7L122)
* [map display](https://github.com/ros-visualization/rviz/pull/1179/commits/3dc2e74dec48123c716e5262c79a6999bf11ab7c#diff-218addc62ac0b6e16ffb70cffef3bfe8L688)
* [pose array display](https://github.com/ros-visualization/rviz/pull/1179/commits/3dc2e74dec48123c716e5262c79a6999bf11ab7c#diff-7ae86917d310d3052a4e52de6d176ec7L67)

@dhood All these are nicely handled by #1179 and I don't really understand why you don't consider #1179 as a basis but instead file new similar PRs (#1180, #1182).

StefanFabian:
This is what the original PR #1167 should have been.
However, one minor thing regarding markers, the actions DELETE and DELETE_ALL will give a warning if their quaternions aren't initialized which is probably not intended.

dhood:
> However, one minor thing regarding markers, the actions DELETE and DELETE_ALL will give a warning if their quaternions aren't initialized which is probably not intended.

We decided that no uninitialised quaternions will give a warning (https://github.com/ros-visualization/rviz/pull/1182/files#diff-0eefc6614225475a9ff350807d529a19R61), so this will only be the case if they have an unnormalised quaternion in the DELETE msg

PR #1189, opened 2018-01-27
check length of display-config parameter before extracting substring

dseifert:
When running `rviz -d bla`, rviz would segfault as it tried to extract a substring of the display-config parameter without checking its size first.

### DIFF
--- a/src/rviz/visualizer_app.cpp
+++ b/src/rviz/visualizer_app.cpp
@@ -172,7 +172,7 @@ bool VisualizerApp::init( int argc, char** argv )
       if (vm.count("display-config"))
       {
         display_config = vm["display-config"].as<std::string>();
-        if( display_config.substr( display_config.size() - 4, 4 ) == ".vcg" )
+        if (display_config.size() >= 4 && display_config.substr( display_config.size() - 4, 4 ) == ".vcg")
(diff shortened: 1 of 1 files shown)


src/rviz/visualizer_app.cpp
```diff
@@ -172,7 +172,7 @@ bool VisualizerApp::init( int argc, char** argv )
       if (vm.count("display-config"))
       {
         display_config = vm["display-config"].as<std::string>();
-        if( display_config.substr( display_config.size() - 4, 4 ) == ".vcg" )
+        if (display_config.size() >= 4 and display_config.substr( display_config.size() - 4, 4 ) == ".vcg" )
```
wjwwood: Please use `&&`. Also, I know the rviz style is weird, but please preserve the whitespace when possible, should be `if ( dis...` if you like.

dseifert: Shall I switch to `if ( dis...` even if (almost) the entire rest of the file does not use this style or shall I reformat the file this way?

wjwwood: Nah, that's ok, it's pretty inconsistent in this version anyways.

PR #1194, opened 2018-02-09
feat: Add ClickedPoint Panel

alireza-hosseini:
A simple panel has been added in order to decrease the complexity of using **rviz/PublishPoint** tool so that the coordinates of the clicked point will be shown in the panel and there will be no need to subscribe to `/clicked_point` topic anymore.

screenshot from 2018-02-09 13-34-52

### DIFF
--- a/icons/classes/ClickedPoint.svg
+++ b/icons/classes/ClickedPoint.svg
@@ -1,0 +1,349 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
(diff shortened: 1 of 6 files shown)


alireza-hosseini:
@dhood Can you please update me about the state of this PR? Is it going to be merged soon? 
Thanks.

dhood:
Hi @alireza-hosseini, thanks for the contribution! And thanks for your patience.

I can appreciate that you are interested in seeing the last clicked pointed visualised in the display. An issue with the implementation though is that it's hard-coded the topic to `/clicked_point`, while the Publish Point tool can be configured to publish on different topics via the Tool Properties panel. That means this panel will stop working if someone changes the tool's topic.

Would the following be sufficient for your use case?
- Update [the default status message of the tool](https://github.com/ros-visualization/rviz/blob/7970ba08cee3810cfa1609c3b0f5136970eb2f7c/src/rviz/default_plugin/tools/point_tool.cpp#L127) (shown at the bottom of rviz) to say the last clicked point.
- Disable the "single click" option on the publish point tool so the status is still displayed after clicking.

If so, that would be the simplest change to be able to see the last clicked point.

rhaschke:
I'm not quite sure about the utility of this PR. If I understand you, @alireza-hosseini, correctly, you want to display the published point. However, having the PublishPoint tool activated, the `to-be-published` point is shown in the status bar. Doesn't that suffice?
If not, and you want to see the actually published point for a while, a simpler way to go (rather than adding another panel) was suggested in https://github.com/ros-visualization/rviz/pull/1194#issuecomment-379415779. However, you didn't react on this.
For this reason, I will close this PR for now. If you are still interested in the proposed solution, please feel free to open a corresponding issue or - even better - file a pull request.

Issue #1081, opened 2017-02-15
Wrench visual deadzone


goretkin:
I strongly dislike the behavior I'm getting of hiding the arrows if they become "shorter than they are wide". It seems like a good idea, but it should be optional and perhaps a separate parameter should decide the threshold.

It doesn't really seem like it's working as intended:
https://youtu.be/6kV5kS81WsQ
(notice how the deadzone really is too big. the arrows materialize out of nowhere, losing a huge range of sensitivity.)

Though from reading the code, I can't tell why:
https://github.com/ros-visualization/rviz/blob/7970ba08cee3810cfa1609c3b0f5136970eb2f7c/src/rviz/default_plugin/wrench_visual.cpp#L63

v4hn:
We just hit this bug too. I would propose to add a boolean flag "hide small forces" for this feature. Would that be ok for you @wjwwood ?

wjwwood:
Fine by me.

PR #1196, opened 2018-02-14
hiding small wrenches should be optional

jgueldenstein:
The default behavior for wrench messages was to hide the visualization if the length of the vector was smaller than the width. This pull request adds a boolean property to enable/disable this feature as discussed in  #1081.

I also renamed the function `updateColorAndAlpha()` to `updateProperties()`. The name made no sense anymore (even before) because scale and width are also updated here. The new name clarifies that several properties are updated.

@v4hn

### DIFF
--- a/src/rviz/default_plugin/wrench_display.cpp
+++ b/src/rviz/default_plugin/wrench_display.cpp
@@ -23,38 +23,43 @@ WrenchStampedDisplay::WrenchStampedDisplay()
     force_color_property_ =
             new rviz::ColorProperty( "Force Color", QColor( 204, 51, 51 ),
                                      "Color to draw the force arrows.",
-                                     this, SLOT( updateColorAndAlpha() ));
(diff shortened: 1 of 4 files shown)


v4hn:
@wjwwood friendly ping. This has been discussed in the issue, CI succeeded and it works perfectly in our setup. Would be great to see this merged soon.

dhood:
@jgueldenstein, @v4hn thank you for the PR addressing https://github.com/ros-visualization/rviz/issues/1081

This change will break ABI (because it adds a member to the `WrenchDisplay` class). This means we won't be able to release it into indigo/kinetic/lunar, so will have to hold it for when we release rviz into melodic.

v4hn:
fair enough, thank you!

wjwwood:
Melodic is only failing due to a (now fixed) deprecation warning unrelated to this pr. So I'll merge.

PR #1198, opened 2018-02-15
Add automatic ABI checker from ros-industrial/industrial_ci

dhood:
If I've understood [the documentation](https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst) correctly this should check against the merge parent

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,6 @@ env:
   matrix:
     - ROS_DISTRO="kinetic" ABICHECK_URL='github:ros-visualization/rviz#kinetic-devel' ABICHECK_MERGE=auto
 
-matrix:
-  allow_failures:
-    - env: ROS_DISTRO="kinetic" ABICHECK_URL='github:ros-visualization/rviz#kinetic-devel' ABICHECK_MERGE=auto
-
 install:
   - git clone --depth=1 https://github.com/ros-industrial/industrial_ci.git .ci_config
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+sudo: required
+dist: trusty
+language: generic
+
+cache:
+  directories:
+    - $HOME/.ccache
+
+env:
+  global:
+    - CCACHE_DIR=$HOME/.ccache
+
+  matrix:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+env:
+  global:
+    - CCACHE_DIR=$HOME/.ccache
+
+  matrix:
+    - ROS_DISTRO="kinetic" ABICHECK_URL='github:ros-visualization/rviz#kinetic-devel' ABICHECK_MERGE=auto
+
+matrix:
+  allow_failures:
+    - env: ROS_DISTRO="kinetic" ABICHECK_URL='github:ros-visualization/rviz#kinetic-devel' ABICHECK_MERGE=auto
```
130s: Note that by `allow_failures`-ing, you may not easily notice on GitHub PR view even if the failure occurred  (sorry for being pedantic).

dhood: Happy to accept input, thanks! Sounds like it's best to remove it for increased visibility in the event that it doesn't pass.

dhood:
argh, we're hitting timeouts with travis. I think it's right on the 50min limit because it looks to be at the last step when it gets killed:
```
[build] Runtime: 18 minutes and 16.2 seconds total.                            
[build] Note: Workspace packages have changed, please re-source setup files to use them.
Detect public symbols
Reading debug-info

< this took 5min in the previous step >

The job exceeded the maximum time limit for jobs, and has been terminated.
```
@130s did you happen to have any suggestions for speeding up the travis job?

ipa-mdl:
Looks like the job just took too long, this might happen for bigger projects.
How long does it take to run it locally?

However, travis had some [problems](https://www.traviscistatus.com/incidents/t68jk6rj92q4) in this time frame, so a retry might resolve the issue.

dhood:
Takes 31 min locally. The three jobs so far have been terminated by travis, but in case it was just bad luck I triggered another

ipa-mdl:
> Takes 31 min locally

It will be a little bit slower on Travis, so perhaps it's just to much.
To get a baseline time, you could add a normal build/test (only 1 run).

dhood:
just running the script locally because of the timeout on travis

PR #1205, opened 2018-03-08
Exposing viewport mouse events with a Qt signal

jr-bor:


wjwwood:
See: https://github.com/ros-visualization/rviz/issues/1206#issuecomment-371438135

But, I will not have time to properly test this out right now, so unfortunately you'll have to wait for us to have time to spend on rviz in the coming months.

src/rviz/render_panel.h
```diff
@@ -100,6 +101,9 @@ Q_OBJECT
 
   virtual void sceneManagerDestroyed( Ogre::SceneManager* source );
 
+Q_SIGNALS:
+  void onViewportMouseEvent( const ViewportMouseEvent& event);
```
simonschmeisser: in Qt signals don't start with "on", that is reserved for slots. ie signal mouseEvent -> slot: onMouseEvent

jr-bor:
Hi!
I think this PR and the issue #1206 can be closed. I made this PR to have a more comprehensive question. This solution tackles the problem by changing the principles on how Rviz is made. Thus I would not merge it and I would close it.

I have seen that the new Rviz 2 is being planned and I will put my inputs if needed in there. 

Thanks for the replays :)

Issue #1209, opened 2018-03-15
QtQuick / QML support missing


machinekoder:
Rviz only supports building QtWidgets applications. Although widgets are great for desktop applications, they are very unsuitable for HMIs and modern touch-friendly apps.

Since ROS may sooner or later be used for industrial robot applications, it makes sense to support QtQuick top enable building modern touch-friendly HMIs.

I have a patch upcoming that integrates the Rviz visualization into QtQuick.

machinekoder:
Moving this issue to ROS2: https://github.com/ros2/rviz/issues/268

PR #1210, opened 2018-03-23
Add QtQuick/QML support

machinekoder:
Fix for enhancement described in https://github.com/ros-visualization/rviz/issues/1209

This patch fundamentally changes the RenderPanel concept. In many places, it is assumed that RenderPanel is a QWidget, however, QtQuick is fundamentally different to widget style application.

I need some feedback on whether my approach to abstracting the OgreRenderWindow is sane or not, Also, updating the OpenGL context works completely different in QtQuick, there are potentially more side-effects since the whole application is rendered in OpenGL and not only the Ogre window.

Please also let me know where I should add documentation on how to use Rviz in a QtQuick application.

Any feedback is appreciated.


dhood:
@wjwwood I'm going to defer to you for input on the direction of this change, incl. how it fits in with refactoring efforts at http://github.com/ros2/rviz

wjwwood:
This is an architecturally significant change, and so I honestly cannot guarantee I (or anyone else) will have time to review it and get it merged. The best chance for this is to write more about exactly what is changing and why (like human readable text) and what impact that will have. You've already touched on a few big items like changing the assumption around being a QtWidget, and that you needed to integrate Ogre with Qt differently because QtQuick might be drawn with opengl itself. However, there's always the risk that even if you spend more time on documenting the changes and breaking them into smaller pieces in order to get it merged, that we won't be able to act on it soon.

As @dhood mentioned we're also working on major changes to rviz in the ROS 2 repo:

https://github.com/ros2/rviz

We've updated to Ogre 1.10 in our rviz2 port and changed how rendering is exposed, e.g. it is now a QWindow which Ogre draws in rather than QWidget. Also, we've refactored it into about four separate packages, one of which is Qt + Ogre but ROS agnostic (that's `rviz_rendering`).

So we're in an awkward state where there are major changes, that we wish eventually main line, over in the ROS 2 port, and your changes here. And they'll continue to diverge I think. So I would recommend that you instead work on QtQuick integration on the ROS 2 port since that is the "future" for the code base, but that probably doesn't work for you short term because that port is not mature enough to be pull back to ROS 1 yet.

It's not good news, but that's the state of affairs. I don't see this pr getting merged as-is, but we might be able to merge small incremental changes that work our way towards your end goal. But we always have to be mindful of the ROS 2 port.

There's a team of contributors on the ROS 2 port which might have some input here or interest in making changes to support QtQuick in the ROS 2 port as well. /cc @greimela-si @Martin-Idel-SI @anhosi @botteroa-si

machinekoder:
@wjwwood Thank you for your feedback.

I will outline the architectural changes a little bit more in a later comment. The most important one is that large parts of RViz assume that the RenderPanel is a QWidget, which lead me to me to refactor out this part of the code and it's also the place where incompatibilities may arise. It's definitely something to consider for ROS2 RViz.

For ROS1 I'm happy to work with my private fork, especially since my current work requires a stable ROS release. I also will look into refactoring out the parts of the patch that are in general interesting for RViz, regardless of QtQuick support.

I'm also willing to contribute towards ROS2 RViz support. If considered early in the development, the required changes in the architecture aren't big. In fact, it's just matter of making sure that the core application logic doesn't depend on either QtWidgets or QtQuick.

wjwwood:
> It's definitely something to consider for ROS2 RViz.

The main render window inherits from `QWindow` in ROS 2, but I don't know if that satisfies your requirements:

https://github.com/ros2/rviz/blob/ros2/rviz_rendering/include/rviz_rendering/render_window.hpp#L63

> I also will look into refactoring out the parts of the patch that are in general interesting for RViz, regardless of QtQuick support.

This is your best bet for getting things merged, i.e. changes that are small, incremental, and well described. I wish I could say "just send one huge patch and we'll sort it out", but realistically we don't have time for that ourselves.

> I'm also willing to contribute towards ROS2 RViz support.

That's great! You can follow along on the fork we have for it right now:

https://github.com/ros2/rviz

> If considered early in the development, the required changes in the architecture aren't big. In fact, it's just matter of making sure that the core application logic doesn't depend on either QtWidgets or QtQuick.

That still sounds like a large architectural change to me, but maybe it's not as bad as I'm imagining. I don't expect it will be much easier in the ROS 2 version than the ROS 1 version, though it might be slightly easier due to some changes to the code layout.

machinekoder:
Pointer to the ROS2 issue: https://github.com/ros2/rviz/issues/268

tdnet12434:
Hi, this work is very interesting.
I tried to build but stuck at mapToGlobal is not available in Qt5.9.0 

```
/home/fx/rviz_ws/src/rviz/src/rviz/ogre_helpers/qt_quick_ogre_render_window.cpp: In member function ‘virtual QPoint rviz::QtQuickOgreRenderWindow::mapFromGlobal(const QPoint&) const’:
/home/fx/rviz_ws/src/rviz/src/rviz/ogre_helpers/qt_quick_ogre_render_window.cpp:287:10: error: ‘mapFromGlobal’ is not a member of ‘QQuickItem’
   return QQuickItem::mapFromGlobal(point).toPoint();
          ^
/home/fx/rviz_ws/src/rviz/src/rviz/ogre_helpers/qt_quick_ogre_render_window.cpp: In member function ‘virtual QPoint rviz::QtQuickOgreRenderWindow::mapToGlobal(const QPoint&) const’:
/home/fx/rviz_ws/src/rviz/src/rviz/ogre_helpers/qt_quick_ogre_render_window.cpp:292:10: error: ‘mapToGlobal’ is not a member of ‘QQuickItem’
   return QQuickItem::mapToGlobal(point).toPoint();

```

machinekoder:
@tdnet12434 We are using Qt 5.10.1 here. If you don't need the mouse interaction you could also comment out the `QQuickItem::mapFromGlobal` usages. I think it should be possible to use Qt 5.10 from the SDK in combination with ROS from packages. I'm encountering some problems with the Qt 5.11, which might be related to my graphics card, but to make sure it works, stick with Qt 5.10 for now.

rhaschke:
@machinekoder, can this be closed here as you seem to target ROS2 now (https://github.com/ros-visualization/rviz/issues/1209#issuecomment-390590724)?

machinekoder:
I haven't done any work on ROS2 yet, but we can close this issue. If anyone needs support, he probably can find it.

PR #1213, opened 2018-03-30
add TF Prefix to effort plugin and fix #892

### LINKED ISSUE #892: Crash on Effort plugin
I noticed a crash in the Effort plugin (indigo) : 

```
[ WARN] [1432914059.915302803]: Message from [/lwr_sim/joint_state_publisher] has a non-fully-qualified frame_id [world]. Resolved locally to [world].  This is will likely not work in multi-robot systems.  This message will only print once.

Program received signal SIGSEGV, Segmentation fault.
0x00007fff9829ccc7 in rviz::EffortVisual::EffortVisual(Ogre::SceneManager*, Ogre::SceneNode*, boost::shared_ptr<urdf::Model>) ()
   from /opt/ros/indigo/lib/libdefault_plugin.so
```

It's when I select a valid joint state topic with a wrong _Robot Description_ value : 
- Joint States topic is `/lwr/joint_states`
- Robot description is at `/lwr/robot_description` (**not** at `/robot_description`)

This order cause the crash. Seems like `boost::shared_ptr<urdf::Model>` is null.

If I first input the correct Robot Description (`/lwr/robot_description`), then I select the joint state topic, it's working fine.

ahoarau:
Simple PR to add a TF Prefix property to the Effort Plugin.
Use case : `/robot1/robot_description` + `tf_prefix` set to `robot1`

~Also fixes #892.~

### DIFF
--- a/src/rviz/default_plugin/effort_display.cpp
+++ b/src/rviz/default_plugin/effort_display.cpp
@@ -322,21 +322,21 @@ namespace rviz
 	    int joint_type = joint->type;
 	    if ( joint_type == urdf::Joint::REVOLUTE )
 	    {
-		// we expects that parent_link_name equals to frame_id.
-		std::string parent_link_name = tf_prefix_property_->getStdString() + "/" + joint->child_link_name;
+		std::string tf_prefix = tf_prefix_property_->getStdString();
+		std::string tf_frame_id = (tf_prefix.empty() ? "" : tf_prefix + "/" ) + joint->child_link_name;
 		Ogre::Quaternion orientation;
 		Ogre::Vector3 position;
 
 		// Here we call the rviz::FrameManager to get the transform from the
 		// fixed frame to the frame in the header of this Effort message.  If
 		// it fails, we can't do anything else so we return.
-		if( !context_->getFrameManager()->getTransform( parent_link_name,
+		if( !context_->getFrameManager()->getTransform( tf_frame_id,
                                                                 ros::Time(),
                                                                 //msg->header.stamp, // ???
                                                                 position, orientation ))
 		{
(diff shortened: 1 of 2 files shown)


dhood:
Thanks for the patch, @ahoarau 

@tfoote I have [read that the tf prefix](http://wiki.ros.org/tf2/Migration#Removal_of_support_for_tf_prefix) was causing headaches in tf1, and I've seen errors before about e.g. `base_link` not resolving to `/base_link` well. Would you be able to comment on the impact of [prepending a forward slash](https://github.com/ros-visualization/rviz/pull/1213/files#diff-271abbb34c7fc6f4388e5cb1b6527d59R326) to a frame where it didn't otherwise have one? I'm thinking we should only add the slash if the tf_prefix is specified, to minimise impact.

ahoarau:
You are absolutely right, we should not add leading slash if no tf_prefix is present.
Let's change it for : 

```cpp
std::string tf_prefix = tf_prefix_property_->getStdString();
std::string parent_link_name =  (tf_prefix.empty() ? "" : tf_prefix + "/" ) + joint->child_link_name;
```

Very glad to hear that tf_prefix is meant to disappear, but I'm not sure how I will spawn multiple robots in the same master. Multimaster is quite a big overhead :)

dhood:
That change looks reasonable, but for clarity the variable should now be named `tf_frame_id` as opposed to `parent_link_name`

btw this change will be held for ROS Melodic since it will cause an ABI change

dhood:
I broke out the non-abi breaking change into https://github.com/ros-visualization/rviz/pull/1229 so it can be released into indigo/kinetic

dhood:
@ros-pull-request-builder retest this please.

wjwwood:
@dhood merge or comment

PR #1216, opened 2018-04-06
Created a new Preferences window, and added PromptSaveOnExit option.

dhood:
Redo of #1106 on behalf of @MasterEric , who wrote:

> For tutorials installed to a read-only location, it may be disorienting for the user to be prompted to save. For this reason, I created a preferences dialog window, with a boolean option for whether to Prompt Save on Exit, defaulting to true. These preferences are saved to any config .rviz files that are created.

### DIFF
--- a/src/rviz/CMakeLists.txt
+++ b/src/rviz/CMakeLists.txt
@@ -47,6 +47,7 @@ add_library(${PROJECT_NAME}
   message_filter_display.h
   mesh_loader.cpp
   new_object_dialog.cpp
+  preferences_dialog.cpp
   add_display_dialog.cpp
   ogre_helpers/apply_visibility_bits.cpp
   ogre_helpers/arrow.cpp
--- a/src/rviz/CMakeLists.txt
+++ b/src/rviz/CMakeLists.txt
@@ -47,7 +47,6 @@ add_library(${PROJECT_NAME}
   message_filter_display.h
   mesh_loader.cpp
   new_object_dialog.cpp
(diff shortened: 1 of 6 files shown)


dhood:
@MasterEric thanks for the work adding the generalised preferences dialog! The PromptSaveOnExit option looks good.

As mentioned in https://github.com/ros-visualization/rviz/pull/1106#issuecomment-370589511, the change will break ABI (because it adds a member to the `VisualizationFrame` class). This means we won't be able to release it into indigo/kinetic/lunar, so will hold it for when we release rviz into melodic.

Since we are breaking ABI anyway, I changed your PR to use a generalised `Preferences` struct, so that we will be able to add additional preferences in the future without breaking ABI again.

MasterEric:
@dhood Great idea creating the Preferences struct! I'm not well-versed in what affects the ABI and what causes conflicts in applications.

I'm certainly excited to see my major changes in the upcoming Melodic release!

I'm wondering if you have any opinions/ideas/suggestions for other options that would be useful to include in the Preferences menu.

dhood:
> I'm certainly excited to see my major changes in the upcoming Melodic release!

congrats! even though your original PR had to be replaced by this one; you will still be credited in the changelog :smile: 

> I'm wondering if you have any opinions/ideas/suggestions for other options that would be useful to include in the Preferences menu.

Not necessarily, but these things pop up as we/contributors add different functionality; it will be helpful to be able to add options to this struct without breaking ABI for those features (from melodic onwards)

PR #1221, opened 2018-04-12
allow classes inheriting from image display access to more state

dhood:
replacement of https://github.com/ros-visualization/rviz/pull/1161

### DIFF
--- a/src/rviz/default_plugin/image_display.h
+++ b/src/rviz/default_plugin/image_display.h
@@ -83,17 +83,19 @@ protected:
   /* This is called by incomingMessage(). */
   virtual void processMessage(const sensor_msgs::Image::ConstPtr& msg);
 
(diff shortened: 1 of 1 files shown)


dhood:
I checked with abi-checker and moving from private to protected didn't seem to affect ABI. If it still feels iffy we can certainly hold off

dhood:
Ah yes, changing the ordering does go against what http://www.ros.org/reps/rep-0009.html says, hey.
I'll run it again to double check there wasn't a mistake on my part -- now I'm suspicious too :smile:

dhood:
input from the team is that there's a chance this could break abi on other compilers, so we'll hold this for melodic to be safe.

dhood:
switched to target melodic-devel branch

Issue #1215, opened 2018-04-06
Migrating RViz DisplayContext to tf2


IanTheEngineer:
TLDR: is there any way of coaxing RViz DisplayContext to return a `tf2_ros::Buffer` pointer? If not, what will it take to upgrade RViz to `tf2`?

A little more background: I am in the midst of migrating MoveIt! to `tf2`, hopefully targeting the ROS Melodic release. This migration was requested by @dirk-thomas as the starting point for eventually porting MoveIt! to ROS2: https://github.com/ros-planning/moveit/issues/745

One snag that I've run into is that RViz' [`DisplayContext` `FrameManager` returns a `tf::TransformListener`](https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/display_context.h#L89-L93), rather than a `tf2_ros::Buffer`, even on the `ros2` branch:
```
/** @brief Return the FrameManager instance. */
virtual FrameManager* getFrameManager() const = 0;

/** @brief Convenience function: returns getFrameManager()->getTFClient(). */
virtual tf::TransformListener* getTFClient() const = 0;
```
MoveIt! uses this pointer to initialize its [`MoveGroupInterface` for the `motion_planning_rviz_plugin`](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp#L302-L303)
```
move_group_.reset(new moveit::planning_interface::MoveGroupInterface(
opt, context_->getFrameManager()->getTFClientPtr(), ros::WallDuration(30, 0)));
```
It will need to receive a pointer to a `tf2_ros::Buffer` object. I have ported nearly all of MoveIt! to `tf2`, and this is (hopefully) one of the last roadblocks :)

dhood:
While there's a ros2 branch on this repo, it's not up-to-date; the work moved to a fork, where tf2 is being used:
https://github.com/ros2/rviz/blob/2398b5055a225bcdf5bdf838124a6af4f5c93c97/rviz_common/include/rviz_common/frame_manager_iface.hpp#L240 (see https://github.com/ros2/rviz/issues/52 for updates on buffer vs listener)

(@wjwwood I might remove that branch so it doesn't mislead people, yeah?)

[...]

PR #1224, opened 2018-04-20
Add tf2 Buffer accessor in FrameManager

IanTheEngineer:
As this modifies headers (though only **adding** API), this may need to target `melodic-devel`.

This PR would mitigate the Issue related to using tf2 in the FrameManager https://github.com/ros-visualization/rviz/issues/1215 by creating a `tf2_ros` `TransformListener` and `Buffer` for use by those who wish to use the RViz API and rely solely on the `tf2` libraries. In particular, MoveIt! is in the middle of a migration to tf2 ( https://github.com/ros-planning/moveit/pull/830 ) and this PR would unblock that migration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,7 @@ find_package(catkin REQUIRED
   std_msgs
   std_srvs
   tf
+  tf2_ros
   urdf
   visualization_msgs
   urdfdom_headers
--- a/package.xml
+++ b/package.xml
@@ -43,6 +43,7 @@
   <build_depend>std_msgs</build_depend>
   <build_depend>std_srvs</build_depend>
   <build_depend>tf</build_depend>
+  <build_depend>tf2_ros</build_depend>
   <build_depend>tinyxml</build_depend>
   <build_depend>urdf</build_depend>
   <build_depend>visualization_msgs</build_depend>
@@ -77,6 +78,7 @@
   <run_depend>std_msgs</run_depend>
   <run_depend>std_srvs</run_depend>
   <run_depend>tf</run_depend>
+  <run_depend>tf2_ros</run_depend>
   <run_depend>tinyxml</run_depend>
   <run_depend>urdf</run_depend>
   <run_depend>visualization_msgs</run_depend>
--- a/src/rviz/frame_manager.cpp
+++ b/src/rviz/frame_manager.cpp
@@ -32,6 +32,8 @@
 #include "properties/property.h"
 
 #include <tf/transform_listener.h>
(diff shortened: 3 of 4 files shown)


IanTheEngineer:
@dhood @wjwwood any chance you could give this PR a lookover? 

The benefits include allowing users of RViz' API to transition away from the deprecated `tf` library, toward `tf2`, by utilizing a shared `Buffer` object. 

As far as I can tell, with this implementation, the only downside is that there lacks an option to re-use a `tf2_ros::Buffer` passed in as an argument to the constructor, but I can add that if it is desired.

dirk-thomas:
@wjwwood Can you please take a look at this since it is necessary for the migration of MoveIt to tf2. Thanks.

dhood:
you are right that this will break ABI and will need to target melodic.

the downside that I see is that we now have duplicated the listener effort for TF, which is a shame since the tf1 listener is actually using a tf2 buffer underneath. I know you had worked to expose it in https://github.com/ros-visualization/rviz/issues/1215#issuecomment-379493477; did you consider that approach here already?

IanTheEngineer:
@dhood I agree that reusing the tf listener would be ideal, but the [underlying `tf2_ros::Buffer` object](https://github.com/ros/geometry/blob/indigo-devel/tf/include/tf/tf.h#L387) is not designed to be accessible outside of inheretance.
I ran into slicing issues in attempting to retrieve the pre-alocated & protected `tf_buffer` object and maintain its access inside a shared pointer, without it being deleted from underneath the `shared_ptr` and causing a coredump. In the end, I figure the best solution is probably the simplest (implemented in this PR) even if it is not the most efficient. However, I am open to recommendations on a better implementation (or pointers as to how my C++ foo is failing).

dhood:
@wjwwood could you pitch in on this please if you get a chance? (and the split `std::shared_ptr` and `boost::shared_ptr` interface: maybe stick to boost for consistency in the API?)

wjwwood:
Yeah, I was hoping to do it today. We'll see...

wjwwood:
I'd really rather not have duplicated tf buffers. So I'm going to pursue why the subclass didn't work and/or adding a new method in tf that would give access to the underlying `tf2::Buffer` without subclassing.

wjwwood:
Closing since I believe this was worked around in https://github.com/ros/geometry/pull/163. Please comment if that's not the case.

IanTheEngineer:
Thanks @wjwwood! I meant to close it myself. I believe this PR is unneeded due to the accessor method added to https://github.com/ros/geometry/pull/163 Thanks for the reviews :)

PR #1231, opened 2018-04-27
change to use non-deprecated header

wjwwood:
Looks like it was missed in this pr or has since regressed: https://github.com/ros-visualization/rviz/pull/1217


dhood:
Turns out it was missed by the auto-PR because of the `# include` format

mikaelarguedas:
> Looks like it was missed in this pr

Yeah, as mentionned to @dhood  offline, I didnt account for whitespaces when writing my search replace script :/. The last released version of pluginlib should not throw the warning anymore so there shouldnt be any rush releasing this

Issue #1235, opened 2018-05-09
rviz is using deprecated tinyxml interface of URDF


wjwwood:
It produces a warning like this in melodic:

```
[ 81%] Building CXX object src/rviz/default_plugin/CMakeFiles/rviz_default_plugin.dir/robot_model_display.cpp.o
/home/william/melodic/src/rviz/src/rviz/default_plugin/robot_model_display.cpp: In member function ‘virtual void rviz::RobotModelDisplay::load()’:
/home/william/melodic/src/rviz/src/rviz/default_plugin/robot_model_display.cpp:185:41: warning: ‘bool urdf::Model::initXml(TiXmlElement*)’ is deprecated: TinyXML API is deprecated, use the TinyXML2 version instead [-Wdeprecated-declarations]
   if( !descr.initXml( doc.RootElement() ))
                                         ^
In file included from /home/william/melodic/src/rviz/src/rviz/default_plugin/robot_model_display.cpp:34:0:
/home/william/melodic/install/include/urdf/model.h:61:99: note: declared here
   URDF_EXPORT URDF_DEPRECATED("TinyXML API is deprecated, use the TinyXML2 version instead") bool initXml(TiXmlElement * xml);
                                                                                                   ^~~~~~~
```

PR #1236, opened 2018-05-09
Deprecate tf for tf2

wjwwood:
This pr deprecates all of the interfaces in rviz which I believe need to be so that in a future version we can drop the dependency on tf completely.

The two main places where interfaces are disabled are:

- `VisualizationManager`: b1bb6a2
- `FrameManager`: 7ee47f3

I migrated the `MessageFilterDisplay` which is a base class that many display plugins inherit from and therefore most of them will require no changes at all. I also migrated a couple other displays to show how it can be done, but I just suppressed the rest of the displays that produced deprecation warnings because I wanted to give an example of how to do that and because the work to convert them can be deferred until we actually remove the dependency on `tf` completely.

Examples of migrating to `tf2`:

- `MessageFilterDisplay`: 0f227e5
- Image related displays (image and camera): ffdfad0
- `DepthCloudDisplay`: 2dc400d

Examples of suppressing the deprecation warnings:

- `GridCellsDisplay`: 89b51e7
- `TFDisplay`: c7f6a4a

There are a few unrelated style/warnings related changes at the end, which I could separate if desired, but they're small and mostly only affect macOS.

### DIFF
--- a/src/rviz/default_plugin/axes_display.cpp
+++ b/src/rviz/default_plugin/axes_display.cpp
@@ -29,8 +29,19 @@
 
 #include <boost/bind.hpp>
 
+#ifndef _WIN32
+# pragma GCC diagnostic push
+# ifdef __clang__
+#  pragma clang diagnostic ignored "-W#warnings"
+# endif
+#endif
+
 #include <OgreSceneNode.h>
 #include <OgreSceneManager.h>
+
+#ifndef _WIN32
+# pragma GCC diagnostic pop
+#endif
 
 #include "rviz/display_context.h"
 #include "rviz/frame_manager.h"
--- a/src/rviz/default_plugin/camera_display.cpp
+++ b/src/rviz/default_plugin/camera_display.cpp
@@ -29,6 +29,13 @@
 
 #include <boost/bind.hpp>
 
+#ifndef _WIN32
+# pragma GCC diagnostic push
+# ifdef __clang__
+#  pragma clang diagnostic ignored "-W#warnings"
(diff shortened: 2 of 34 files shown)


wjwwood:
@IanTheEngineer FYI

I plan on closing the other pr. Also, this will probably introduce a deprecation warning in moveit (sorry about that), but I'm happy to make a pr for that if you can point me to where you needed to change things there. The warnings are easy to either address or suppress. There are some examples above in the issue description.

wjwwood:
I'll also mention that for the changes to the Constructor of `VisualizationManager` and `FrameManager` which optionally take a `tf::TransformListener`, I deprecated them but could not offer the alternative right now. The reason is that if someone gives us a `tf2` equivalent I cannot provide the `tf` version of the client when someone asks for it via `getTFClient()`. So they'll have to suppress the warning and then when we completely remove `tf` I can provide a new constructor that can take the `tf2` equivalent.

IanTheEngineer:
That makes sense. I’m out on vacation this week, but I’ll point you to the right spot to help suppress the deprecation warnings next week when I return (around 5/16). There’s only two lines in all of MoveIt using the FrameManager

wjwwood:
The CI looks good to me, the remaining failure is in `robot_model_display.cpp` and should be addressed separately, see: https://github.com/ros-visualization/rviz/issues/1235

wjwwood:
I'll add an note on this wiki page when this is merged:

https://wiki.ros.org/melodic/Migration

PR #1238, opened 2018-05-10
Add a checkbox to toggle Tool keyboard shortcuts.

AndyZe:
We had an "incident" where a hotkey for another executable was 'g'. Of course, that's the same hotkey for sending a move_base command in RViz so the robot took off unexpectedly.

I'd like to allow hotkeys to be disabled for that reason.

By default, they will still be enabled, so this should be a transparent change for most users.

### DIFF
--- a/src/rviz/tool.h
+++ b/src/rviz/tool.h
@@ -160,6 +160,8 @@ public:
   const QCursor& getCursor() { return cursor_; }
 
   void setStatus( const QString & message );
+
+  bool access_all_keys_;
   
 Q_SIGNALS:
     void close();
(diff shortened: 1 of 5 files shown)


wjwwood:
@AndyZe unfortunately this is an ABI breaking change due to moving and adding the member variables on a public class, therefore we'd need to target it for `melodic-devel` only (and the window is closing soon on that one too).

Also, @dhood mentioned to me that there's a new preferences class that _might_ be a useful place to put your new option. See: https://github.com/ros-visualization/rviz/pull/1216

wjwwood:
I went a head and retargeted it to `melodic-devel` which caused a merge conflict. You can either fix the conflict or I can point it back at `kinetic-devel` at your request if you'd like to try to make it ABI compatible (I'm not sure that's possible however).

PR #1242, opened 2018-05-11
Fix compatibility with Ogre 1.11

de-vri-es:
This PR makes rviz compatible with ogre 1.11 while maintaining compatibility with 1.7.

This PR:
* adds a cross-version compatibility function for setting materials on a `SimpleRenderable`.
* prevents removing materials that are unknown to the resource manager (I may have missed some).
* loads the now split-off `Codec_STBI` plugin for ogre >= 1.11, required for PNG decoding.
* avoids the use of relative paths in shader programs because they break in Ogre 1.11.
* prevents removing anonymous scene nodes by name.
* makes sure to preserve resource group names when referring to materials by name.
* adds a `BaseWhiteNoLighting` material to the "rviz" resource group, because it seems impossible to specify a resource group for materials in a `.mesh` file.

The main changes in Ogre that make this PR necessary are the removal of `SimpleRenderable::setMaterial(const std::string & name)` and the now default resource manager strict mode. This strict mode implies that resources are not looked up in alternative resource groups anymore.

I think I managed to keep things compatible with atleast Ogre 1.7. I checked the headers of 1.7 and 1.11 when making use of different APIs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,15 +8,6 @@ if (POLICY CMP0054)
   cmake_policy(SET CMP0054 NEW)
 endif()
 
-find_package(Boost REQUIRED
-  COMPONENTS
-  filesystem
-  program_options
-  signals
-  system
-  thread
-)
-
 find_package(urdfdom_headers REQUIRED)
 
 find_package(PkgConfig REQUIRED)
@@ -141,9 +132,14 @@ find_package(catkin REQUIRED
   urdfdom_headers
 )
 
+set(RVIZ_BOOST_COMPONENTS filesystem program_options system thread)
+
 if(${tf_VERSION} VERSION_LESS "1.11.3")
   add_definitions("-DRVIZ_USE_BOOST_SIGNAL_1")
+  list(APPEND RVIZ_BOOST_COMPONENTS signals)
 endif()
+
+find_package(Boost REQUIRED COMPONENTS ${RVIZ_BOOST_COMPONENTS})
 
 find_package(PythonLibs "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" REQUIRED)
 
--- a/ogre_media/fonts/liberation_sans.fontdef
+++ b/ogre_media/fonts/liberation_sans.fontdef
@@ -1,4 +1,4 @@
-Liberation Sans
+font Liberation Sans
 {
   type truetype
   source liberation-sans/LiberationSans-Regular.ttf
--- a/ogre_media/materials/glsl120/glsl120.program
+++ b/ogre_media/materials/glsl120/glsl120.program
@@ -1,8 +1,8 @@
 
 //includes:
-fragment_program rviz/glsl120/include/circle_impl.frag glsl { source include/circle_impl.frag }
-fragment_program rviz/glsl120/include/pack_depth.frag glsl { source include/pack_depth.frag }
-vertex_program rviz/glsl120/include/pass_depth.vert glsl { source include/pass_depth.vert }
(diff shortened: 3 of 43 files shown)


de-vri-es:
Added more compatibility fixes. I updated the PR description with details.

de-vri-es:
Been testing this for a while now, it seems to run stable.

Any feedback?

G-Pereira:
I can confirm that this works and is stable

G-Pereira:
thanks @de-vri-es god job!

de-vri-es:
Any progress on a review?

de-vri-es:
I don't want to seem impatient, but it's been almost a month without a word from a maintainer. In the mean time, more work on Ogre 1.11 is probably needed, but the longer unmerged PRs like this stay open the harder it will be to consolidate efforts afterwards.

wjwwood:
@de-vri-es We're not likely to have any time to work on this in the next few weeks. I'm sorry that is the case, but one of us will address it as we have time. It's in the queue to be reviewed (added the review label 16-days ago).

Normally, I'd just do a quick code review and merge it, since it builds on CI and has positive reviews from other users, but because this touches the mesh visualization stuff, I'd want to build it and run it on Ubuntu 16.04 and checkout some of the manually run acceptance tests.

Again, apologies for making you wait, but we're doing the best we can with our limited maintenance time.

de-vri-es:
Alright, thanks for the explanation. If there's anything I can do to help you guys test this, let me know. Maybe I can perform some of these test on Ubuntu 16.04 and document the result. Is there a check list of sorts to work though?

wjwwood:
Basically there is a marker demo in the tests folder which produces an expect grid of meshs and objects with various texture, color, and alpha settings, e.g.: https://github.com/ros-visualization/rviz/pull/1132#pullrequestreview-54439551

I believe it's this one:

https://github.com/ros-visualization/rviz/blob/melodic-devel/src/test/mesh_marker_test.cpp

If you build with tests then it should be somewhere in the test folder.

Sorry it's not better structured or automated than that, but we're working on that in the ROS 2 port (which hopefully will get mainlined in the medium-term future).

de-vri-es:
Thanks. When I get round to it (hopefully soon™) I'll post some screenshots on Arch and Ubuntu 16.04

hgaiser:
Any update on this?

hgaiser:
I've updated the PR because Ogre 1.11.3 changed the definition of `Ogre::Vector3` to be a `typedef` instead of a class and `rviz` tries to forward declare `class Vector3`. Instead of changing `class` to `typedef` in `rviz` (which would have to be in a preprocessor directive), I opted to include `OgrePrerequisites.h`, which forward declares most (everything?) things in Ogre.

rhaschke:
Sorry @de-vri-es for the long delay. There wasn't an active maintainer of rviz for a while.
I jumped in recently and try to process the open PRs now. Material handling has changed recently (#1294) to fix a memory leak accumulating shared material pointers in robot links. Could you please rebase your branch on the latest head and also provide test images as promised in https://github.com/ros-visualization/rviz/pull/1242#issuecomment-396331107 :wink:

rhaschke:
Dear @de-vri-es and @ProfFan,
I know that this PR and #1252 are already open for quite a time.
Could you please consolidate your two PRs into one and rebase onto the current melodic-devel branch?
My suggestions is as follows:
- rebase the #1242 onto melodic-devel, resolving conflicts
  (we have remove boost::signals1 (f230eabd2953ede985146f45f707c6563a3387cc), and fixed leakages in material handling (#1294))
- rebase #1252's additional commits on top of this

rhaschke:
Closing in favor of #1434. Thanks for this great initial work, @de-vri-es!

PR #1243, opened 2018-05-11
Add a checkbox to toggle Tool hotkeys.

AndyZe:
A followup to #1238

### DIFF
--- a/src/rviz/tool.h
+++ b/src/rviz/tool.h
@@ -160,6 +160,8 @@ public:
   const QCursor& getCursor() { return cursor_; }
 
   void setStatus( const QString & message );
+
+  bool access_all_keys_;
   
 Q_SIGNALS:
     void close();
@@ -175,7 +177,6 @@ protected:
   DisplayContext* context_;
 
   char shortcut_key_;
-  bool access_all_keys_;
 
   QIcon icon_;
 
--- a/src/rviz/tool.h
+++ b/src/rviz/tool.h
@@ -78,6 +78,8 @@ public:
 
   bool accessAllKeys() { return access_all_keys_; }
 
+  void setKeyboardAccess(bool enable_keyboard_access) { access_all_keys_ = enable_keyboard_access; }
+
   virtual void activate() = 0;
   virtual void deactivate() = 0;
 
@@ -160,8 +162,6 @@ public:
   const QCursor& getCursor() { return cursor_; }
 
   void setStatus( const QString & message );
-
-  bool access_all_keys_;
   
 Q_SIGNALS:
     void close();
@@ -177,6 +177,8 @@ protected:
   DisplayContext* context_;
 
   char shortcut_key_;
+
+  bool access_all_keys_;
 
   QIcon icon_;
 
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -599,9 +599,13 @@ void VisualizationFrame::openPreferencesDialog()
     preferences_ = boost::make_shared<Preferences>( temp_preferences );
 
     if (preferences_->allow_tool_kb_shortcuts)
+    {
       manager_->getToolManager()->enableKeyboardShortcuts();
+    }
     else
+    {
       manager_->getToolManager()->disableKeyboardShortcuts();
+    }
   }
   manager_->startUpdate();
 }
@@ -944,9 +948,13 @@ void VisualizationFrame::loadPreferences( const Config& config )
   config.mapGetBool( "AllowToolKBShortcuts", &(preferences_->allow_tool_kb_shortcuts) );
 
   if (preferences_->allow_tool_kb_shortcuts)
+  {
     manager_->getToolManager()->enableKeyboardShortcuts();
+  }
   else
+  {
     manager_->getToolManager()->disableKeyboardShortcuts();
+  }
 }
 
 void VisualizationFrame::savePreferences( Config config )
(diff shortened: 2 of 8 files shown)


AndyZe:
Would love to get a review before the Melodic release. Should be quick...

src/rviz/tool.h
```diff
@@ -175,7 +177,6 @@ Q_OBJECT
   DisplayContext* context_;
 
   char shortcut_key_;
-  bool access_all_keys_;
```
dhood: please create getters/setters for this instead of making it public

src/rviz/visualization_frame.h
```diff
@@ -32,6 +32,7 @@
 
 #include <boost/shared_ptr.hpp>
 
+#include <QCheckBox>
```
dhood: is this change just leftover from https://github.com/ros-visualization/rviz/pull/1238?

src/rviz/visualization_frame.cpp
```diff
@@ -598,6 +597,11 @@ void VisualizationFrame::openPreferencesDialog()
   if( dialog->exec() == QDialog::Accepted ) {
     // Apply preferences.
     preferences_ = boost::make_shared<Preferences>( temp_preferences );
+
+    if (preferences_->allow_tool_kb_shortcuts)
```
dhood: please add curly braces around all the if/else statements

AndyZe:
FYI I won't be making any more PR's until this gets approved. I think that's fair.

The hotkeys are very annoying to me and I think there are hundreds of users out there who don't realize why their interactive markers have suddenly disappeared. (Hint: it's because they pressed a hot key and didn't realize it)

wjwwood:
> FYI I won't be making any more PR's until this gets approved. I think that's fair.

We'll get to this pull request as soon as we have some time to do so. I'm sorry to hear you won't be making any additional pull requests (assuming you have some ready) until this is approved, but if your intention is to make this go faster or get us to pay more attention to it by withholding future improvements, that's not going to work because it doesn't give us any extra time in which to look at these things.

> The hotkeys are very annoying to me and I think there are hundreds of users out there who don't realize why their interactive markers have suddenly disappeared. (Hint: it's because they pressed a hot key and didn't realize it)

Our inability to get this in before the release or having time since to look at it has nothing to do with how useful it is, and everything to do with time constraints. We will get to it when we have time, that's the best we can do.

Sorry, and thanks for your patience.

AndyZe:
Yep. I'm not trying to be a dick, but there's no point in making PRs that get reviewed but not merged. Or not even reviewed.

AndyZe:
Maybe just pass the concept that the natives are restless up the command chain...

awatson3:
Status?

AndyZe:
They are too busy even to hire an intern to review these minor pull requests.

rhaschke:
Dear @AndyZe,

I took over maintainership for rviz recently and I'm trying to merge all open PRs.
As this one had conflicts, I rebased and cleaned up a little bit in my own fork: https://github.com/rhaschke/rviz/tree/toggle-hotkeys

Doing so, I stumbled over your commit message in https://github.com/ros-visualization/rviz/pull/1243/commits/78b7c2f45610b894dcaec25a795f77c7198bcfcd:

> Disable hotkeys works with the Preferences dialog, **but logic is flipped**?

Looking into this issue in more detail, I think that `Tool::access_all_keys_` is the wrong variable to use.
This variable [determines whether key presses should be handled by the current tool](https://github.com/ros-visualization/rviz/blob/3a41f37bdfe0959f305ab9a50e7aebdfcf1642db/src/rviz/tool_manager.cpp#L156-L161), even though the pressed key corresponds to a selection hotkey for another tool. This is a private property of the specific tool and, in my opinion, completely unrelated to your goal of disabling all tool selection hotkeys.
I'm not even sure that this was your actual goal. _Selecting_ the `NavGoal` tool and actually _sending_ a goal message should be two different things, right?

A really clean solution would be to make hotkeys completely configurable. Then, one could selectively disable specific hotkeys only.

AndyZe:
:+1:  I'm glad somebody responsive took ownership!  Unfortunately my personal motivation and available time to fix this issue is gone  :/

rhaschke:
> Unfortunately my personal motivation and available time to fix this issue is gone.

While I can understand your disappointment, please contribute to this issue.
I "just" took maintainership to save this project from death, although I'm already heavily busy with maintaining MoveIt! and several other projects. I cannot to this job without the active help of the community.
I thought that I could quickly merge this PR after resolving the conflicts. But I think, this PR is conceptually wrong. Either you can clarify, why I am wrong or you can propose a valid solution.

AndyZe:
Like many people, I often type in a different window while I have RViz open, and that was causing some safety issues. For my use case, I wanted a way to disable ALL hotkeys.

I agree that enabling or disabling individual hotkeys would be a nice feature and it would also accomplish what I needed. Is it worth implementing at this stage in RViz for ROS1? How would it be implemented? I'm not sure

rhaschke:
You are probably right that nobody has resources to implement individual hotkeys...
However, your current approach doesn't work out as well. I was thinking about a global flag used in ToolManager to suppress all hotkeys.

Could you please explain in more detail your safety issue. Just pressing 'g' only switches to the NavGoal tool. To actually send a goal, you also need to interact with the mouse, don't you?

By the way, another option to disable NavGoal, is to simply remove the tool. Would this be an option for you as well?

AndyZe:
I think what happened was, the student opened a terminal with Ctrl+Alt+t and started typing. Then he clicked back in RViz to change the view, accidentally sending a base motion command. I've also seen people accidentally press 'm' which hides the interactive marker. Then they can't move the robot.

I think your global flag idea sounds promising. I should have some bandwidth in about a week.

rhaschke:
> I think your global flag idea sounds promising. I should have some bandwidth in about a week.

Great. Thanks a lot. You probably want to start off my already rebased branch: https://github.com/rhaschke/rviz/tree/toggle-hotkeys

rhaschke:
Closing in favour of #1363.

PR #1245, opened 2018-05-21
Small code cleanup

machinekoder:
Code cleanups from https://github.com/ros-visualization/rviz/pull/1210


src/rviz/ogre_helpers/render_widget.cpp
```diff
@@ -48,7 +48,7 @@ namespace rviz
 RenderWidget::RenderWidget( RenderSystem* render_system, QWidget *parent )
   : QWidget( parent )
   , render_system_( render_system )
-  , render_window_( 0 )
+  , render_window_( nullptr )
```
simonschmeisser: I wonder why that compiles on xenial, it uses g++ 5.3.1 which should default to C++98 which should not know about nullptr?

Otherwise I think this stuff is trivial and should be merged (in melodic)

PR #1246, opened 2018-05-23
Transform poses in Path message correctly into fixed frame.

wngreene:
Orientations were not being transformed into fixed frame in the Path plugin.

### DIFF
--- a/src/rviz/default_plugin/path_display.cpp
+++ b/src/rviz/default_plugin/path_display.cpp
@@ -497,13 +497,11 @@ void PathDisplay::processMessage( const nav_msgs::Path::ConstPtr& msg )
       for( uint32_t i=0; i < num_points; ++i)
       {
         const geometry_msgs::Point& pos = msg->poses[ i ].pose.position;
+        const geometry_msgs::Quaternion& quat = msg->poses[ i ].pose.orientation;
         Ogre::Vector3 xpos = transform * Ogre::Vector3( pos.x, pos.y, pos.z );
+        Ogre::Quaternion xquat = orientation * Ogre::Quaternion( quat.w, quat.x, quat.y, quat.z );
         axes_vect[i]->setPosition(xpos);
-        Ogre::Quaternion orientation_i(msg->poses[ i ].pose.orientation.w,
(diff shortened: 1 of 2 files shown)


src/rviz/default_plugin/path_display.cpp
```diff
@@ -499,11 +499,11 @@ void PathDisplay::processMessage( const nav_msgs::Path::ConstPtr& msg )
         const geometry_msgs::Point& pos = msg->poses[ i ].pose.position;
         Ogre::Vector3 xpos = transform * Ogre::Vector3( pos.x, pos.y, pos.z );
         axes_vect[i]->setPosition(xpos);
-        Ogre::Quaternion orientation(msg->poses[ i ].pose.orientation.w,
-                                     msg->poses[ i ].pose.orientation.x,
-                                     msg->poses[ i ].pose.orientation.y,
-                                     msg->poses[ i ].pose.orientation.z);
-        axes_vect[i]->setOrientation(orientation);
```
rhaschke: Couldn't you simply use
`axes_vect[i]->setOrientation(transform * orientation);`
here?

src/rviz/default_plugin/path_display.cpp
```diff
@@ -499,11 +499,11 @@ void PathDisplay::processMessage( const nav_msgs::Path::ConstPtr& msg )
         const geometry_msgs::Point& pos = msg->poses[ i ].pose.position;
         Ogre::Vector3 xpos = transform * Ogre::Vector3( pos.x, pos.y, pos.z );
         axes_vect[i]->setPosition(xpos);
-        Ogre::Quaternion orientation(msg->poses[ i ].pose.orientation.w,
```
rhaschke: Please use `const geometry_msgs::Quaternion& quat = msg->poses[ i ].pose.orientation` as a shortcut here.
I know, it wasn't used before, but if we touch this code anyway ...

wjwwood:
> Is there a silent agreement that those stamps should be ignored? @wjwwood, can you shed light on this?

Not that I'm aware. I don't have any insight based on my experience.

PR #1248, opened 2018-05-31
add minimum required pluginlib version

mikaelarguedas:
Follow-up of https://github.com/ros-visualization/rviz/pull/1217

For users trying to build rviz with an outdated pluginlib version

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -31,7 +31,7 @@
   <build_depend>map_msgs</build_depend>
   <build_depend>message_filters</build_depend>
   <build_depend>nav_msgs</build_depend>
-  <build_depend>pluginlib</build_depend>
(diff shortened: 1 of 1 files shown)


wjwwood:
Is this the right target branch? Or should it be melodic?

wjwwood:
I guess since the other was for kinetic, this is fine. Does it need to be forward ported or maybe the forward port needs different version requirements? Or maybe it doesn’t need forward port because the required changes are part of melodic from the start.

mikaelarguedas:
As the very first version of pluginlib released in Melodic provided both headers I didn't think it was necessary to put the requirement there. If you do add it, the first version of pluginlib released in Melodic was 1.12.0

PR #1255, opened 2018-06-11
Add properties for the covariance of the initial pose tool

ipa-fez:
Currently the initial pose tool has a hardcoded covariance matrix. Sometimes it is useful to send an initial pose with a low covariance. This PR adds properties to the initial pose tool that allows specifying the main elements of the covariance matrix.

### DIFF
--- a/src/rviz/default_plugin/tools/initial_pose_tool.cpp
+++ b/src/rviz/default_plugin/tools/initial_pose_tool.cpp
@@ -33,6 +33,7 @@
 
 #include "rviz/display_context.h"
 #include "rviz/properties/string_property.h"
+#include "rviz/properties/float_property.h"
 
 #include "rviz/default_plugin/tools/initial_pose_tool.h"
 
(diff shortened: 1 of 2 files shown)


rhaschke:
I recently took over maintenance of rviz and I'm trying to catch up with open PRs.
I like the idea behind this PR, but I suggest to modify the interface to enter std deviations instead of variance values. Requiring the user to enter sigma^2 seems odd. What do you think?
Could you also target this PR to Melodic, please?

ipa-fez:
Thanks for looking at this PR! I've changed it to std deviations and targeted it against melodic-devel.

rhaschke:
Unfortunately, adding new member variables breaks ABI compatibility:

```
Size of InitialPoseTool has been increased from 320 bytes to 344 bytes.
Effects:
1) An object of this class can be allocated by the applications and old size will be hardcoded at the compile time. Call of any exported constructor will break the memory of neighboring objects on the stack or heap.
2) The memory layout and size of subclasses will be changed.
```

Personally, I don't think that this is critical, because probably nobody instantiates the InitialPoseTool manually. The common usage should be instantiating this class via the plugin mechanism, which ensures that the proper, new version is allocated.
However, @wjwwood could you please have a short look at this ABI breakage?

wjwwood:
My hard and fast rule is "no ABI breaks in existing releases", but as you pointed out @rhaschke, in this case it might be protected by the plugin interface. So I think it's fine to make a judgement call here to allow it if you want.

PR #1256, opened 2018-06-11
Don't hide the toolbar when pressing esc

ipa-fez:
The fullscreen code does not properly keep track of the toolbar visibility state. This leads to the following undesired behaviour:
- Pressing esc without ever having entered fullscreen may hides the toolbar because toolbar_visible_ is never set.
- Pressing esc after having entered fullscreen previously sets the toolbar visibility to what it was the last time fullscreen was entered.

By only calling setFullScreen(false) after we really were in fullscreen mode, we guarantee that toolbar_visible_ is valid and don't mess with the toolbar visibility otherwise.


rhaschke:
Thanks for this contribution. To avoid breaking ABI, I removed the addition bool member variable and figure out the full-screen state on the fly.

PR #1269, opened 2018-07-16
Added missing PoseWithCovariance icon

reinzor:


wjwwood:
Cool, thanks! I assume this is a new icon made by you? If so are you ok with it being under "public domain" as the other icons are?

Also please add yourself to the README (as part of this pull request) to that effect:

https://github.com/ros-visualization/rviz/blob/melodic-devel/README.md

reinzor:
Yes, just updated the PR :)

PR #1275, opened 2018-08-03
systematic check of incoming markers

jgueldenstein:
Using rviz Markers for debugging is often very useful, but sometimes the developer makes mistakes when filling in the fields of the marker.
In some cases the resulting markers are not visible or not displayed as intended.
In a few cases this lead to long debugging sessions when the actual problem was the visualization...

Until now there was little and unorganized debug output from rviz to assist the developer in some of these cases.

I have implemented a systematic check of all marker types as well as marker arrays
as described [here](http://wiki.ros.org/rviz/DisplayTypes/Marker) and implemented in the Display.

During testing I found some other bugs as well:

- `TEXT_VIEW_FACING` Markers with an empty text cause rviz to segfault.
- `TEXT_VIEW_FACING` Markers with text that consists exclusively of whitespace trigger a vertex buffer overflow.

These two bugs are resolved in commit d30cda4 

- Markers with an undefined type cause rviz to crash

This is resolved in commit 9c15d9a

Apart from the bugfixes, this pull-request does not change the rendering behavior of rviz.
It simply provides additional warning output to the console and to the display status.

We would like to see this merged in kinetic&upwards, so the request intentionally targets `kinetic-devel`.

@v4hn

### DIFF
--- a/src/rviz/default_plugin/marker_utils.cpp
+++ b/src/rviz/default_plugin/marker_utils.cpp
@@ -232,12 +232,12 @@ void checkQuaternion(const visualization_msgs::Marker& marker, std::stringstream
     && marker.pose.orientation.z == 0.0
     && marker.pose.orientation.w == 0.0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "uninitialized quaternion assuming identity";
   }
   else if(!validateQuaternions(marker.pose))
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "unnormalized quaternion in marker message";
   }
 }
@@ -252,11 +252,11 @@ void checkScale(const visualization_msgs::Marker& marker, std::stringstream& ss)
   {
     if(marker.type == visualization_msgs::Marker::TRIANGLE_LIST)
     {
-      addNewlineIfRequired(ss);
+      addCommaAndNewlineIfRequired(ss);
       ss << "scale contains 0.0 in x, y or z, TRIANGLE_LIST coordinates are scaled";
       return;
     }
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "scale contains 0.0 in x, y or z";
   }
 }
@@ -265,12 +265,12 @@ void checkScaleLineStripAndList(const visualization_msgs::Marker& marker, std::s
 {
   if(marker.scale.x == 0.0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "width LINE_LIST or LINE_STRIP is 0.0 (scale.x)";
   }
   else if(marker.scale.y != 0.0 || marker.scale.z != 0.0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "scale.y and scale.z of LINE_LIST or LINE_STRIP are ignored";
   }
 }
@@ -279,12 +279,12 @@ void checkScalePoints(const visualization_msgs::Marker& marker, std::stringstrea
 {
   if(marker.scale.x == 0.0 || marker.scale.y == 0.0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "width and/or height of POINTS is 0.0 (scale.x, scale.y)";
   }
   else if(marker.scale.z != 0.0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "scale.z of POINTS is ignored";
   }
 }
@@ -293,12 +293,12 @@ void checkScaleText(const visualization_msgs::Marker& marker, std::stringstream&
 {
   if(marker.scale.z == 0.0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "text height of TEXT_VIEW_FACING is 0.0 (scale.z)";
   }
   else if(marker.scale.x != 0.0 || marker.scale.y != 0.0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "scale.x and scale.y of TEXT_VIEW_FACING are ignored";
   }
 }
@@ -307,7 +307,7 @@ void checkColor(const visualization_msgs::Marker& marker, std::stringstream& ss)
 {
   if(marker.color.a == 0.0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "marker is fully transparent (color.a is 0.0)";
   }
 }
@@ -316,7 +316,7 @@ void checkPointsArrow(const visualization_msgs::Marker& marker, std::stringstrea
 {
   if(marker.points.size() != 0 && marker.points.size() != 2)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "Number of points for an ARROW marker should be either 0 or 2";
   }
 }
@@ -325,22 +325,22 @@ void checkPointsNotEmpty(const visualization_msgs::Marker& marker, std::stringst
 {
   if(marker.points.empty())
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "points should not be empty for specified marker type" ;
   }
   else if(marker.type == visualization_msgs::Marker::TRIANGLE_LIST && (marker.points.size() % 3) != 0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "number of points should be a multiple of 3 for TRIANGLE_LIST Marker";
   }
   else if(marker.type == visualization_msgs::Marker::LINE_LIST && (marker.points.size() % 2) != 0)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "number of points should be a multiple of 2 for LINE_LIST Marker";
   }
   else if(marker.type == visualization_msgs::Marker::LINE_STRIP && marker.points.size() <= 1)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "at least two points are required for a LINE_STRIP Marker";
   }
 }
@@ -349,7 +349,7 @@ void checkPointsEmpty(const visualization_msgs::Marker& marker, std::stringstrea
 {
   if(!marker.points.empty())
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "points array is ignored by specified marker type";
   }
 }
@@ -363,7 +363,7 @@ void checkColors(const visualization_msgs::Marker& marker, std::stringstream& ss
   }
   if(marker.colors.size() != marker.points.size())
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "number of colors is not equal to number of points or 0";
   }
 }
@@ -372,7 +372,7 @@ void checkColorsEmpty(const visualization_msgs::Marker& marker, std::stringstrea
 {
   if(!marker.colors.empty())
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "colors array is ignored by specified marker type";
   }
 }
@@ -381,12 +381,12 @@ void checkTextNotEmptyOrWhitespace(const visualization_msgs::Marker& marker, std
 {
   if(marker.text.empty())
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "text is empty for TEXT_VIEW_FACING type marker";
   }
   else if(marker.text.find_first_not_of(" \t\n\v\f\r") == std::string::npos)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "text of TEXT_VIEW_FACING Marker only consists of whitespaces";
   }
 }
@@ -395,7 +395,7 @@ void checkTextEmpty(const visualization_msgs::Marker& marker, std::stringstream&
 {
   if(!marker.text.empty())
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "text is ignored for specified marker type";
   }
 }
@@ -404,7 +404,7 @@ void checkMesh(const visualization_msgs::Marker& marker, std::stringstream& ss)
 {
   if(marker.mesh_resource.empty())
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "path to mesh resource is empty for MESH_RESOURCE marker";
   }
 }
@@ -413,21 +413,21 @@ void checkMeshEmpty(const visualization_msgs::Marker& marker, std::stringstream&
 {
   if (!marker.mesh_resource.empty())
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "mesh_resource is ignored for specified marker type";
   }
   if (marker.mesh_use_embedded_materials)
   {
-    addNewlineIfRequired(ss);
+    addCommaAndNewlineIfRequired(ss);
     ss << "using embedded materials is not supported for markers other than MESH_RESOURCE";
   }
 }
 
-void addNewlineIfRequired(std::stringstream& ss)
+void addCommaAndNewlineIfRequired(std::stringstream& ss)
 {
   if (ss.tellp() != 0) // check if string is not empty
   {
-    ss << "\n";
+    ss << ",\n";
   }
 }
 
--- a/src/rviz/default_plugin/marker_utils.cpp
+++ b/src/rviz/default_plugin/marker_utils.cpp
@@ -32,10 +32,18 @@
 #include "rviz/default_plugin/markers/arrow_marker.h"
 #include "rviz/default_plugin/markers/line_list_marker.h"
 #include "rviz/default_plugin/markers/line_strip_marker.h"
+#include "rviz/default_plugin/marker_display.h"
 #include "rviz/default_plugin/markers/points_marker.h"
 #include "rviz/default_plugin/markers/text_view_facing_marker.h"
 #include "rviz/default_plugin/markers/mesh_resource_marker.h"
 #include "rviz/default_plugin/markers/triangle_list_marker.h"
+#include "rviz/display_context.h"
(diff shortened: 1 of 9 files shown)


v4hn:
@wjwwood any chances to see this merged in *any* (ROS1) branch? Any changes required?
Is there any other maintainer by now that one could ping?

This is very helpful when working with students when they don't know what they are doing really.

wjwwood:
I'm sorry, but I don't have time to merge any features right now, or release them. I will say that if this changes API or ABI, I'll only take it for melodic or later, but I haven't checked this pr at all yet.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,289 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+    break;
+
+  case visualization_msgs::Marker::MESH_RESOURCE:
+    ss << checkQuaternion(marker);
+    ss << checkColor(marker);
+    ss << checkScale(marker);
+    ss << checkPointsEmpty(marker);
+    ss << checkColorsEmpty(marker);
+    ss << checkTextEmpty(marker);
+
```
rhaschke: Remove empty line
```suggestion
```

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,289 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
   }
 }
 
+bool checkMarkerMsg(const visualization_msgs::Marker& marker, MarkerDisplay* owner)
+{
+  std::stringstream ss;
```
rhaschke: Move this below the fast-track return in line 92.
Allocating memory for `stringstream` is costly and should only done when needed.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,289 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+bool checkMarkerArrayMsg(const visualization_msgs::MarkerArray& array, MarkerDisplay* owner)
+{
+  std::set<MarkerID> marker_set;
+
+  for(int i = 0; i<array.markers.size(); i++)
+  {
+    if(i != 0 && array.markers[i].action == visualization_msgs::Marker::DELETEALL)
+    {
+      std::stringstream warning;
+      warning << "found a DELETEALL at index " << i << ", previous markers in the MarkerArray will never show";
```
rhaschke: You should probably check that previous marker msgs actually try to show a marker - just set a boolean flag if so.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,289 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+    return false;
+  }
+
+  owner->deleteMarkerStatus(MarkerID(marker.ns, marker.id));
+  return true;
+}
+
+bool checkMarkerArrayMsg(const visualization_msgs::MarkerArray& array, MarkerDisplay* owner)
+{
+  std::set<MarkerID> marker_set;
```
rhaschke: Please use a **sorted** `std::vector` here, reserving `array.markers.size()` elements. [`std::set` is rather costly](http://lafstern.org/matt/col1.pdf) and not needed here. You can easily check for duplicate items using `std::lower_bound` as indicated in the reference.

src/rviz/default_plugin/markers/text_view_facing_marker.cpp
```diff
@@ -57,6 +57,8 @@ void TextViewFacingMarker::onNewMessage(const MarkerConstPtr& old_message, const
 {
   ROS_ASSERT(new_message->type == visualization_msgs::Marker::TEXT_VIEW_FACING);
 
+  if(new_message->text.empty()|| new_message->text.find_first_not_of(" \t\n\v\f\r") == std::string::npos)
```
rhaschke: I cannot "hide" the text anymore by simply displaying an empty string?

v4hn: > I cannot "hide" the text anymore by simply displaying an empty string?

You never could. With an empty text field (at least after something has been published there), the plugin segfaults.
With whitespace-only strings, OGRE triggers an exception:

> 1550484180.013238148: ros.rviz: OGRE EXCEPTION(7:InternalErrorException): Vertex Buffer: Out of memory in GLHardwareVertexBuffer::lock at /usr/src/ogre-v1-9-0/RenderSystems/GL/src/OgreGLHardwareVertexBuffer.cpp (line 124)
terminate called after throwing an instance of 'Ogre::InternalErrorException'

That being said, it remains underspecified semantics whether empty-string should be allowed to disable a marker or not.
We went ahead and forbade it because there are already a number of other pitfalls that can make you publish markers that never show in Rviz.
If you prefer the alternative semantics to allow empty-string to serve as an alternative to `color.a=0` or `action=DELETE`, this request can be changed accordingly.

rhaschke: OK. Could you please address the other comments?

v4hn:
@jgueldenstein is currently finalizing his thesis (due in two weeks).
He promised me to take care of this afterwards.

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Where possible, `std:string` was replaced with `char*` in d02db0b

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Done in eaa5419

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Done in 1f5f8fd

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Set has been replaced as suggested by a `std::vector` in 08c8351

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Formatting has been done in 63e1d6a

jgueldenstein: I just talked to @v4hn again. I misunderstood your comment and will fix it by passing the `std::stringstream` to the checker functions.

jgueldenstein: In 86c4ffa the return values of the checker functions are replaced by a reference to the `std::stringstream`. Furthermore, `tellp` is used to check if a comma has to be added before adding something to the error message and to check if there is an error at all. 

db705e2 just removes an unnecessary check for adding a comma.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+  default:
+    ss << "Unknown marker type: " << marker.type ;
+  }
+
+  if(ss.tellp() != 0) //stringstream is not empty
+  {
+    std::string warning = ss.str();
+
+    owner->setMarkerStatus(MarkerID(marker.ns, marker.id), StatusProperty::Warn, warning);
+    ROS_WARN("Marker '%s/%d': %s", marker.ns.c_str(), marker.id, warning.c_str());
```
rhaschke: All messages have now the same severity level of "warning". Previously, e.g. an unknown marker type would raise an error, quaternion issues would raise a warning only _once_.
I suggest to add another reference argument to your check functions and pass the desired warning level.
Initialized to Ok, the checkers can further increase the warning level (but not decrease).

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+  return true;
+}
+
+bool checkMarkerArrayMsg(const visualization_msgs::MarkerArray& array, MarkerDisplay* owner)
+{
+  std::vector<MarkerID> marker_ids;
+
+  bool add_marker_in_array = false;
+
+  for(int i = 0; i<array.markers.size(); i++)
```
rhaschke:
```suggestion
  for(int i = 0; i < array.markers.size(); i++)
```

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+    if(array.markers[i].action == visualization_msgs::Marker::ADD)
+    {
+      add_marker_in_array = true;
+    }
+
+    if(add_marker_in_array && i != 0 && array.markers[i].action == visualization_msgs::Marker::DELETEALL)
+    {
+      std::stringstream warning;
+      warning << "found a DELETEALL at index " << i << ", previous markers in the MarkerArray will never show";
+      ROS_WARN("MarkerArray: %s", warning.str().c_str());
```
rhaschke: Calling warning.str() twice in a row here, will allocate memory for the string twice.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+      return false;
+    }
+    MarkerID current_id(array.markers[i].ns, array.markers[i].id);
+    std::vector<MarkerID>::iterator search = std::lower_bound(marker_ids.begin(), marker_ids.end(), current_id);
+    if (search != marker_ids.end())
+    {
+      std::stringstream ss;
+      ss << "found '" <<  array.markers[i].ns.c_str() << "/" << array.markers[i].id << "' multiple times";
+      ROS_WARN("MarkerArray: %s", ss.str().c_str());
+      owner->setStatusStd(StatusProperty::Warn, "marker_array", ss.str());
```
rhaschke: Same here.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+      return false;
```
rhaschke: Why would you bail out here? You should continue to check the remaining markers.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+
+void checkScale(const visualization_msgs::Marker& marker, std::stringstream& ss)
+{
+  // for ARROW markers, scale.z is the optional head length
+  if(marker.type == visualization_msgs::Marker::ARROW && marker.scale.x != 0.0 && marker.scale.y != 0.0)
+    return;
+
+  if(marker.scale.x == 0.0  || marker.scale.y == 0.0  || marker.scale.z == 0.0)
+  {
+    if(marker.type == visualization_msgs::Marker::TRIANGLE_LIST)
```
rhaschke: Why mentioning `TRIANGLE_LIST` here separately?

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+{
+  if(marker.colors.size() == 0)
+  {
+    checkColor(marker, ss);
+  }
+
+  if(marker.colors.size() != marker.points.size())
+  {
+    addCommaIfRequired(ss);
+    ss << "number of colors is not equal to number of points or 0";
```
rhaschke: The error message doesn't match the checks: marker.colors might be empty here, while marker.points are not.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+    addCommaIfRequired(ss);
+    ss << "using embedded materials is not supported for markers other than MESH_RESOURCE";
+  }
+}
+
+void addCommaIfRequired(std::stringstream& ss)
+{
+  if (ss.tellp() != 0) // check if string is not empty
+  {
+    ss << ", ";
```
rhaschke: What about using newline instead of comma?
The list of warnings might get quite long...

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -76,4 +84,346 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext*
...
+      ss << "found '" <<  array.markers[i].ns.c_str() << "/" << array.markers[i].id << "' multiple times";
+      ROS_WARN("MarkerArray: %s", ss.str().c_str());
+      owner->setStatusStd(StatusProperty::Warn, "marker_array", ss.str());
+      return false;
+    }
+    else
+    {
+      marker_ids.insert(search, current_id);
+    }
+  }
```
rhaschke: Shouldn't there be a `checkMarkerMsg(array.markers[i])` here as well to check the individual markers?

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: It seems not obvious that triangle lists should be scaled in my opinion. The sizes of the triangles should be already specified by the positions of the vertices. While the scaling is a nice convenience feature, forgetting to set the scale might lead to confusion.

The other markers are checked for their specific scale semantics as well.

What do you suggest? Comment the code about the reason for this check? Or a different option?

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Single allocation is done in b8ef09d

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Single allocation is done in b8ef09d

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: I think that is a good idea since the error messages might get quite long and the different errors are more clearly separated by a new line. Implemented in c028953

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Previously there was a return before this check so it was always true that the colors array is not empty. This has been re-added in c414eac

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Individual markers in a MarkerArray are already checked because they are added to the tf::MessageFilter which calls the incomingMarker function. The callback is registered [here](https://github.com/TAMS-Group/rviz/blob/c028953511c959d88d266fc0f971e6decc1bc09b/src/rviz/default_plugin/marker_display.cpp#L81)

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Solved in 428f564

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Firstly, do you think that the warnings should only be printed once as it was before?

Secondly, I agree that for some messages ROS_INFO would be enough. The others should be ROS_WARN when following the definition [here](http://wiki.ros.org/Verbosity%20Levels). I do not see a need for ROS_ERROR. What is your opinion on this?

v4hn:
> do you think that the warnings should only be printed once as it was before?

ROS_*_ONCE is not an option here because the notification can be triggered by any error type.
To properly reduce warnings in a stream of repeating messages, the display would have to maintain a memory of which namespace / error combination was observed before.
The user (or at least I) would expect at least one notification per "type" of marker, not just one per display.
I believe the required overhead outweighs the benefits here.

> for some messages ROS_INFO would be enough. The others should be ROS_WARN when following the definition [here](http://wiki.ros.org/Verbosity%20Levels). I do not see a need for ROS_ERROR. What is your opinion on this?

Firstly, this is not only about `ROS_*`, rviz also defines their own status family to display in the GUI (in the line right above), which is similar though.

Secondly, if a vital part of the marker message is filled in incorrectly and the marker cannot be displayed due to that,
this constitutes an actual error, not just a warning.
I have a hard time coming up with a message that could constitute an `Info`-style notification, but it might exist.

src/rviz/default_plugin/marker_utils.cpp
v4hn: The only thing this `if` does is to add `, TRIANGLE_LIST coordinates are scaled` to the notification.
I believe @rhaschke wondered why that should be necessary.
Given your explanation, I would propose to drop the special case.
As a developer, you would still fill in the missing scale values if you see a warning `scale contains 0.0 in x, y or z` and don't see your mesh, wouldn't you?

src/rviz/default_plugin/marker_utils.cpp
v4hn: > do you think that the warnings should only be printed once as it was before?

ROS_*_ONCE is not an option here because the notification can be triggered by any error type.
To properly reduce warnings in a stream of repeating messages, the display would have to maintain a memory of which namespace / error combination was observed before.
The user (or at least I) would expect at least one notification per "type" of marker, not just one per display.
I believe the required overhead outweighs the benefits here.
> for some messages ROS_INFO would be enough. The others should be ROS_WARN when following the definition [here](http://wiki.ros.org/Verbosity%20Levels). I do not see a need for ROS_ERROR. What is your opinion on this?

Firstly, this is not only about `ROS_*`, rviz also defines their own status family to display in the GUI (in the line right above), which is similar though.

Secondly, if a vital part of the marker message is filled in incorrectly and the marker cannot be displayed due to that,
this constitutes an actual error, not just a warning.
I have a hard time coming up with a message that could constitute an `Info`-style notification, but it might exist.

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: In addition to a newline, a comma is nice to have for the rviz status display since newlines are not displayed there. I did this in 3a75c5b

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: Please excuse the delay on this, I finally got to implementing this. All issues which cause the marker not to be displayed are now an error. All issues that are related to additional information in the marker which is not used in this type of marker are a warning. 3aadbff

src/rviz/default_plugin/marker_utils.cpp
jgueldenstein: I agree. I dropped the special case in cf0d693.

jgueldenstein:
Thank you very much for your patience, time and your review, I checked your commits and I think they do some nice cleanup and consistency. 
f7dea2b is a nice feature to have to immediately see all errors in a marker array. Thanks for that.
I do not quite understand when an undefined owner would occur but I think checking for that is good.

v4hn:
Thank you for all the time you spent on this @rhaschke !

PR #1280, opened 2018-08-14
Seemingly minor fix to render RobotModel transparency properly

robustify:
Hello! I was experimenting with including UV mapped textures with transparency into a URDF model and encountered some strange behavior in how Rviz rendered it. Below is a screenshot of the latest Melodic release of Rviz rendering the model with the alpha of the entire robot set to 1:

original_opaque

The logo on the hood of the car is a .png texture with transparent background, but the transparent part is being rendered as black. However, if I set the alpha of just the hood logo link to 0.99, it looks fine:

original_link_99

If I set the entire RobotModel alpha to 0.99 though, the rest of the model gets completely butchered:

original_99

After observing this, I started going through the Rviz source and found the section of code that I modified in this PR. The alpha setting in the RobotModel display is compared to a seemingly arbitrary value of 0.9998, and changes the material rendering properties depending on which side of that value the setting is on.

It seems like the best of both worlds is realized if the `setSceneBlending` call from the `if` is combined with the `setDepthWriteEnabled` call from the `else` in all cases; the transparency in the DAE file texture as well as the overall transparency applied to the model work at the same time. This is the result after building this PR with the entire RobotModel set to alpha = 1:

fixed_opaque

... and alpha = 0.5:

fixed_transparent

I'm not sure if this will break anything in someone else's use case, or if this is some quirk with my graphics hardware. I don't think it's hardware related though, because I've gotten the same results on several computers with different graphics cards. After building this PR, it has worked well in all the situations I've tried it in.

If you'd like to reproduce these results, the repository with the model is here: [https://github.com/robustify/audibot](https://github.com/robustify/audibot)


philipp1234:
Hi,

unfortunately, this would probably break more than it would fix. For small edge cases where Alpha is just below 1.0, this might be looking better, but when it comes to those cases for which alpha blending is actually intended (objects which are actually transparent, alpha around 0.5 or even lower), this would usually just make things worse. With two actually transparent objects we would expect to at least somehow be able to see both, but if depth writing is enabled, it can happen that one of the objects disappears completely. 

And I think it's most important that rviz is at least able to somehow convey all relevant information when configured properly, so if the user intentionally sets objects to be transparent (with proper alpha values around 0.5), the user is telling rviz that both objects may be behind one another but both of them are conveying relevant information, so both of those objects should at least somehow be visible (even if the exact blending percentages may be slightly wrong). 

It would of course be nice to have alpha blending with depth writing as an optional parameter, but it should be off by default.

Methods for rendering correct transparency:
- Sorting transparent polygons before rendering (only works correctly without intersections, performance might also be bad if done on the CPU)
- Depth peeling (possibly bad performance in case of many layers, may require many rendering passes)
- Per-pixel linked lists using shaders (performance may be bad on older hardware, might not work at all on very old GPUs)

robustify:
Understood, and thanks for the detailed feedback! I took a closer look with smaller alpha values and observed what you mentioned; occluded components of the example model disappear even though it is being set as "transparent".

I just updated this PR by adding a boolean property for the `RobotModel` display plugin to enable using both alpha blending and depth buffer writing. This property defaults to false, which leaves the behavior of the `updateAlpha()` method in `robot_link.cpp` unchanged.

I think using alpha blending with depth writing enabled can result in preferable performance in certain situations more than just mine, so I hope this option can be useful to the community!

philipp1234:
Ok, nice, sounds good. Someone from my research group asked me about my opinion on this. I haven't tested it and I'm not an official rviz maintainer, but I think it would be useful now and could be merged.

philipp1234:
Oh, by the way, since we've got a thread here about fixing transparency in rviz...

Ogre seems to support Alpha to coverage. That might also be useful. 
https://www.stevestreeting.com/2008/08/23/alpha-to-coverage-support-added/

And different blending modes ("multiply", "add", "screen") would always be order independent if all objects are using the same mode. For example, when rendering a transparent point cloud or marker array, if all elements are using the same one of those other blending modes, the order in which they're rendered wouldn't matter (when rendering smoke or fire in games, they're often doing it this way). So it might also be good to have those other blending modes in rviz.

PR #1282, opened 2018-08-22
Replace display error assert with warning, fallback and error 

josephduchesne:
Replace assert with warning, fallback and error so that when in release mode, librviz does not segfault with a misconfigured DISPLAY env.

In our use of librviz, RenderSystem::setupDummyWindowId() was segfaulting since display was not set correctly. This will provide warnings, fallback and a useful error in such cases.

### DIFF
--- a/src/rviz/ogre_helpers/render_system.cpp
+++ b/src/rviz/ogre_helpers/render_system.cpp
@@ -142,7 +142,7 @@ void RenderSystem::setupDummyWindowId()
 
     if (display == NULL) {
       ROS_FATAL("Can't open default or :0 display. Try setting DISPLAY environment variable.");
-      std::runtime_error( "Can't open default or :0 display!\n");
+      throw std::runtime_error("Can't open default or :0 display!\n");
     }
 
   }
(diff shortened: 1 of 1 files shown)


src/rviz/ogre_helpers/render_system.cpp
```diff
@@ -134,7 +134,18 @@ void RenderSystem::setupDummyWindowId()
...
-  assert( display );
+
+  if (display == NULL) {
+
+    ROS_WARN("$DISPLAY is invalid, falling back on default :0");
+    display = XOpenDisplay(":0");
+
+    if (display == NULL) {
+      ROS_FATAL("Can't open default or :0 display. Try setting DISPLAY environment variable.");
+      std::runtime_error( "Can't open default or :0 display!\n");
```
wjwwood: This needs to throw right?

wjwwood: Also, it has an extra whitespace between the `(` and the `"`.

josephduchesne:
This PR should be good now.

PR #1288, opened 2018-09-05
Poll at 1Hz for robot description until it is found.

asymingt:
This commit safeguards against the case where a robot description only becomes available as a ROS parameter **after** rviz loads. A timer is used to poll for the robot description at 1Hz until it is found. As soon as the robot description is found, then the timer is stopped.


rhaschke:
Implemented via d171d8239.

PR #1292, opened 2018-09-19
Fix rendering order for movable text

roehling:
I recently had to render transparent textures for a map visualization
plugin. Traditional wisdom says that solid objects are rendered first,
and everything transparent afterwards. However, choosing a later Ogre
rendering queue overwrites all movable text elements, since they disable
depth writing for some reason.

This PR therefore moves the movable text to the latest possible
rendering queue. For always-on-top rendering, the depth check is
disabled. I decided to leave depth buffer writing always on, although it
won't make much of a difference either way.

**Edit:** I forgot about overlapping text elements, so I re-disabled depth buffer writing.

### DIFF
--- a/src/rviz/ogre_helpers/movable_text.cpp
+++ b/src/rviz/ogre_helpers/movable_text.cpp
@@ -129,7 +129,7 @@ void MovableText::setFontName(const String &fontName)
 
     mpMaterial->setDepthCheckEnabled(!mOnTop);
     mpMaterial->setDepthBias(1.0, 1.0);
-    mpMaterial->setDepthWriteEnabled(true);
+    mpMaterial->setDepthWriteEnabled(false);
     mpMaterial->setLightingEnabled(false);
     mNeedUpdate = true;
   }
--- a/src/rviz/ogre_helpers/movable_text.cpp
+++ b/src/rviz/ogre_helpers/movable_text.cpp
@@ -88,6 +88,7 @@ MovableText::MovableText(const String &caption, const String &fontName, Real cha
   mRenderOp.vertexData = NULL;
   this->setFontName(mFontName);
   this->_setupGeometry();
+  this->setRenderQueueGroup(Ogre::RENDER_QUEUE_OVERLAY);
 }
 
 MovableText::~MovableText()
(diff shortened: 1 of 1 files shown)


dhood:
thanks for the PR @roehling . Is there a reproducible example you could provide to help with testing?

rhaschke:
> I forgot about overlapping text elements, so I re-disabled depth buffer writing.

Could you please explain, why you always disabled depth buffer writing, while before it was depending on the `mOnTop` variable?
I guess moving it to the latest queue, renders text always on top now?

roehling:
There are two distinct mechanisms at work here: the depth buffer check will prevent the text from rendering if something else is occluding the text. If the text is to be rendered always-on-top, the depth buffer must be ignored, and if not, the depth buffer must be checked.

The depth buffer writing would update the depth buffer with the depth of the quad that displays the text. If some other text further in the background is rendered later (which can und will happen, since there is no explicit Z-ordering of the text labels). The transparent parts of the first label (technically just a solid quad) will occlude the second, which looks weird.

Actually, this PR only solves halve of my problem, since I get a similar rendering error with transparent map tiles in my plugin, so the "true" solution would be proper Z-ordering of all transparent elements, which are then rendered together in a queue after all solids.

rhaschke:
Thanks for the clarification.

> the "true" solution would be proper Z-ordering of all transparent elements

I agree that this would be nice, but I think we also agree that this is out of scope for rviz as a generic, real-time rendering tool. Z-ordering should consider pixel level actually because parts of two rendered items might have different depths. Consider for example two text items which are crossing each other.

rhaschke:
Dear @roehling, I tried to reproduce the issue, but I couldn't. Actually, I couldn't see any rendering difference between the released version and your PR version. Could you, please, provide a code snipped, e.g. a MarkerArray yaml to reproduce the issue, as well as an image illustrating the problem?

Using marker_test from rviz' test folder, I get the following renderings, which look good to me:
released version:
old

new version:
new

rhaschke:
@roehling Could you please comment on https://github.com/ros-visualization/rviz/pull/1292#issuecomment-471154972?

roehling:
As I said in the original post, it's a very specific setting (i.e. using a non-standard rendering queue to add objects with transparent textures after the bulk of rviz rendering took place) where this patch would improve rendering. Ideally, there should not be a single instance where rendering changes for any of the default plugins.

I can probably come up with minimal example where this patch helps eventually, but I'm a bit starved for time right now. 

Arguably, this whole PR is more of a wishlist bug than an actual pressing matter. I think it would be cleaner to move overlay texts to the overlay queue, but since no one besides me seems to be affected, maybe it's not worth the effort after all.

PR #1294, opened 2018-10-04
fix material handling

### LINKED ISSUE #1222: meshes of RobotLinks are not cached and reused?
When enabling the robot train in MoveIt's MotionPlanning plugin, I observe a dramatic increase of memory usage that is linear with the number of steps in the trail. However, all these visualizations use the same meshes and they should be reused. Obviously they are not. In my case, with a complex robot model, my machine memory of 12GByte is exhausted already with >650 trail steps.

rhaschke:
This attempts to resolve #1222. As noted in https://github.com/ros-visualization/rviz/issues/1222#issuecomment-426464344, I was able to track down the "memory leak" to material handling of robot links. Each time, a new `rviz::RobotLink` was created, the corresponding link material(s) were registered again with the `Ogre::MaterialManager` and never released until rviz finished.

This PR addresses the issue by essentially rewriting material handling of robot links: Instead of registering materials with the `MaterialManager`, they are directly created using `new Ogre::Material()`. This avoids both the unique naming problem, and the memory leak. However, the [Ogre documentation](https://www.ogre3d.org/docs/api/1.9/class_ogre_1_1_material.html#ae3c92b81e6c247286c2304f320a368ef) explicitly suggests to use `MaterialManager::create()` instead of using the Material constructor. You need to decide whether you accept this.
If not, it is important to also remove the material(s) from MaterialManager in the RobotLink destructor, e.g. by calling `MaterialManager::remove(material->getHandle())`. However, I consider this overkill, because the material is explicitly not reused from the MaterialManager.

I also reworked #1079: Instead of searching the corresponding material by the passed name, I directly pass the `urdf::MaterialSharedPtr`. This ensures that the correct material is used, even when two different materials are identically named.


src/rviz/robot/robot_link.cpp
```diff
@@ -620,49 +605,22 @@ void RobotLink::createEntityForGeometryElement(const urdf::LinkConstSharedPtr& l
...
       const std::string& material_name = sub->getMaterialName();
 
       if (material_name == "BaseWhite" || material_name == "BaseWhiteNoLighting")
       {
-        sub->setMaterialName(default_material_name_);
+        sub->setMaterial(default_material_);
       }
-      else
-      {
-        // Need to clone here due to how selection works.  Once selection id is done per object and not per material,
```
rhaschke: @wjwwood I'm not sure whether this code is already obsolete, because selection handling was changed.
Anyway, removing this code, I didn't observed any issues in selecting links. However, probably you better know, what to test...

wxmerkt:
Ping - is this PR ready for merging or what is blocking it? We'd really love to see this fix make its way into rviz :)

wxmerkt:
Hi @rhaschke,
I am extensively testing this bug fix and it indeed fixed the memory leak - CPU is still at 100% when rendering though while only achieving ~17fps. 

However, when I sent invalid data (from a failed optimisation) rviz is now prone to crash with:
```
rviz: /build/ogre-1.9-B6QkmW/ogre-1.9-1.9.0+dfsg1/OgreMain/include/OgreAxisAlignedBox.h:252: void Ogre::AxisAlignedBox::setExtents(const Ogre::Vector3&, const Ogre::Vector3&): Assertion `(min.x <= max.x && min.y <= max.y && min.z <= max.z) && "The minimum corner of the box must be less than or equal to maximum corner"' failed.
[1]    12038 abort (core dumped)  rosrun rviz rviz
```
This does not happen with the same erroneous data using rviz from binaries.

Best,
Wolfgang

PR #1295, opened 2018-10-04
fix gcc warnings

rhaschke:
This PR fixes several compiler warnings cluttering my console...


src/rviz/default_plugin/map_display.cpp
```diff
@@ -704,7 +701,7 @@ void MapDisplay::showMap()
   }
 
   bool map_status_set = false;
-  if( width * height != current_map_.data.size() )
+  if( width * height != (int)current_map_.data.size() )
```
chapulina: Nitpick: you could avoid the C-style cast and use `static_cast<int>`

PR #1300, opened 2018-10-11
WIP: Adding a panel now loads it properties

### LINKED ISSUE #1172: load member of custom Dock-able panel is only called once when creating/deleting the panel multiple times
Minimal example package:
https://gitlab.com/InstitutMaupertuis/rviz_plugin_load_bug

I expect the `load` member of my RViz panels to be called each time the panel is created, it is not the case.

How to reproduce:
- Start RViz
- Add a custom panel with a `load(const rviz::Config& config)` member
- Delete the panel
- Add the same panel again

The `load` member will not be called after the first creation, resulting in a partially initialized panel.
This is problematic for me because I use the `load` member to make sure service connections are done etc...

I'm not sure if it is a bug or an intended behaviour.

VictorLamoine:
Fixes #1172 

`openNewPanelDialog` deals with newly added panels, it now calls the `load` member of the panel to load it's properties.

To do that I needed to have access to the RViz config file so I added a new member to the class holding the data structure. `loadDisplayConfig` fills this data member so that it is available later.

Fixes https://gitlab.com/InstitutMaupertuis/ros_additive_manufacturing/issues/127

### DIFF
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -622,6 +622,11 @@ void VisualizationFrame::openNewPanelDialog()
     if ( dock )
     {
       connect( dock, SIGNAL( dockLocationChanged( Qt::DockWidgetArea )), this, SLOT( onDockPanelChange() ) );
+      Panel* panel = qobject_cast<Panel*>( dock->widget() );
+      if( panel )
+      {
+        panel->load( config_ );
+      }
     }
   }
   manager_->startUpdate();
@@ -726,11 +731,10 @@ void VisualizationFrame::loadDisplayConfig( const QString& qpath )
   }
 
   YamlConfigReader reader;
-  Config config;
-  reader.readFile( config, QString::fromStdString( actual_load_path ));
+  reader.readFile( config_, QString::fromStdString( actual_load_path ));
   if( !reader.error() )
   {
-    load( config );
+    load( config_ );
   }
 
   markRecentConfig( path );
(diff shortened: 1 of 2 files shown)


VictorLamoine:
You can easily test this PR with this package: https://gitlab.com/InstitutMaupertuis/rviz_plugin_load_bug

VictorLamoine:
Let's say your panel is configurable and subscribes to topics to display their values (https://gitlab.com/InstitutMaupertuis/topics_rviz_plugin).

- The topics to subscribe to are stored in the RViz config so that you don't have to re-configure the panel each time you launch RViz.
- When you delete and add the panel again you don't expect your configuration to be lost.

With this configuration I have to subscribe to the topics after the `load` function has loaded everything that needs to be displayed (eg: https://gitlab.com/InstitutMaupertuis/topics_rviz_plugin/blob/melodic/src/display_topics.cpp#L72-73).

I understand that in the case the configuration is not right then the only option is to edit the RViz config file by hand which is not nice.

If my explanation is not clear please ask me for more details
Any ideas on how to fix this issue?

VictorLamoine:
After thinking about it I think @rhaschke is right, it will not possible to add a "default" value panel anymore and that is a problem.

Closing as I don't think it is something that we want to do :)

rhaschke:
> When you delete and add the panel again you don't expect your configuration to be lost.

Here I would strongly disagree. When I delete a panel, I am fine with loosing its config.
If you don't want to loose the config, why would you delete the panel in first place?

PR #1307, opened 2018-10-25
feat: Configurable tool button style kinetic

reinzor:
When using a lot of tools, it's annoying that not all tools fit in the default toolbar. This PR adds an additional tool menu (next to the remove tool menu) for configuring the tool button style: http://doc.qt.io/archives/qt-4.8/qt.html#ToolButtonStyle-enum

### DIFF
--- a/src/rviz/visualization_frame.h
+++ b/src/rviz/visualization_frame.h
@@ -353,7 +353,8 @@ protected:
   };
   QList<PanelRecord> custom_panels_;
 
-  QAction* toolbar_button_separator_;
+  //! @todo Rename to toolbar_button_separator_ in Noetic
+  QAction* add_tool_action_;
   QMenu* remove_tool_menu_;
 
   bool initialized_;
--- a/src/rviz/visualization_frame.h
+++ b/src/rviz/visualization_frame.h
@@ -258,6 +258,7 @@ protected:
 
   void initMenus();
 
+  /** @brief Sets up the top toolbar with QToolbuttions for adding/deleting tools and modifiying the tool view **/
   void initToolbars();
 
   /** @brief Check for unsaved changes, prompt to save config, etc.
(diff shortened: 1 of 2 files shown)


src/rviz/visualization_frame.h
```diff
@@ -276,6 +279,9 @@ protected Q_SLOTS:
   /** @brief Loads custom panels from the given config node. */
   void loadPanels( const Config& config );
 
+  void configureToolbars( const Config& config );
+  void saveToolbars( Config config );
```
chapulina: Mind adding some documentation to these new functions?

reinzor:
Thanks for the review, I updated the PR in order to address your first two points. I don't know how to address the last point.

Now I clearly made a separation between the "actions" (added dynamically via the toolbar_manager) and the static tool buttons. If we want the user to also control the visibility of the static buttons, an additional menu should be added but I think this is beyond the scope of this PR.

src/rviz/visualization_frame.h
```diff
@@ -343,7 +353,7 @@ protected Q_SLOTS:
   };
   QList<PanelRecord> custom_panels_;
 
-  QAction* add_tool_action_;
+  QAction* toolbar_button_separator_;
```
chapulina: This header is installed, so changing this variable name will break ABI. Mind keeping the old name and add a note that the variable should be renamed on Noetic?

reinzor:
Thanks for your comment, I updated both PR's to maintain the ABI

chapulina:
@ros-pull-request-builder please test this

wjwwood:
@ros-pull-request-builder please test this

wjwwood:
@dirk-thomas CI doesn't seem to be getting trigger now (previous commits on this pr were tested), what's the best way to debug it?

dirk-thomas:
You need to use the default phrase of the Jenkins plugin: see https://plugins.jenkins.io/ghprb

wjwwood:
@dirk-thomas it looks like it is rerunning CI on the second latest commit but no the latest.

dirk-thomas:
> it looks like it is rerunning CI on the second latest commit but no the latest.

No, the referenced checks in the GitHub UI are pointing to a build of the latest commit.

wjwwood:
Weird, it didn't look like that before. Ok, well I dunno I guess it's working fine, thanks!

chapulina:
Thanks @wjwwood and @dirk-thomas !

PR #1309, opened 2018-10-25
feat: Configurable tool button style melodic

reinzor:
Same as #1307 for melodic

### DIFF
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -122,7 +122,7 @@ VisualizationFrame::VisualizationFrame( QWidget* parent )
   , splash_( NULL )
   , toolbar_actions_( NULL )
   , show_choose_new_master_option_( false )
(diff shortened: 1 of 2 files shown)


reinzor:
Can this one be merged as well?

reinzor:
ping

reinzor:
Can this one be merged as well? It is already merged in the kinetic branch ..

wjwwood:
Normally we do forward porting ourselves, unless there's significant changes required to forward port (normally we request pr's to the latest release branch instead and do backports, but didn't in this case).

We do forward and back porting when we release, so this will get addressed then if not before. I cannot say right now when we'll do the next release though.

In the meantime, maybe @chapulina has time to review and merge this one too since she did the original one as well.

chapulina:
This is on my radar together with lots of other things. I'll try to get to it next week. :crossed_fingers:

reinzor:
Okay, thanks for the update :)

PR #1317, opened 2018-12-09
The data size is missing in Collada loader

mutr:
I found that my robot does't show propelly due to XML parse error in Collada loader. The data size is missing after converting from array to cstring, this led to garbage at the end of XML structure. This patch fixes this error. Also I added error message.


mutr:
guys, will you merge it to upstream? This is really annoying to patch and rebuild rviz on every machine.
Without this fix my robot looks like this:

rhaschke:
Duplicate of #1341. Considered parser warning in 6a769aa7e1c7703c353768e46f182a88e1fd0b87.

PR #1319, opened 2019-01-02
Improve color support for themes

VictorLamoine:
This improves readability when using dark or other themes in RViz.

# Before
## White theme
before_white

## Dark theme
before_dark

There are some other colors that might be improved but this fixes the main issues I have seen.

---

Does someone know where the color is set for the status tree and similar entries?
status tree

### DIFF
--- a/src/rviz/properties/property_tree_with_help.cpp
+++ b/src/rviz/properties/property_tree_with_help.cpp
@@ -68,7 +68,7 @@ void PropertyTreeWithHelp::showHelpForProperty( const Property* property )
   {
     QString body_text = property->getDescription();
     QString heading = property->getName();
-    QString html = "<html><body bgcolor=\"#EFEBE7\"><strong>" + heading + "</strong><br>" + body_text + "</body></html>";
+    QString html = "<html><body><strong>" + heading + "</strong><br>" + body_text + "</body></html>";
     help_->setHtml( html );
   }
   else
--- a/help/help.html
+++ b/help/help.html
@@ -8,8 +8,8 @@
 <html>
 <head>
 <style type="text/css">
-.odd { background-color: #e0e0e0 }
-.even { background-color: #f0f0f0 }
+.odd { background-color: transparent }
+.even { background-color: transparent }
 .key { font-weight: bold; color: #04597f }
 p { margin-left: 10px }
 </style>
--- a/src/rviz/display.cpp
+++ b/src/rviz/display.cpp
@@ -108,16 +108,6 @@ QVariant Display::getViewData( int column, int role ) const
 {
   switch( role )
   {
-  case Qt::BackgroundRole:
-  {
-    /*
-    QLinearGradient q( 0,0, 0,5 );
-    q.setColorAt( 0.0, QColor(230,230,230) );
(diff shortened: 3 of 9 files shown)


VictorLamoine:
Firendly :pig:

src/rviz/properties/property_tree_with_help.cpp
```diff
@@ -68,7 +68,7 @@ void PropertyTreeWithHelp::showHelpForProperty( const Property* property )
   {
     QString body_text = property->getDescription();
     QString heading = property->getName();
-    QString html = "<html><body bgcolor=\"#EFEBE7\"><strong>" + heading + "</strong><br>" + body_text + "</body></html>";
+    QString html = "<html><strong>" + heading + "</strong><br>" + body_text + "</body></html>";
```
rhaschke:
```suggestion
    QString html = "<html><body><strong>" + heading + "</strong><br>" + body_text + "</body></html>";
```

VictorLamoine:
I hard a hard time finding this color because it was not initialized.
This pull request is ready on my side.

# After

## White theme
screenshot_20190218_132219

## Black theme
screenshot_20190218_132311

# Other color problems
In the help panel:
screenshot_20190218_132630

This panel uses HTML so it's not easy to fix

VictorLamoine:
For the help panel we can force the color of the text in the tabs like this:
screenshot_20190218_133358

```html
  <tr class="odd">
    <td class="key">M</td>
    <td><font color="black">Switch to the Move-Camera tool.</font></td>
  </tr>
  <tr class="even">
```

It's not very elegant but it works.

VictorLamoine:
@rhaschke can you have an other look?

rhaschke:
I will have another look, but probably not earlier than March 9th.

rhaschke:
Dear @VictorLamoine,
I further tweaked the coloring and cleaned up the commit history in my own fork:
https://github.com/rhaschke/rviz/tree/colors_dark_theme

Could you please git-reset your branch to this one and verify that everything looks fine for you as well?
As you see from the images below, the panel title bars are not correctly colored on my side - probably due to a bug in the `Adwaita-dark` theme that I'm using for testing. On your side it looked good.

bright:
bright

dark:
dark

VictorLamoine:
Most likely you adwaita theme is incomplete!

RViz with Kubuntu breeze theme look nice:
screenshot_20190307_163336
screenshot_20190307_163628

PeterMitrano:
how do I use this?

rhaschke:
Just change your global desktop theme to a dark one.

PeterMitrano:
Hm.. I think I've done that. For instance, `nautilus` and `jstest-gtk` is in dark theme. But rviz is not (even after restarting). I am on Melodic on Ubuntu 18.04

rhaschke:
This might have run into a well-known [Qt issue](https://wiki.archlinux.org/index.php/Uniform_look_for_Qt_and_GTK_applications): In order to use gtk desktop themes, you need to install qt5-style-plugins and maybe 
`export QT_QPA_PLATFORMTHEME=gtk2`

PeterMitrano:
that did the trick -- and setting that variable seems nessecary. Thanks!

PR #1320, opened 2019-01-18
Fix compatibility with boost 1.69.

de-vri-es:
The old signals library has been removed from boost 1.69. Because of that, `find_package(Boost COMPONENTS signals ...)` doesn't work anymore. The newer signals2 library is header only and does not need to be listed as boost component at all.

Since `rviz` seems to still want to be compatible with the old `signals` library, this PR makes the inclusion of `signals` as boost component optional, depending on whether `signals` or `signals2` will be used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,15 +8,6 @@ if (POLICY CMP0054)
   cmake_policy(SET CMP0054 NEW)
 endif()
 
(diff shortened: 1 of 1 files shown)


rhaschke:
Replaced by #1344. I think support for tf1 < 1.11.3, which is ~5 years old, can be dropped.

PR #1321, opened 2019-01-24
Add a shortcut key to the Publish Point tool

DaikiMaekawa:


DaikiMaekawa:
I often use the Publish Point to clean up a Point Cloud data with my private tool. A shortcut key makes me faster to do the same work.
If you think the PR helps others, please merge it.

rhaschke:
The choice of `c` as a short cut for `Publish Point` seems weird.

DaikiMaekawa:
@rhaschke 
OK, then which key do you think the best?
I picked `c` from `click`, and unfortunately `p` is already used for `2D Pose Estimate tool`.

beetleskin:
I like `c`.

DaikiMaekawa:
> Is this automatically merged into melodic-devel?

I guess not. 

@rhaschke Should I open the same PR onto the other distro's branches?

rhaschke:
No need to file a PR for Melodic.

PR #1323, opened 2019-01-27
Fixed joint display hierarchy in EffortDisplay

d-walsh:
Moves all joint information into the **Joints** category.


rhaschke:
Could you please motivate this change?

d-walsh:
@rhaschke The Joints category was unused previously:

screenshot from 2019-02-18 12-07-18

Now it is used:

screenshot from 2019-02-18 12-03-55

I believe it was a bug.

PR #1325, opened 2019-02-08
[Windows][melodic] add missing DLL exports.

seanyen-msft:
Define DLL exports for rviz (http://wiki.ros.org/win_ros/Contributing/Dll%20Exports).

This change is verified by https://aka.ms/ros project.


rhaschke:

seanyen-msft:
Closed it since it is re-sent by #1335.

PR #1326, opened 2019-02-08
[Windows][melodic] RViz sip python binding fix

seanyen-msft:
* Define correct linker usage of rviz_sip_LDFLAGS_OTHER for MSVC.
* Use .pyd as Python module extension for Windows build. (https://docs.python.org/2/faq/windows.html#is-a-pyd-file-the-same-as-a-dll)


rhaschke:

seanyen-msft:
Closed it since it is re-sent in #1336

PR #1327, opened 2019-02-08
[Windows][melodic] Use c++11 chrono to replace gettimeofday.

seanyen-msft:
Use c++11 chrono API to replace gettimeofday for better code portability.


rhaschke:

seanyen-msft:
Closed it since it is re-sent in #1337

PR #1332, opened 2019-02-13
[Windows][melodic] Fix install destination path.

seanyen-msft:
Follow the catkin guide to place output binaries correctly for cross-platform:
- Place shared libraries like: http://docs.ros.org/api/catkin/html/howto/format2/building_libraries.html
```  
install(TARGETS your_library your_other_library
        ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
        LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
        RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
```
 
- Place executables like: http://docs.ros.org/api/catkin/html/howto/format2/building_executables.html
```
install(TARGETS your_node
        RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
```

PR #1333, opened 2019-02-13
[Windows][melodic] Disable XLib code path on Windows

seanyen-msft:
On native Windows environment (not MingGW or Cygwin), it doesn't have XLib, so conditionally compile the code like macOS.


rhaschke:

seanyen-msft:
Closed it since it is moved to #1339

PR #1334, opened 2019-02-13
[Windows][melodic] Use portable TARGET_LINKER_FILE_NAME to get shared lib name.

seanyen-msft:
In CMake, `TARGET_FILE_NAME` returns `.so` on Linux but `.lib` on Windows. However, `.lib` is the extension for imported library, and `.dll` is the expected extension to return (for shared libraries).

Using `TARGET_LINKER_FILE_NAME` makes sure that it returns `.so` and `.dll` for both platforms.


rhaschke:

seanyen-msft:
Closed it since it is moved to #1338

PR #1335, opened 2019-02-13
[Windows][melodic] add missing DLL exports.

seanyen-msft:
Define DLL exports for rviz (http://wiki.ros.org/win_ros/Contributing/Dll%20Exports).

This change is verified by https://aka.ms/ros project.


src/rviz/default_plugin/interactive_markers/interactive_marker.h
```diff
@@ -30,6 +30,18 @@
 #ifndef RVIZ_INTERACTIVE_MARKER_H_
 #define RVIZ_INTERACTIVE_MARKER_H_
 
+// Import/export for windows dll's and visibility for gcc shared libraries.
```
rhaschke: Please move this boiler-plate block into an include file.

src/rviz/rviz_macros.h
```diff
@@ -0,0 +1,45 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef RVIZ_MACROS_H_
+#define RVIZ_MACROS_H_
+
+#include <ros/macros.h>
+
+#ifdef ROS_BUILD_SHARED_LIBS // ros is being built around shared libraries
```
rhaschke: Where do these defines (`ROS_BUILD_SHARED_LIBS` and `rviz_EXPORTS`) come from?
I know that cmake can [auto-generate corresponding include files](https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html). Why don't you use this mechanism?

seanyen-msft: The primary reason is we saw that many ROS projects have been already following this [ROS Wiki](http://wiki.ros.org/win_ros/Contributing/Dll%20Exports) to define the exports header, and figured we are not going to use another idiom to diverge from this existing one for ROS1 community.

`ROS_BUILD_SHARED_LIBS`: This is defined in `catkin`. See https://github.com/ros/catkin/blob/kinetic-devel/cmake/tools/libraries.cmake#L18

`rviz_EXPORTS`: This is defined by CMake, when the module is built as a shared library. See https://cmake.org/cmake/help/v3.0/prop_tgt/DEFINE_SYMBOL.html

src/rviz/default_plugin/interactive_markers/interactive_marker.h
seanyen-msft: Sure thing!

seanyen-msft:
> I suggest to use cmake's functions [`generate_export_header`](https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html).

The primary reason is we saw that many ROS projects have been already following this [ROS Wiki](http://wiki.ros.org/win_ros/Contributing/Dll%20Exports) to define the exports header, and figured we are not going to use another idiom to diverge from this existing one for ROS1 community. What do you think?

rhaschke:
The ROS wiki page was created back in 2013, when cmake didn't yet provided this neat function.
I think, you should update the wiki and let cmake do it's job. It doesn't make sense to write such boiler-plate code manually.

seanyen-msft:
> The ROS wiki page was created back in 2013, when cmake didn't yet provided this neat function.
> I think, you should update the wiki and let cmake do it's job. It doesn't make sense to write such boiler-plate code manually.

Understood. I switched the implementation to make use of CMake built-in generate_export_header.

seanyen-msft:
`catkin_lint` seems not to get along with dynamically generated files?
```
catkin_lint /root/rviz
rviz: src/rviz/CMakeLists.txt(183): error: install() needs missing file '${PROJECT_BUILD_DIR}/src/rviz/rviz/rviz_export.h'
rviz: src/rviz/default_plugin/CMakeLists.txt(87): error: install() needs missing file '${PROJECT_BUILD_DIR}/src/rviz/default_plugin/rviz/default_plugin/rviz_default_plugin_export.h'
```

rhaschke:
> `catkin_lint` seems not to get along with dynamically generated files?

Yes, I know. Please file a bug at https://github.com/fkie/catkin_lint.
There is the possibility to [ignore issues once](http://fkie.github.io/catkin_lint/messages/).

seanyen-msft:
@rhaschke Thanks. `#catkin_lint: ignore_once missing_file` does the trick.

seanyen-msft:
Looking into this error...
```
Errors     << rviz:make /root/ws_moveit/logs/rviz/build.make.000.log
make[3]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
In file included from panel_dock_widget.sip:7:0:
/root/ws_moveit/src/rviz/src/rviz/panel_dock_widget.h:34:10: fatal error: rviz/rviz_export.h: No such file or directory
 #include "rviz/rviz_export.h"
          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

seanyen-msft:
travis CI errors are fixed.

seanyen-msft:
@rhaschke this is ready for another round of review. Let me know if any feedbacks, thanks!

PR #1337, opened 2019-02-13
[Windows][melodic] Use c++11 chrono to replace gettimeofday.

seanyen-msft:
Use c++11 <chrono> API to replace gettimeofday for better code portability.

### DIFF
--- a/src/test/connect_test.cpp
+++ b/src/test/connect_test.cpp
@@ -28,7 +28,7 @@
  */
 
 #include <stdio.h>
-#include <sys/time.h>
+#include <chrono>
 
 // This is a simple speed test to compare suppressing a signal/slot
 // emission by one of two methods:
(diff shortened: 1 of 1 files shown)


src/test/connect_test.cpp
```diff
@@ -44,9 +44,8 @@
 
 double now()
 {
-  struct timeval tv;
-  gettimeofday( &tv, NULL );
-  return double(tv.tv_sec) + double(tv.tv_usec) / 1000000.0;
+  auto now = std::chrono::system_clock::now();
```
rhaschke: Please use `std::chrono::steady_clock` as suggested [here](https://stackoverflow.com/a/18685338/4089003). Directly return the clock value (instead of double) and compute the duration where necessary.

seanyen-msft:
@rhaschke I saw this Travis failure: Is it something safe to ignore?
```
Unknown TEST: abi
Travis script finished with errors.
The command ".moveit_ci/travis.sh" exited with 1.
```

PR #1342, opened 2019-02-14
[Windows][melodic] Add Ogre::PF_R8G8B8 support in selection_manager

seanyen-msft:
## Problem
- Users are not able to interact with interactive markers when running this [tutorial](http://wiki.ros.org/rviz/Tutorials/Interactive%20Markers%3A%20Getting%20Started).
- Meanwhile, `Incompatible pixel format [10]` are printed in RViz verbose mode.

## Analysis & Solution
On Windows, PF_R8G8B8 could be returned by pixel_buffer in `selection_manager`:
https://github.com/ros-visualization/rviz/blob/26ffa641a8e325aa7ced426c543a4433b8077684/src/rviz/selection/selection_manager.cpp#L735

However, `colorToHandle()` doesn't support PF_R8G8B8, so it errors out with `handle=0`. Add the proper case to fix it.

BTW, on Windows, it does honor the pixel format passed at the below line:
https://github.com/ros-visualization/rviz/blob/26ffa641a8e325aa7ced426c543a4433b8077684/src/rviz/selection/selection_manager.cpp#L396

After changing it to a valid & currently supported pixel format (e.g., PF_A8R8G8B8), the issue went away too.

### DIFF
--- a/src/rviz/selection/forwards.h
+++ b/src/rviz/selection/forwards.h
@@ -68,7 +68,7 @@ typedef boost::unordered_map<CollObjectHandle, Picked> M_Picked;
 inline uint32_t colorToHandle(Ogre::PixelFormat fmt, uint32_t col)
 {
   uint32_t handle = 0;
-  if (fmt == Ogre::PF_A8R8G8B8 || fmt == Ogre::PF_X8R8G8B8)
+  if (fmt == Ogre::PF_A8R8G8B8 || fmt == Ogre::PF_X8R8G8B8 || fmt == Ogre::PF_R8G8B8)
   {
     handle = col & 0x00ffffff;
   }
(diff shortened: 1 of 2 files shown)


seanyen-msft:
@ros-pull-request-builder ping

wjwwood:
@ros-pull-request-builder test this please

wjwwood:
Oh, this will not get CI because the target of the pull request is not a branch that our CI is setup to test. If it were targeted at `melodic-devel`, then CI would start. I don't know what you and @rhaschke want to do about that.

seanyen-msft:
> Oh, this will not get CI because the target of the pull request is not a branch that our CI is setup to test. If it were targeted at `melodic-devel`, then CI would start. I don't know what you and @rhaschke want to do about that.

For myself, it would be nice at least I can on-demand trigger the CI (it's okay not to build it after each push). Is that `@ ros-pull-request-builder` does the trick? ;)

wjwwood:
No matter what I do it will not run CI, because it doesn't know what version of the project the target branch is (indigo, kinetic, melodic, etc...). So as long as it is not targeting one of these branches it will not work.

You can do review and merge them into `wip-windows-port`, at which point CI will then run.

PR #1344, opened 2019-02-17
remove support for tf < 1.11.3

### LINKED ISSUE #1343: Remove signals from find_packages()?
Is [this issue](https://github.com/ros/ros_comm/pull/1580) also relevant here?

I'm building on arch linux with cmake 3.13 and boost 1.69 and get the same error. When removing [this line](https://github.com/ros/ros_comm/pull/1580)  from `CMakeLists.txt`, I'm able to build.

rhaschke:
... and thus the need for boost::signals(1), which was removed from boost in version 1.69.
Resolves #1343.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,6 @@ find_package(Boost REQUIRED
   COMPONENTS
   filesystem
   program_options
-  signals
   system
   thread
 )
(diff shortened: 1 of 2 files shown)


jwhendy:
I just built off of your branch successfully. Thanks for the quick fix!

PR #1356, opened 2019-03-18
[Windows][melodic] Fix the run_tests build breaks on Windows.

seanyen:
This change is fixing the run_tests build breaks when `catkin_make run_tests` on Windows. `random` doesn't exist on Windows so switch to C++ `<random>` library for better portability.

### DIFF
--- a/src/test/send_images.cpp
+++ b/src/test/send_images.cpp
@@ -76,7 +76,7 @@ int main( int argc, char **argv )
         {
           int index = (x + y * width) * 3;
           std::uniform_int_distribution<int> uniform(0, RAND_MAX);
-          auto rand = random(seed);
+          auto rand = uniform(random_generator);
           msg.data[ index ] = rand & 0xff;
           index++;
           msg.data[ index ] = (rand >> 8) & 0xff;
--- a/src/test/send_images.cpp
+++ b/src/test/send_images.cpp
@@ -67,7 +67,7 @@ int main( int argc, char **argv )
     msg.step = width * 3;
 
     int count = 0;
-    std::default_random_engine seed;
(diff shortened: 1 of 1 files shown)


src/test/send_images.cpp
```diff
@@ -66,14 +67,16 @@ int main( int argc, char **argv )
     msg.step = width * 3;
 
     int count = 0;
+    std::default_random_engine seed;
```
rhaschke: The engine isn't a seed. It's the random number generator.
```suggestion
    std::default_random_engine random_generator;
```

src/test/send_images.cpp
```diff
@@ -66,14 +67,16 @@ int main( int argc, char **argv )
...
     while( ros::ok() )
     {
       for( int x = 0; x < width; x++ )
       {
         for( int y = 0; y < height; y++ )
         {
           int index = (x + y * width) * 3;
-          long int rand = random();
+          std::uniform_int_distribution<int> random(0, RAND_MAX);
```
rhaschke: This one maps pseudo random numbers into a random numbers of a specific, here uniform, distribution.
```suggestion
          std::uniform_int_distribution<int> uniform(0, RAND_MAX);
```

src/test/send_images.cpp
```diff
@@ -66,14 +67,16 @@ int main( int argc, char **argv )
...
+          auto rand = random(seed);
```
rhaschke:
```suggestion
          auto rand = uniform(random_generator);
```

seanyen:
@rhaschke Thank for the feedbacks! Fixed the naming.

PR #1363, opened 2019-03-29
Enable/Disable hotkeys

rhaschke:
This is a rework of #1243. @AndyZe, please review.

### DIFF
--- a/src/rviz/preferences.h
+++ b/src/rviz/preferences.h
@@ -36,6 +36,7 @@ namespace rviz
 struct Preferences
 {
   bool prompt_save_on_exit = true;
+  bool enable_tool_hotkeys = true;
 };
 
 } //namespace rviz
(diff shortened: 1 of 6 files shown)


rhaschke:
@AndyZe, while I guess this PR is ready for merging, I'm not yet convinced that we really need this feature. As an alternative, you can always disable (i.e. remove) individual tools. I would consider this a much more flexible approach than disabling all tool hotkeys. Did you considered such an option?

AndyZe:
Well, I didn't think of that. That does work, if people are aware of that option.

Is there any harm in merging this?

AndyZe:
I don't have a strong opinion. Like I said, my motivating issue was resolved a long time ago.

AndyZe:
I will say, most well-polished apps don't have single-character hotkeys. It's usually "Ctrl+something"...

rhaschke:
I don't think disabling all hotkeys is the right approach to disable a single specific tool.
We should strive for a more general solution, allowing to define custom hotkeys.

Issue #1364, opened 2019-04-05
Support ascii STL files


simonschmeisser:
rviz currently uses handcrafted code for loading STLs and assimp for everything else
https://github.com/ros-visualization/rviz/blob/74334b765e0eea6d1de371ae8a9c32d769e51e4d/src/rviz/mesh_loader.cpp#L700

does anybody know why? I'm asking because this handcrafted code does not support ASCII format of STL files which creates a lot of confusion among our test users/customers. I will open a PR to simply use assimp for everything

rhaschke:
@wjwwood, @jfaust, do you remember the reason for not using assimp for STLs?
Looking into the commit history, it looks like STL loading was there first and later assimp was added to support more formats as well. So, if assimp can handle all formats, this proposal looks reasonable to me.

wjwwood:
I don't know, if code can be deleted for assimp instead, that's fine with me. Though something like this should be tested well (maybe even ask the community to test is specifically).

jfaust:
The commit in question: https://github.com/ros-visualization/rviz/commit/3117a98853b5afe8d3d1ba54308a8e70b58033e6

That's from 9 years ago, so I don't remember specifically, but based on my experience with assimp in general, I would guess it choked on some of the STL files we had at the time, and it was easier to write our own parser than to fix theirs.

rhaschke:
Addressed in #1365.

PR #1365, opened 2019-04-06
Use assimp to load STL files

simonschmeisser:
### Description

rviz currently uses home grown code for loading STL files which does not support ASCII format but only binary format. As discussed in #1364 this is probably for historical reasons, ie stl_loader predates usage of assimp for loading other file formats (most notably collada dae)

This PR removes STL_loader and uses assimp instead.

### Remaining issues
 - [x] Loads a sample setup from ros-i/fanuc
 - [ ] assumes all files to be collada and tries to read scaling info from stl 
` [ERROR] [1554565767.625447979]: XML parse error [package://fanuc_cr7ia_support/meshes/cr7ia/visual/base_link.stl]: XML_ERROR_PARSING_TEXT
`
 - [x] Unit test fails to load stl files, ~~most likely~~ due to incomplete setup
 - [ ] test for regressions on a subset of ros-i meshes
 - [ ] discuss regression of last sample file (file contains more binary data than specified in header)


simonschmeisser:
Quick note on the ascii test file included with rviz:
if contains `end loop` and `end facet`  while typical files and the sample at wikipedia have `endloop` and `endfacet` (assimp seems to start a new facet instead of ending the current at the space between end and facet)

simonschmeisser:
@rhaschke I had another look at this testcase and there seems to be no way to run this truly headless (`unset DISPLAY`). What works however is running it with
`xvfb-run make run_tests_rviz_gtest_stl_loader_test
` Do you think this is a viable approach and if so do you know how to combine it with moveit_ci? I see two options, either telling `catkin_add_gtest` somehow to prefix xvfb-run to just this test or running the whole test within xvfb-run. (Starting Xvfb as a background process unfortunately lead to a segfault when running the test for me)

rhaschke:
The test fails on the ROS build farm, which indeed doesn't provide a (virtual) X server. moveit_ci in Travis does. What we do in MoveIt in such a case is disabling the test, see [here](https://github.com/ros-planning/moveit/blob/9f3773d849f7fb98c8d7519d29d508ab9f5d4ca7/moveit_ros/perception/mesh_filter/CMakeLists.txt#L17-L22).
I'm puzzled though, why this PR is not tested on Travis as well.

simonschmeisser:
I had the "WIP" set until a moment ago, maybe that stops travis from running

rhaschke:
> I had the "WIP" set until a moment ago, maybe that stops Travis from running.

Interesting. Looks like this disables Travis indeed. Didn't know so far...

As your PR changes ABI, we can only schedule it for Noetic.

simonschmeisser:
Thanks for your hint for Xvfb!

I fixed ABI compatibility by re-adding the stl_loader class. I guess I should add a `deprecated` of some sort to it? Will this "dead-code" then be removed somehow?

I re-added the files using `git checkout HEAD~4 filename`, does this preserve history (after squashing)?

rhaschke:
Yes, please mark the class (or its constructor(s)) as deprecated with a TODO comment when to remove it (in Noetic). Squashing creates a new commit based on the actual remaining differences, so yes, history will be preserved.

simonschmeisser:
I added a deprecation note and comment. Please note that this will be quite noisy without #1367

src/test/stl_loader_test.cpp
```diff
@@ -1,44 +1,51 @@
...
-  EXPECT_TRUE(loader.load(meshFilePath.string()));
+  meshFilePath = meshDir + "valid.stl";
+  EXPECT_TRUE(rviz::loadMeshFromResource(meshFilePath).get());
 
   // Load a "potentially" valid STL binary file with bigger size than the
   // expected. The extra "unexpected" data at the end of the file should be
-  // ignored.
-  meshFilePath = meshDir / "valid_extra.stl";
-  EXPECT_TRUE(loader.load(meshFilePath.string()));
+  // ignored. Maybe. But is not ignored by assimp
```
rhaschke: I'm a little bit hesitant to merge this as is, because of this "regression".
Formally, of course, proper .stl files shouldn't have excess data at the end of the file.
However, I'm not sure how many users are out there that have such broken .stl files and rely on rviz to load them anyway. @wjwwood What is your recommendation?

src/test/stl_loader_test.cpp
```diff
@@ -1,44 +1,51 @@
...
+  meshFilePath = meshDir + "valid_extra.stl";
+  EXPECT_FALSE(rviz::loadMeshFromResource(meshFilePath).get());
 }
 
 int main( int argc, char **argv ) {
   testing::InitGoogleTest( &argc, argv );
+
+  // initialize ogre resource system
+  rviz::RenderSystem::get();
```
rhaschke: It's somewhat annoying that we need to initialize the whole rendering system (and thus rely on X11), just to load some meshes.
Due to this limitation, this unit test will not run on the build farm anymore...

I tried to work around this issue, just initializing the `ResourceManager`, but finally materials didn't have any techniques configured.

src/test/stl_loader_test.cpp
simonschmeisser: Well at least MoveIt ( https://github.com/ros-planning/geometric_shapes/blob/melodic-devel/src/mesh_operations.cpp ) uses assimp directly as well so these files would not work in a MoveIt context

simonschmeisser:
thanks for merging @rhaschke

Issue #1366, opened 2019-04-08
Error output loading e.g. 3ds meshes


dseifert:
Due to the merge of #1317 (subsequently changed in #1352), I am now seeing the following error for my 3ds meshes when using rviz 1.13.3 in ROS Melodic on Ubuntu 18.04:

`[ERROR] [1554754289.703535903, 1.974491951]: /rviz: XML parse error [package://my_package/meshes/fancy_mesh.3ds]: XML_ERROR_PARSING_TEXT`

Reason: `loadMeshFromResource` calls `meshFromAssimpScene` for 3ds files, and then subsequently `getMeshUnitRescale` is called even if the file is not a Collada XML file.

J-Rojas:
I have the same error for OBJ mesh files.

rhaschke:
Addressed in #1365.

PR #1367, opened 2019-04-08
Verify file extension before trying to parse Collada XML (.dae) files…

dseifert:
Before trying to parse a Collada XML file in mesh_loader.cpp's `getMeshUnitRescale()`, verify the file extension. This function is called for all files that are loaded via Open Asset Import Library (Assimp) (#1366)

NOTE: I do not have any Collada files for testing, this needs to be done before accepting this PR.

### DIFF
--- a/src/rviz/mesh_loader.cpp
+++ b/src/rviz/mesh_loader.cpp
@@ -30,6 +30,7 @@
 #include "mesh_loader.h"
 #include <resource_retriever/retriever.h>
 
+#include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
 
 #include "ogre_helpers/stl_loader.h"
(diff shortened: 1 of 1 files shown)


simonschmeisser:
I wonder if there could be any other extension used for collada (or maybe none at all?), otherwise +1

dseifert:
None that I am aware of or that googling pointed me to.

simonschmeisser:
@rhaschke we could add some test-cases for collada with #1364

rhaschke:
Instead of ignoring files based on their suffix, I suggest to silently ignore xml parse errors:
If the file couldn't be parsed in first place, this would be handled before.

simonschmeisser:
true, so ROS_DEBUG instead of ROS_ERROR or complete silence?

rhaschke:
Yes.

simonschmeisser:
which one :D complete silence?

rhaschke:
ROS_DEBUG will be fine

rhaschke:
Integrated alternative patch 68a375a into #1365.

PR #1370, opened 2019-04-24
Fix compilation with Qt4

### LINKED ISSUE #1369: Qt5 dependency on Kinetic - Cannot build with Qt4
rviz will not build with Qt4 on branch **kinetic-devel** : 
```
.../rviz/src/rviz/visualization_frame.cpp:1253:66: error: ‘FindDirectChildrenOnly’ is not a member of ‘Qt’
   QList<QTabBar *> tab_bars = findChildren<QTabBar *>(QString(), Qt::FindDirectChildrenOnly);
```
[Merged PR #1168](https://github.com/ros-visualization/rviz/pull/1168) relies on `Qt::FindDirectChildrenOnly` to get a list of all `QTabBar` children of `VisualizationFrame`, which is a flag to `QObject::findChild()` that was introduced in Qt5.

See `visualization_frame.cpp`, line 1253 :  
```cpp
QList<QTabBar *> tab_bars = findChildren<QTabBar *>(QString(), Qt::FindDirectChildrenOnly);
```
Qt4 does not seem to provide an "easy" equivalent of this flag. The subject of reproducing its behaviour has been covered [here](https://stackoverflow.com/questions/8740278/is-there-a-simple-way-of-limiting-qobjectfindchild-to-direct-children-only), but proposed solutions rely on the children's object names, which I am not sure will be reliable in the context of rviz.

### Your environment
* OS Version: e.g. Ubuntu 16.04
* ROS Distro: Kinetic
    ```
    no build, Qt 4.8.7
    ```

VictorLamoine:
Fixes #1369

This removes text eliding in dock widgets because the `FindDirectChildrenOnly` flag is only available starting from Qt 5

@theobslhc can you test compilation with Qt4?

### DIFF
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -625,10 +625,12 @@ void VisualizationFrame::openNewPanelDialog()
   if( dialog->exec() == QDialog::Accepted )
   {
     QDockWidget *dock = addPanelByName( display_name, class_id );
+#if QT_VERSION >= 0x050000
     if ( dock )
     {
       connect( dock, SIGNAL( dockLocationChanged( Qt::DockWidgetArea )), this, SLOT( onDockPanelChange() ) );
     }
+#endif
   }
   manager_->startUpdate();
 }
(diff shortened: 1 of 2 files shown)


theobslhc:
@VictorLamoine thank you !
Compilation works fine with Qt4.
No problem at runtime with the configuration file you link to in [PR #1168](https://github.com/ros-visualization/rviz/pull/1168).

rhaschke:
This solution essentially reverts #1168 for Qt4. To make the change less intrusive, I suggest to mask out only the content of `onDockPanelChange` but keep all other code as is. This will also maintain ABI and API compatibility.
Even better, you should mimic `FindDirectChildrenOnly` by checking `child->parent == this` ;-)

theobslhc:
Thanks @rhaschke for the suggestion. Here's another PR : #1371

PR #1374, opened 2019-05-10
Enable non-ascii chars for MovableText

rhaschke:
As pointed out in https://github.com/ros-visualization/interactive_markers/issues/40, rviz doesn't support non-ASCII chars in rendered text. This was due to two shortcomings:
1. The std::string, which is utf8-encoded on Linux nowadays, wasn't interpreted as an utf8-encoded string, but as an ASCII string.
2. To render non-ASCII chars, the font needs to load corresponding textures. By default, it looks like only 0-127 are loaded, which makes sense to limit memory usage. 

I extended the range arbitrarily to 0-999, solving the mentioned issue:
rviz-utf8

However, that's not an optimal solution at all, because it loads many textures that aren't probably used by most applications. On the other hand, still lots of chars are missing.
What would be needed, is a mechanism to explicitly specify the required codepoint range, either
- dynamically, extending the range based on required chars and reloading the font if necessary
- or statically, using e.g. some ROS parameter to list the code point range

While the first option would be the most convenient, I have no idea how to reload a font resource.

### DIFF
--- a/src/rviz/ogre_helpers/movable_text.cpp
+++ b/src/rviz/ogre_helpers/movable_text.cpp
@@ -116,6 +116,8 @@ void MovableText::setFontName(const String &fontName)
       throw Exception(Exception::ERR_ITEM_NOT_FOUND, "Could not find font "
           + fontName, "MovableText::setFontName");
 
+    // to support non-ascii letters, setup the codepoint range before loading
+    mpFont->addCodePointRange(std::make_pair<Ogre::Font::CodePoint>(0, 999));
     mpFont->load();
     if (!mpMaterial.isNull())
     {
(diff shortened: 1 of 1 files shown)


simonschmeisser:
Do you know how much additional memory we are talking about? 1000 is indeed a small subset of the 1,112,064 code points

rhaschke:
> Do you know how much additional memory we are talking about?

No. not at all.

> 1000 is indeed a small subset of the 1,112,064 code points

But already way better than 127 :wink:

simonschmeisser:
I guess a superior solution would be to read out which code point ranges the current font supports and load only these ( I suspect those not to be necessarily continuous ...), but it already looks like a good improvement

Issue #1382, opened 2019-06-25
Problems with the QT_INSTALL_LIBS environment var during build.


bionade24:
Hi, I'm the AUR ros melodic maintainer and many people are facing during build the issue with the latest rviz release (1.13.3), that the generated Makefile uses `QT_INSTALL_LIBS`, but this var is not set on Arch.
```
g++: error: 0/libQt5Core.so: No such file or directory
make[3]: *** [Makefile:38: "/home/user/.cache/yay/ros-melodic-rviz/src/build/devel/lib/python3.7/site-packages/rviz/librviz_sip".so] Error 1
make[2]: *** [src/python_bindings/sip/CMakeFiles/librviz_sip.dir/build.make:61: devel/lib/python3.7/site-packages/rviz/librviz_sip.so] Error 2
make[1]: *** [CMakeFiles/Makefile2:4935: src/python_bindings/sip/CMakeFiles/librviz_sip.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```
When you set `QT_INSTALL_LIBS` to `"/usr/lib"` or `'/usr/lib'`, `/bin/sh` returns 1, because it can't handle the patterns.
```
/bin/sh: /usr/lib: syntax error: operand expected (error token is "/usr/lib")
```
Is the second issue from a shell script without shebang or from the Makefile itself?

calvertdw:
I fixed it by doing the following on Arch:
- Run `colcon build --symlink-install` on ros2.repos master
- Get the error above
- Edit `build/qt_gui_cpp/sip/qt_gui_cpp_sip/Makefile:12` replacing `$$[QT_INSTALL_LIBS]` with `/usr/lib`
- Running `colcon` build again

bionade24:
@calvertdw [We already did this](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ros-melodic-rviz), but I don't want to have such a messy hack in my PKGBUILDs. This has to be fixed upstream but I don't have the time currently to find out what is causing this. Will look after it next week.

jwhendy:
Is [this](https://github.com/ros-visualization/python_qt_binding/pull/64) related?

bionade24:
Seems so. Hope it get's merged soon. Or should we patch it manually?

jwhendy:
[...]

PR #1383, opened 2019-07-09
Add python bindings option

ahoarau:
This just adds a python binding option that defaults to *ON* to be safe.
This is in part due to https://github.com/ros-visualization/rviz/issues/1382

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,7 +143,10 @@ find_package(catkin REQUIRED
   visualization_msgs
 )
 
-find_package(PythonLibs "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" REQUIRED)
+option(BUILD_PYTHON_BINDINGS "Build the python bindings using sip" ON)
+if(BUILD_PYTHON_BINDINGS)
+    find_package(PythonLibs "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" REQUIRED)
+endif()
 
 find_package(Eigen3 QUIET)
 if(NOT EIGEN3_FOUND)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -143,7 +143,10 @@ find_package(catkin REQUIRED
   visualization_msgs
 )
 
-find_package(PythonLibs "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" REQUIRED)
+option(BUILD_PYTHON_BINDINGS "Build the python bindings using sip" ON)
```
rhaschke: `sip` is only one option to build the python bindings.
```suggestion
option(BUILD_PYTHON_BINDINGS "Build python bindings" ON)
```

ahoarau:
I'm using the c++ `librviz` for one of my projet. I had the exact same error as #1382 that stopped me from compiling anything on melodic.  One way to pass the option : 
```
catkin config --cmake-args -DBUILD_PYTHON_BINDINGS=OFF
```
But it applies for all the packages in the ws, it's true. Better than nothing :) 
Copying the sources in the project that needs `librviz` and adding it with `add_subdirectory(rviz)` might work just as well and allow to pass the option to it individually.

ahoarau:
So are there any issues on this one ?

rhaschke:
The issue #1382 should be solved via https://github.com/ros-visualization/python_qt_binding/pull/64.
Could you please verify? If so, I suggest to cancel this PR.

> `catkin config --cmake-args -DBUILD_PYTHON_BINDINGS=OFF`

The option name is too generic. Better use something like `RVIZ_PYTHON_BINDINGS`.

rhaschke:
Closing due to inactivity. Issue should be fixed as indicated in https://github.com/ros-visualization/rviz/pull/1383#issuecomment-525206015

PR #1387, opened 2019-07-15
Fix opacity of Collada meshes

### LINKED ISSUE #1385: For collada meshes alpha of multiple RobotModel interfere
We noticed today that adding a second RobotModel display (or derived MoveIt ones) with alpha = 0.2 changes the display of the primary RobotModel to also have transparency. This only happens for Collada meshes, not STL

I first enable one RobotModel with alpha = 1, the robot uses individual collada meshes while gripper and camera are stl meshes

image

now I enabled a second RobotModel where alpha=0.2 and it will (visually) override the value for both (the input field is not affected)

image

Notice how both robots now have alpha=0.2 except for the gripper and camera

I suspect this is due to non unique material names, any idea @rhaschke ?

### Your environment
* OS Version: Ubuntu 18.04
* ROS Distro: Melodic
    ```
    [ INFO] [1562941360.215562081]: rviz version 1.13.3
[ INFO] [1562941360.215614003]: compiled against Qt version 5.9.5
[ INFO] [1562941360.215624573]: compiled against OGRE version 1.9.0 (Ghadamon)
[ INFO] [1562941360.431414141]: Stereo is NOT SUPPORTED
[ INFO] [1562941360.432597198]: OpenGl version: 3 (GLSL 1.3).
    ```

rhaschke:
Fixes an issue introduced in #1294 (e906f82eb2ef04e63dea03f834ae722f144bc712):
Materials of Collada mesh models need to be cloned for each instance of a robot link.
Fixes #1385. @simonschmeisser, could you please verify?

### DIFF
--- a/src/rviz/robot/robot_link.cpp
+++ b/src/rviz/robot/robot_link.cpp
@@ -632,6 +632,13 @@ void RobotLink::createEntityForGeometryElement(const urdf::LinkConstSharedPtr& l
       {
         sub->setMaterial(default_material_);
       }
+      else
+      {
+          // create a new material copy for each instance of a RobotLink
+          Ogre::MaterialPtr mat = Ogre::MaterialPtr(new Ogre::Material(nullptr, material_name, 0, ROS_PACKAGE_NAME));
+          *mat = *sub->getMaterial();
+          sub->setMaterial(mat);
+      }
       materials_[sub] = sub->getMaterial();
     }
   }


src/rviz/robot/robot_link.cpp
```diff
@@ -632,6 +632,13 @@ void RobotLink::createEntityForGeometryElement(const urdf::LinkConstSharedPtr& l
       {
         sub->setMaterial(default_material_);
       }
+      else
+      {
+          // create a new material copy for each instance of a RobotLink
+          Ogre::MaterialPtr mat = Ogre::MaterialPtr(new Ogre::Material(nullptr, material_name, 0, ROS_PACKAGE_NAME));
+          *mat = *sub->getMaterial();
```
simonschmeisser: I have to admit I don't quite understand this line

simonschmeisser:
@rhaschke good job, works as expected now. Thanks!

src/rviz/robot/robot_link.cpp
rhaschke: Create a new material instance and copy the existing material into this instance.

simonschmeisser: ah, ok, thanks, I was reading to much into the first line/constructor

PR #1388, opened 2019-07-19
Fix #1120 Visualization markers ignore color and show everything as white

### LINKED ISSUE #1120: Visualization markers ignore color and show everything as white

xbroquer:
### Description
Fix #1120 Visualization markers ignore color and show everything as white using  #752 as base patch

### Checklist

- Rendering issues:
  - Attached broken rendering image: issue

  - Attached fixed rendering image: mesh_marker_fixed

- Source code snippet for three cases:

// display marker not using mesh material in transparent blue
          marker.color.r = 0;
          marker.color.g = 0;
          marker.color.b = 1;
          marker.color.a = 0.3;
          marker.scale.x = 1;
          marker.scale.y = 1;
          marker.scale.z = 1;
          marker.type = visualization_msgs::Marker::MESH_RESOURCE;
          marker.mesh_resource = "package://my_pkg/moonwalk.dae";
          marker.mesh_use_embedded_materials=false;

// display marker  using mesh material and tinted with transparent green 
        marker.scale.x = 1;
        marker.scale.y = 1;
        marker.scale.z = 1;
        marker.color.r = 0;
        marker.color.g = 1;
        marker.color.b = 0;
        marker.color.a = 0.5f;
        marker.type = visualization_msgs::Marker::MESH_RESOURCE;
        marker.mesh_resource = "package://my_pkg/moonwalk.dae";
        marker.mesh_use_embedded_materials=true;

// display marker  using mesh material and not tinted
        marker.scale.x = 1;
        marker.scale.y = 1;
        marker.scale.z = 1;
        marker.color.r = 0;
        marker.color.g = 0;
        marker.color.b = 0;
        marker.color.a = 0;
        marker.type = visualization_msgs::Marker::MESH_RESOURCE;
        marker.mesh_resource = "package://my_pkg/moonwalk.dae";
        marker.mesh_use_embedded_materials=true;

### DIFF
--- a/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
+++ b/src/rviz/default_plugin/markers/mesh_resource_marker.cpp
@@ -104,10 +104,6 @@ void MeshResourceMarker::onNewMessage(const MarkerConstPtr& old_message, const M
   Ogre::SceneBlendType blending;
   bool depth_write;
 
-  // Keep a reference of default material created below to update color when needed
-  Ogre::MaterialPtr default_material;
-  default_material.setNull();
-
   if (a < 0.9998)
   {
     blending = Ogre::SBT_TRANSPARENT_ALPHA;
@@ -152,7 +148,7 @@ void MeshResourceMarker::onNewMessage(const MarkerConstPtr& old_message, const M
     
     // create a default material for any sub-entities which don't have their own.
     ss << "Material";
-    default_material = Ogre::MaterialManager::getSingleton().create(ss.str(), ROS_PACKAGE_NAME);
+    Ogre::MaterialPtr default_material = Ogre::MaterialManager::getSingleton().create(ss.str(), ROS_PACKAGE_NAME);
     default_material->setReceiveShadows(false);
     default_material->getTechnique(0)->setLightingEnabled(true);
     default_material->getTechnique(0)->setAmbient(0.5, 0.5, 0.5);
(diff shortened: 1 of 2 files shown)


rhaschke:
#752 should already part of the source history. Did you have a look into #893 as well, which seems to revert the first one?

xbroquer:
Hi Robert,

Thanks for pointing me #893. Seems that this issue has a long story :)

With rviz melodic release, only external image textures (like PR2 base) are  displayed. When materials are fully defined inside collada file, the mesh is white.

I also used garaemon/rviz_collada_marker fir my tests and added moonwalk marker (that is not using external image files for materials).

The result of rviz melodic release (arrow and moonwalk are white):
Screenshot from 2019-07-22 11-09-00

When mesh_use_embedded_materials=True, color is not dynamically updated

Here is the result with my latest commit:
Screenshot from 2019-07-22 11-08-41

Here the 2 videos that shows the dynamic color update:
[rviz-melodic-collada-test.zip](https://github.com/ros-visualization/rviz/files/3416753/rviz-melodic-collada-test.zip)

However, as you can see, the usage of passes changes the behavior when mesh_use_embedded_materials=True and a color is set. 
The material from the mesh cannot be transparent. If we want the marker transparent we have to set mesh_use_embedded_materials=False (third marker).
**Do you think this behavior change is acceptable ?**

rhaschke:
I don't think that adding a second pass is the right solution to the problem: It essentially just redraws the mesh with the user-defined color. As you see from the comparison, this doesn't allow to change the transparency of the mesh.
Another option is to use alpha from the user-defined color for both passes, and use the second pass to tint the mesh with the user-defined color as suggested in this PR. However, this will also modify the result for texture-colored meshes.
I think, we need to modify the color "towards" the user-defined one in first place. A single pass will suffice then for drawing. One option is to multiply with the user-defined color.

 multiply | current rviz | pull request 
-----------|-----------|--------
mesh-multiply | mesh-original | mesh-pr

rhaschke:
Closing in favor of #1424. @xbroquer, could you please review the new PR?

PR #1393, opened 2019-07-31
Fix segfault when camera dispay closed. Closes #1372.

### LINKED ISSUE #1372: Camera display type crashes rviz (Melodic) when deselected or deleted
Camera display type crashes rviz when it it deselected or deleted. Recreated with multiple cameras on multiple machines.

### To recreate 
- Add any camera with a valid calibration
rviz_add camera

- Camera display image should be displayed with overlay
rviz_camer_active

- Either delete or de-select Camera display to cause crash.
rviz_crash

### Your environment
* OS Version: Ubuntu 18.04
* ROS Distro: Melodic
    rviz version 1.13.3
    compiled against Qt version 5.9.5
    Stereo is NOT SUPPORTED
    OpenGl version: 4.6 (GLSL 4.6).
    
* Console output
```
Gtk-Message: 12:27:56.341: GtkDialog mapped without a transient parent. This is discouraged.
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
[1]    25031 abort (core dumped)  rviz
```

* Install type 
    Aptitude

hexonxons:
### Description

Actually segfault happens in [`CBQueueCallback::call`](https://github.com/ros/geometry2/blob/f0c6ffd465619cfeb0bf89a9fd9d0ab3fb46c80e/tf2_ros/include/tf2_ros/message_filter.h#L596) method while calling `filter_->signalMessage(event_)`:
```c++
virtual CallResult call()
{
  if (success_)
  {
    filter_->signalMessage(event_);
  }
  else
  {
    filter_->signalFailure(event_, reason_);
  }

  return Success;
}
```

`CBQueueCallback` object is constructed in [`MessageFilter::messageReady`](https://github.com/ros/geometry2/blob/f0c6ffd465619cfeb0bf89a9fd9d0ab3fb46c80e/tf2_ros/include/tf2_ros/message_filter.h#L626) method:
```c++
ros::CallbackInterfacePtr cb(new CBQueueCallback(this, evt, true, filter_failure_reasons::Unknown));
callback_queue_->addCallback(cb, (uint64_t)this);
```
and it [holds `MessageFilter` pointer](https://github.com/ros/geometry2/blob/f0c6ffd465619cfeb0bf89a9fd9d0ab3fb46c80e/tf2_ros/include/tf2_ros/message_filter.h#L584) as argument:

```c++
 CBQueueCallback(MessageFilter* filter, const MEvent& event, bool success, FilterFailureReason reason)
    : event_(event)
    , filter_(filter)
    , reason_(reason)
    , success_(success)
    {}
```

At call time `filter_` object is already destroyed in https://github.com/ros-visualization/rviz/blob/57325fa075893de70f234f4676cdd08b411858ff/src/rviz/image/image_display_base.cpp#L204-L208 and this leads to segfault.

`CBQueueCallback` object is called from [`callback_queue_`](https://github.com/ros/geometry2/blob/f0c6ffd465619cfeb0bf89a9fd9d0ab3fb46c80e/tf2_ros/include/tf2_ros/message_filter.h#L631):
```c++
ros::CallbackInterfacePtr cb(new CBQueueCallback(this, evt, true, filter_failure_reasons::Unknown));
callback_queue_->addCallback(cb, (uint64_t)this);
```
which actually the same as `update_nh_.getCallbackQueue()`:

* `tf_filter` constructed from `update_nh_`: https://github.com/ros-visualization/rviz/blob/57325fa075893de70f234f4676cdd08b411858ff/src/rviz/image/image_display_base.cpp#L179-L185
* `MessageFilter` ctor [init `callback_queue_` as  `callback_queue_(nh.getCallbackQueue())`](https://github.com/ros/geometry2/blob/f0c6ffd465619cfeb0bf89a9fd9d0ab3fb46c80e/tf2_ros/include/tf2_ros/message_filter.h#L145):
```c++
template<class F>
  MessageFilter(F& f, tf2::BufferCore& bc, const std::string& target_frame, uint32_t queue_size, const ros::NodeHandle& nh)
  : bc_(bc)
  , queue_size_(queue_size)
  , callback_queue_(nh.getCallbackQueue())
  {
    init();

    setTargetFrame(target_frame);

    connectInput(f);
  }
```
This all looks like a thread sync problem, but I wasn't able to track it.
I am unsure if it is bug in TF2 message filter, so I suggest to use less invasive method and fix this bug in RViz code by removing callback by id before message_filter reset.

### DIFF
--- a/src/rviz/image/image_display_base.cpp
+++ b/src/rviz/image/image_display_base.cpp
@@ -46,6 +46,7 @@ ImageDisplayBase::ImageDisplayBase() :
     Display()
     , sub_()
     , tf_filter_()
+    , tf_filter_callback_id_(std::numeric_limits<uint64_t>::max())
     , messages_received_(0)
 {
   topic_property_ = new RosTopicProperty("Image Topic", "",
@@ -176,14 +177,17 @@ void ImageDisplayBase::subscribe()
       }
       else
       {
-        tf_filter_.reset(new tf2_ros::MessageFilter<sensor_msgs::Image>(
+        auto filter = new tf2_ros::MessageFilter<sensor_msgs::Image>(
           *sub_,
           *context_->getTF2BufferPtr(),
           targetFrame_,
           queue_size_property_->getInt(),
           update_nh_
-        ));
+        );
+        tf_filter_.reset(filter);
         tf_filter_->registerCallback(boost::bind(&ImageDisplayBase::incomingMessage, this, _1));
+        // See https://github.com/ros-visualization/rviz/issues/1372 PR.
+        tf_filter_callback_id_ = (uint64_t) filter;
       }
     }
     setStatus(StatusProperty::Ok, "Topic", "OK");
@@ -203,6 +207,7 @@ void ImageDisplayBase::subscribe()
 
 void ImageDisplayBase::unsubscribe()
 {
+  update_nh_.getCallbackQueue()->removeByID(tf_filter_callback_id_);
   tf_filter_.reset();
   sub_.reset(new image_transport::SubscriberFilter());
 }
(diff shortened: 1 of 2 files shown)


rhaschke:
Thanks a lot for your detailed analysis. Based on your work, I filed a PR in https://github.com/ros/geometry2/pull/402. Closing here, as this should be fixed upstream.

PR #1396, opened 2019-08-06
correctly initialise Ogre types and check return value of 'transform'

christian-rauch:
### Description

In some sections of the code, the Ogre types where not correctly initialized and return values have not been checked. This PR resolves these issues to prevent crashes due to invalid marker scales from uninitialized memory.

### DIFF
--- a/src/rviz/default_plugin/markers/shape_marker.cpp
+++ b/src/rviz/default_plugin/markers/shape_marker.cpp
@@ -79,9 +79,12 @@ void ShapeMarker::onNewMessage( const MarkerConstPtr& old_message,
     handler_->addTrackedObjects( shape_->getRootNode() );
   }
 
-  Ogre::Vector3 pos, scale, scale_correct;
+  Ogre::Vector3 pos(0,0,0);
+  Ogre::Vector3 scale(0,0,0);
+  Ogre::Vector3 scale_correct(0,0,0);
   Ogre::Quaternion orient;
-  transform(new_message, pos, orient, scale);
+
+  if(!transform(new_message, pos, orient, scale)) { return; }
 
   if (owner_ && (new_message->scale.x * new_message->scale.y
       * new_message->scale.z == 0.0f))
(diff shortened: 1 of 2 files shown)


src/rviz/default_plugin/markers/shape_marker.cpp
```diff
@@ -79,9 +79,12 @@ void ShapeMarker::onNewMessage( const MarkerConstPtr& old_message,
...
   }
 
-  Ogre::Vector3 pos, scale, scale_correct;
+  Ogre::Vector3 pos(0,0,0);
+  Ogre::Vector3 scale(0,0,0);
+  Ogre::Vector3 scale_correct(0,0,0);
   Ogre::Quaternion orient;
-  transform(new_message, pos, orient, scale);
+
+  if(!transform(new_message, pos, orient, scale)) { return; }
```
rhaschke: Wouldn't it suffice to have this line fixed, i.e. just return on failure?
All the uninitialized variables wouldn't be used in this case.

christian-rauch:
Yes, in this case it would be sufficient to return and keep the values uninitialised.
But uninitialised values are the root of evil. I usually initialise values on their creating as a good habit and the performance costs of initialising 3-element vectors is neglectable.

rhaschke:
@christian-rauch, could you provide a marker message causing rviz to fail on this?
From my understanding, `transform()` shouldn't fail here, since the validity of the transform is checked in advance by a TF message filter. So, I'm curious, which message causes this issue.

christian-rauch:
A transform was missing and `transform` returned `false` at https://github.com/ros-visualization/rviz/blob/f9b357fdc0bfea5361e27b506059e62eab1f861b/src/rviz/default_plugin/markers/marker_base.cpp#L93. This caused the function to return without setting `scale`.

rhaschke:
My point is, that if the transform is missing, the message filter should bail out before already.
If that's not happening, we have another issue.

christian-rauch:
I understand. But I am unable to reproduce this issue since I fixed the original problem in the URDF and a couple of other places.

rhaschke:
Superseded by #1400. @christian-rauch, please validate.

PR #1397, opened 2019-08-07
Add extra float checks before setting shape properties

### LINKED ISSUE #1394: rviz crashes when displaying faulty marker message
When displaying a marker, rviz crashes with the message:
`rviz: /build/ogre-1.9-mqY1wq/ogre-1.9-1.9.0+dfsg1/OgreMain/src/OgreNode.cpp:630: virtual void Ogre::Node::setScale(const Ogre::Vector3&): Assertion !inScale.isNaN() && "Invalid vector supplied as parameter"' failed.`
One of the values in the field `scale` in the marker has a very extreme value (6.34119308747e+180). This is of course itself an error, but rviz should catch that error instead of crashing. I think what happens is that the `validateFloats` in `marker_display.cpp` does not complain because it is a vector of doubles, but the `Real`-type in Ogre is a float, which cannot hold this large value, so the assertion in `setScale` fails. I am not sure how to best fix this issue, so I cannot provide a pull request.

To reproduce: display a marker topic in rviz and publish a faulty marker message with rostopic

### Your environment
* OS Version: Ubuntu 16.04
* ROS Distro: Kinetic (but probably also a problem in Melodic)
    ```
    rviz version 1.12.17
    compiled against Qt version 5.5.1
    Stereo is NOT SUPPORTED
    OpenGl version: 4.6 (GLSL 4.6).
    ```
* If source build, which git commit? latest kinetic-devel

mvieth:
Fixes #1394 
Please see issue description for further information

### DIFF
--- a/src/rviz/ogre_helpers/shape.cpp
+++ b/src/rviz/ogre_helpers/shape.cpp
@@ -28,6 +28,7 @@
  */
 
 #include "shape.h"
+#include "rviz/validate_floats.h"
 #include <ros/assert.h>
 
 #include <OgreSceneManager.h>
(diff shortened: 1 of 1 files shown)


rhaschke:
Superseded by #1400. @mvieth, please validate there.

rhaschke:
This PR added a full extra layer of checks, while #1400 just correctly routes the `validateFloats()` check to the float version.

PR #1400, opened 2019-08-12
Fix marker handling

rhaschke:
Fix segfaults due to invalid marker messages.


christian-rauch:
Could you backport the other changes as well? E.g. setting the scale before checking the transform. This allows alternative uses of the function to make use of the scale even if the transform is not valid. Also I would encourage to initialise all values in general in RViz. It helps to debug issues, e.g. by having zeros instead of random values.

rhaschke:
I intentionally didn't port these changes because they are unnecessary overhead as discussed in https://github.com/ros-visualization/rviz/pull/1396#pullrequestreview-273741397.
If `transform()` fails, the arguments are (still) invalid. Period.

PR #1402, opened 2019-08-19
Overload FailedDisplay::save for proper override

### LINKED ISSUE #1390: Coverity report bad_override
Method "rviz::FailedDisplay::save" hides but does not override "rviz::Property::save" because some type qualifiers do not match.
https://github.com/ros-visualization/rviz/blob/f9b357fdc0bfea5361e27b506059e62eab1f861b/src/rviz/failed_display.cpp#L72-L80

The function should add `const` keyword

artivis:
### Description

Overload `FailedDisplay::save` for properly overriding `Display::save` without breaking ABI.

Fix #1390

### DIFF
--- a/src/rviz/failed_display.cpp
+++ b/src/rviz/failed_display.cpp
@@ -71,6 +71,11 @@ void FailedDisplay::load( const Config& config )
 
 void FailedDisplay::save( Config config )
 {
+  save(config);
+}
+
+void FailedDisplay::save( Config config ) const
+{
   if( saved_config_.isValid() )
   {
     config.copy( saved_config_ );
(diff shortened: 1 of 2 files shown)


src/rviz/failed_display.cpp
```diff
@@ -70,6 +70,11 @@ void FailedDisplay::load( const Config& config )
 }
 
 void FailedDisplay::save( Config config )
+{
+  save(config);
```
rhaschke: Shouldn't this use a `const_cast` for `this` to actually redirect to the `const`-version and avoid an infinite loop?

PR #1403, opened 2019-08-22
Add TF aligned view controller

mcevoyandy:
### Description

This PR adds a view controller that aligns the camera to an axis of a published TF frame. The user is able to select the frame, select which axis the camera should be aligned to, and also change the roll angle.

View Panel before:
tf_aligned_view_before

View Panel after:
tf_aligned_view_after

Gif showing how the camera view can be aligned to a frame and how it follows that frame as it is moved around: (Note that I'm using [tf_visual_tools](https://github.com/PickNikRobotics/tf_visual_tools) to move the TF around)
tf_aligned_view

### DIFF
--- a/src/rviz/default_plugin/view_controllers/tf_aligned_view_controller.cpp
+++ b/src/rviz/default_plugin/view_controllers/tf_aligned_view_controller.cpp
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2019, PickNik, LLC.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of PickNik, LLC nor the names of its
+ *       contributors may be used to endorse or promote products derived from
+ *       this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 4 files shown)


rhaschke:
From what I understand, this ViewController is very similar to the FPS controller, isn't it?
The only thing missing from FPS is that the orientation of the reference frame is considered as well.

rhaschke:
Is it intended that the camera doesn't react to any mouse events? Wouldn't that be much easier and more flexible than choosing the x/y/z axis via GUI elements?

rhaschke:
I implemented a simple alternative to this PR, based on the existing FPS view controller: #1405.
Please give it a try.

rhaschke:
The key difference between #1405 and FPS is that #1405 also considers the orientation of the tracked frame, which FPS doesn't do.

> The intention of this PR is to allow an easy way get perfectly aligned with one of the axes

Why do you think it is important to _exactly_ align the view direction with an axis?

rhaschke:
> Mostly the use case is to align the RViz view with a camera position

This not only includes camera pose but also internal camera parameters like zoom.
To serve this use case, there is the `Camera` display available, which reads these camera parameters and renders the scene from exactly this view. Usually, you also underlay the image stream from this camera, but that's not necessary.

src/rviz/default_plugin/view_controllers/tf_aligned_view_controller.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+  camera_->setOrientation(reference_orientation_);
+  camera_->setPosition(reference_position_);
+  updateCamera();
+}
+
+void TfAlignedViewController::update(float dt, float ros_dt)
+{
+  FramePositionTrackingViewController::update( dt, ros_dt );
+  updateCamera();
```
rhaschke: I think it is not necessary to perform the computations in every `update()` cycle.
This should only be done in response to gui changes.

src/rviz/default_plugin/view_controllers/tf_aligned_view_controller.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+void TfAlignedViewController::updateCamera()
+{
+  Ogre::Quaternion q_offset;
+  int view_direction = invert_axis_->getBool() ? -1 : 1;
+  int view_axis = axis_property_->getOptionInt() * view_direction;
+
+  // reference_orientation_ will be aligned to -z-axis by default.
+  switch (view_axis) {
```
rhaschke: Using hand-crafted equations for these computations is bad programming practice: The source code doesn't reveal the underlying concepts, rendering this code nearly un-maintainable.

rhaschke:
Thanks for the suggestion of this view controller.

Issue #1260, opened 2018-06-15
Starting rviz in fullscreen


davidbsp:
Is there an option to start rviz in fullscreen (cmd line args or via saved configuration file) ?

zaid-g:
Looking for the same thing. Hope someone has an answer

VictorLamoine:
There is no way right now, full screen was added in this merge request: https://github.com/ros-visualization/rviz/pull/1017/files

There is a menu and a shortcut only at the moment

rhaschke:
Fixed via #1413: new --fullscreen cmdline option.

edgariph98:
how would I be able to set the configuration to be full screen?

rhaschke:
> how would I be able to set the configuration to be full screen?

@edgariph98, as written in the previous post (and document with `--help`):
`rviz --fullscreen`

edgariph98:
> > how would I be able to set the configuration to be full screen?
> 
> @edgariph98, as written in the previous post (and document with `--help`): `rviz --fullscreen`

is there a way to set a flag from a config.rviz file to start the rviz in full-screen mode? I am currently loading the rviz visualization frame into a window and I only want to see the rendering window, is there another way to achieve this instead of full-screen mode?

rhaschke:
No, I don't think so.

PR #1413, opened 2019-09-01
Improve cmdline parsing and add --fullscreen option

rhaschke:
Address #1260.

### DIFF
--- a/src/rviz/visualizer_app.cpp
+++ b/src/rviz/visualizer_app.cpp
@@ -131,28 +131,25 @@ bool VisualizerApp::init( int argc, char** argv )
 
     startContinueChecker();
 
+    std::string display_config, fixed_frame, splash_path, help_path;
+    int force_gl_version = 0;
+
     po::options_description options;
     options.add_options()
       ("help,h", "Produce this help message")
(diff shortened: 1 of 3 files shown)


src/rviz/visualizer_app.cpp
```diff
@@ -164,62 +161,12 @@ bool VisualizerApp::init( int argc, char** argv )
...
-      {
-        disable_anti_aliasing = true;
-      }
-
-      if (vm.count("verbose"))
-      {
-        verbose = true;
+        std::cerr << "ERROR: the config file '" << display_config << "' is a .vcg file, which is the old rviz config format." << std::endl;
+        std::cerr << "       New config files have a .rviz extension and use YAML formatting.  The format changed" << std::endl;
+        std::cerr << "       between Fuerte and Groovy.  There is not (yet) an automated conversion program." << std::endl;
```
simonschmeisser: > not (yet)
sounds optimistic :wink:

rhaschke: Yeah, probably it's time to remove this check completely. I will do this for Noetic.

Issue #1414, opened 2019-09-01
Melodic release


rhaschke:
I would like to prepare a new release into Melodic asap. However, because there are many new commits, I would like to get some feedback from the community first to resolve potential regressions. Particularly, @wjwwood, @simonschmeisser, @VictorLamoine, @v4hn could you please use the `melodic-devel` branch on a daily basis and report (new) issues?

* [feature] RobotModelDisplay: Poll robot_description at 1Hz until found
* [feature] Use assimp to load STL files (#1365)
* [feature] Added frame-aligned view controller (#1405)
* [feature] Systematically check parameters of incoming marker messages (#1275, #1400)
* [feature] MarkerDisplay: clear statuses on disable()
* [feature] Increase zoom range in orbit view controllers (#1373)
* [feature] Add shortcut "C" to the Publish Point tool (#1321)
* [fix] Fixed several issues with Camera and Image display (#1409)
* [fix] Fixed several issues with view controllers, particularly FPS (#1407)
* [fix] Fixed (small) memory leak in TFDisplay (#1408)
* [fix] Fixed memory leak in MapDisplay (#1406)
* [fix] Properly resize RenderWidget on high-DPI displays (#1263)
* [fix] Fixed opacity of Collada meshes (#1387)
* [fix] Enable non-ascii chars for MovableText (#1374)
* [fix] Ignore nan and inf values when normalizing images (#1378)
* [fix] MovableText: consider full translation vector (#1375)
* [maintanence] ToolManager: simplify key handling code
* [maintanence] Move sip bindings' build directory into the build space. (#1360)
* [maintanence] Use std random generators for portability on Windows (#1356)
* [api] public CovarianceVisual (#1410)
* [api] Properly override FailedDisplay::save (#1402)
* [api] Removed extra wrapper MarkerArrayDisplay::handleMarkerArray (#1401)

v4hn:
Thanks for asking.
Running it, I noticed a regression introduced by #1275 . See #1415 .

Also, I've been fighting a lot with the bugs mentioned in #1409 recently.
I'll test the latest branch with our setup later today.

simonschmeisser:
There seems to be quite some visual regression. Before:
![pastedImage](https://user-images.githubusercontent.com/372442/64107045-da606c80-cd79-11e9-9b39-d1b3b9f59a08.png)

after:
[...]

PR #1417, opened 2019-09-03
fix crash on empty text in Interactive Marker

simonschmeisser:
this fixes a regression/crash introduced in #1275 and discussed in #1414 that occurred when adding an empty text marker

### DIFF
--- a/src/rviz/default_plugin/markers/text_view_facing_marker.cpp
+++ b/src/rviz/default_plugin/markers/text_view_facing_marker.cpp
@@ -44,7 +44,7 @@ namespace rviz
 
 TextViewFacingMarker::TextViewFacingMarker(MarkerDisplay* owner, DisplayContext* context, Ogre::SceneNode* parent_node)
 : MarkerBase(owner, context, parent_node)
-, text_(0)
+, text_(nullptr)
 {
 }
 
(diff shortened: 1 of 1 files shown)


src/rviz/default_plugin/markers/text_view_facing_marker.cpp
```diff
@@ -58,7 +58,10 @@ void TextViewFacingMarker::onNewMessage(const MarkerConstPtr& old_message, const
   ROS_ASSERT(new_message->type == visualization_msgs::Marker::TEXT_VIEW_FACING);
 
   if (new_message->text.find_first_not_of(" \t\n\v\f\r") == std::string::npos)
+  {
+    delete text_;
```
v4hn: Without having tested it:

```suggestion
    delete text_;
    text_ = nullptr;
```

simonschmeisser: Thanks, good catch @v4hn 

I'm not sure about the c++ version we are targeting here, I think melodic is only systems that default to c++14 but as I didn't see anything specific in CMakeLists.txt I would probably do `text_ = 0;`

```suggestion
    delete text_;
    text_ = 0;
```

simonschmeisser: I found the line where c++11 is enabled so `nullptr` is fine, I also added a line to reset the `MarkerSelectionHandler`

PR #1419, opened 2019-09-04
InteractiveMarker: continue with next marker if type unknown

### LINKED ISSUE #1418: InteractiveMarker: do not crash on unknown marker type
https://github.com/ros-visualization/rviz/blob/325e74cf3dc1bdd1db859beae88db85c926948c9/src/rviz/default_plugin/interactive_markers/interactive_marker_control.cpp#L93

should add a continue

simonschmeisser:
this fixes #1418 by no longer accessing nullptr
if an unknown marker type was requested


rhaschke:
Fixed as part of b1fc2e6214341b69d8d6370479577d9dd90505c3.

PR #1423, opened 2019-09-08
Fixes for #1275 and #1365

rhaschke:
This fixes some more issues of #1275:
- invalid quaternion: issue warning only
- triangle list accepts points.size()/3 colors as well (face colors)
- don't create (empty) text marker for interactive marker control with empty description 
(to avoid warning)

as well as #1365:
- use default material for .stl meshes

@v4hn and @simonschmeisser, could you please have a look?

### DIFF
--- a/src/rviz/default_plugin/marker_utils.cpp
+++ b/src/rviz/default_plugin/marker_utils.cpp
@@ -40,6 +40,7 @@
 #include "rviz/display_context.h"
 #include "rviz/properties/property.h"
 #include "rviz/properties/property_tree_model.h"
+#include "rviz/properties/status_property.h"
 #include "rviz/properties/status_list.h"
 #include "rviz/validate_quaternions.h"
 #include "rviz/validate_floats.h"
@@ -83,6 +84,22 @@ MarkerBase* createMarker(int marker_type, MarkerDisplay* owner, DisplayContext* 
   }
 }
 
+namespace
+{
+void addSeparatorIfRequired(std::stringstream& ss)
+{
+  if (ss.tellp() != 0) // check if string is not empty
+  {
+    ss << " \n";
+  }
+}
+
+void increaseLevel(::ros::console::levels::Level new_status, ::ros::console::levels::Level& current_status)
+{
+  if(new_status > current_status)
+    current_status = new_status;
+}
+
 template <typename T>
 constexpr const char* fieldName();
 template <>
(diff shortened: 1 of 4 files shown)


src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -338,12 +338,14 @@ void checkScaleText(const visualization_msgs::Marker& marker, std::stringstream&
     ss << "Text height of TEXT_VIEW_FACING is 0.0 (scale.z).";
     increaseWarningLevel(StatusProperty::Warn, level);
   }
+#if 0  // There is too much code out there, which sets all 3 axis scales
```
v4hn: It's somewhat sad to see this disabled. Still if you must, how about checking instead whether all three scales are equal to ignore "the common issue"?
This would still catch users who set scales independently.

I would still propose to enable the full check for noetic. After all this is almost trivial to fix in code bases.

src/rviz/default_plugin/marker_utils.cpp
```diff
@@ -279,7 +279,7 @@ void checkQuaternion(const visualization_msgs::Marker& marker, std::stringstream
   {
     addSeparatorIfRequired(ss);
     ss << "Unnormalized quaternion in marker message.";
-    increaseWarningLevel(StatusProperty::Error, level);
+    increaseWarningLevel(StatusProperty::Warn, level);
```
v4hn: :+1: 

minor point:
A few lines above, the warning states `Uninitialized quaternion assuming identity`, but whether that's actually the case depends on the module calling the function and is not determined here.
Are the quaternions implicitly normalized when they are not already?
It might make sense to add this to the warning message too then.

src/rviz/default_plugin/marker_utils.cpp
simonschmeisser: +1 for restricting this to actual issues, ie ignoring x == y == z

simonschmeisser:
Thanks for implementing the special handling for STL, I was going to do it the same way. :+1: 

Generally speaking while this is perfectly compatible with the previous stl loader I wonder about one thing:

- urdf spec does not specify what to do with materials/colors when the mesh file already supports/includes them. I'd assume the info in URDF overrides the info from the mesh (like we do with stl files containing colors [1]) but my understanding is that we currently just ignore the info from urdf. Which is sufficient I guess.

As I'm at home with a cold I have only slightly tested this code with publicly available data. I'm fine with merging as is but would like to wait with the release a few more days to get better testing.

[1] assimp supports stl files with color data but I have not yet encountered one in the wild ...

src/rviz/default_plugin/marker_utils.cpp
rhaschke: `rviz` becomes extremely verbose with all this checking enabled. Even a `interactive_markers` utility function sets all three scales.

v4hn: I do see your point :sunglasses: , it probably makes sense to ignore the special case `x==y==z` altogether then.

Still, I would think checking for different values and `x != 0 || y != 0` is a reasonable verification that could expose problems.
After all, the provided information for `x` and/or `y` is silently discarded.

simonschmeisser:
I did some testing on our actual software and so far it looks good :+1:

rhaschke:
I added another fixup for #1275 according to https://github.com/ros-visualization/rviz/issues/1120#issuecomment-530745422.

src/rviz/default_plugin/marker_utils.cpp
rhaschke: I solved the verboseness issue by distinguishing all console debug levels (and not only ok, warn, error as the rviz display does). This way, users can disable the verbose console output by adapting their verbosity level for the newly introduced logger `rviz.marker`.
The scaling check is performed again.

src/rviz/default_plugin/marker_utils.cpp
rhaschke: I reduced the severity level to a warning, because rviz doesn't crash with unnormalized quaternions. Errors will always cause the corresponding marker to be dropped, which shouldn't happen in this case.

PR #1424, opened 2019-09-09
Fix tinting of mesh markers

### LINKED ISSUE #1120: Visualization markers ignore color and show everything as white

rhaschke:
Fix #1120. This issue has a long story:
* Originally, it was fixed in #752 by adding a new rendering pass to realize color tinting.
* #893 essentially reverted this patch. The motivation at that time was that the second pass has essentially overwritten the first pass when the alpha value of the message color was `1.0`. Also, it wasn't possible to make the mesh transparent by setting a smaller alpha value.
* #1084 attempted to fix the white rendering issue again, but actually only skipped color tinting at all, when the message arrived the first time.
* This caused just another regression that was fixed in #1132, which brought us back to the start.
* #1388 attempts to re-apply #752 - an ever going cycle :wink:

Fortunately, @wjwwood in #1132 added a test program to validate rendering results. However, this test didn't include color-based meshes, which behave differently than texture-based ones.

Originally, i.e. before #752, tinting was achieved by overriding the ambient and diffuse colors of the mesh materials. That worked for texture-based meshes, but not for color-based ones (which just changed their color to the set one, ignoring their original mesh color).
The idea of #752 was to add a new rendering pass to re-render the mesh with the marker color.
In principle, this worked, but it became impossible to make the mesh transparent: The mesh was always rendered unmodified.

This PR additionally adapts the alpha value of the mesh rendering according to the alpha value of the marker color. This seems to yield the expected results:

current rviz | #752 / #1388  | this PR
-----------|-----------|--------
mesh-original | mesh-pr | mesh-pass

Notice, that texture-based meshes are rendered slightly differently now: Setting, like in the example, a red marker color (`rbga=[1,0,0,1]`) doesn't result in a fully red PR2 base anymore. Instead the originally white base is _tinted_ into red, resulting in a pinkish color. This is the very same effect we yield from tinting a white color-based mesh.
Actually, we would need another alpha parameter to exactly control the transparency / opaqueness of the tinting color separately from the mesh. However, we only have a single alpha parameter available and I decided to use it to control the transparency of the mesh itself, while the tinting transparency is `std::min(alpha, 0.5)`.


rhaschke:
Note: The Travis failure is an ABI incompatibility due to a removed, but previously unexposed symbol.
That's not related to this PR.

src/test/mesh_marker_test.cpp
```diff
@@ -38,26 +36,37 @@ publishText(int id, float x, float y, const std::string & text)
...
+      marker.type = Marker::SPHERE;
+      break;
   }
   marker.action = Marker::ADD;
   marker.pose.orientation.x = 0.0;
   marker.pose.orientation.y = 0.0;
-  marker.pose.orientation.z = 0.1;
-  marker.pose.orientation.w = 1.0;
+  marker.pose.orientation.z = 0.0995;
+  marker.pose.orientation.w = 0.995;
```
wjwwood: What's the purpose of this?

wjwwood: I suppose it doesn't matter since this is the test program, just curious.

rhaschke: This is just normalizing the quaternion. Having some more restrictive checking in place (#1275),  we should provide correctly normalized quaternions...

wjwwood: Ah makes sense. 👍

rhaschke: Please don't yet merge this PR, even if you have approved. I want to wait for some feedback from affected people (https://github.com/ros-visualization/rviz/issues/1120#issuecomment-529695711) until the weekend at least.

wjwwood:
@iche033 can you take a look.

src/test/mesh_marker_test.cpp
wjwwood: For sure, I'll let you take the reins on that. I also asked Ian to have a look, as he does most of the rendering related work in Gazebo.

iche033:
The idea of using multi-pass with alpha scene blending makes sense and the code changes also looks good to me. I ran the test and the results produced are expected.

src/rviz/default_plugin/markers/mesh_resource_marker.cpp
```diff
@@ -192,49 +179,57 @@ void MeshResourceMarker::onNewMessage(const MarkerConstPtr& old_message, const M
...
-      technique->setDepthWriteEnabled( depth_write );
-      technique->setLightingEnabled( true );
+      Ogre::Technique* technique = material->getTechnique(0);
+      Ogre::Pass* pass0 = technique->getPass(0);
+      Ogre::Pass* passT = technique->getPass(technique->getNumPasses()-1);
+      if (tinting)
+      {
+        // modify material's original color to use given alpha value
+        Ogre::ColourValue color = pass0->getDiffuse();
+        color.a = a;
```
iche033: we faced a problem in gazebo before in which the mesh's material is also semi-transparent. We ended up just combining the alpha values, e.g. `color.a *= a`, so the alpha effect is multiplied. Overriding the value is also fine as long as the user knows what to expect

rhaschke: @iche033, thanks for reviewing. 
Multiplying the values is dangerous as multiple updates of the marker's alpha value will continuously decrease the effective alpha value. (We don't have a memory of the original alpha value of the mesh's material.)
To allow to increase the alpha value again, we need to overwrite as proposed.
Hence, the first time the user provides a custom alpha value via the marker's colors, this will take precedence over the mesh's alpha values.

Eberty:
@rhaschke Is it possible to do the same thing in kinect devel? I saw that this bug fix was made for melodic-devel only. If you can do that, I will be grateful.

rhaschke:
I do not have resources for this anytime soon. I suggest building melodic-devel from source.

PR #1425, opened 2019-09-20
ImageDisplay: report frame issues

rhaschke:
Currently, image displays report "no image received" - also in case images were received but had an unknown image frame. This PR adds a corresponding failure handling for TF frames.

However, for `ImageDisplay` an issue remains with `CameraInfo`: If the camera info is published latched, rviz only receives a single message (either success or failure depending on the frame status). If rviz' fixed frame changes (which probably changes the frame status of the camera as well), this change is not noticed anymore, because updates can only happen when a new TF message is received.

@wjwwood, hence my question: Should we latched CameraInfo messages be supported?
If so, we would need to memorize the latest received message in any case and validate its frame "manually" as soon as the fixed frame changes.

### DIFF
--- a/src/rviz/default_plugin/camera_display.cpp
+++ b/src/rviz/default_plugin/camera_display.cpp
@@ -95,7 +95,7 @@ CameraDisplay::CameraDisplay()
   , render_panel_( 0 )
   , caminfo_tf_filter_( nullptr )
   , new_caminfo_( false )
-  , caminfo_ok_(false)
+  , caminfo_ok_( false )
   , force_render_( false )
 {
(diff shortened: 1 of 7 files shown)


wjwwood:
> Should we latched CameraInfo messages be supported?

Assuming you mean "should we support them", then I'd say, without looking into it much, yes. It seems like something that doesn't change often and therefore is a prime candidate for latching.

> If so, we would need to memorize the latest received message in any case and validate its frame "manually" as soon as the fixed frame changes.

That seems like a reasonable thing to do. It would be nice if there was a "on fixed frame changed" callback that displays could implement. Then the `ImageDisplay` could just implement that and react appropriately when it occurs.

PR #1427, opened 2019-09-20
Fix point cloud renderable bounding radius

c-andy-martin:
### Description

The PointCloudRenderable returns an incorrect bounding radius when the furthest corner of the box is not the minimum or maximum corner.

This results in incorrect culling of point cloud renderables in certain viewing angles when zooming in on the cloud.

Fix this by finding the largest absolute value of each coordinate to find the furthest corner from the origin. Then return the distance to the origin as the bounding radius.

### DIFF
--- a/src/rviz/ogre_helpers/point_cloud.cpp
+++ b/src/rviz/ogre_helpers/point_cloud.cpp
@@ -849,7 +849,21 @@ void PointCloudRenderable::_notifyCurrentCamera(Ogre::Camera* camera)
 
 Ogre::Real PointCloudRenderable::getBoundingRadius(void) const
 {
-  return Ogre::Math::Sqrt(std::max(mBox.getMaximum().squaredLength(), mBox.getMinimum().squaredLength()));
+  // Find the furthest point on the bounding box from the origin.
+  // That will define our bounding radius.
+  Ogre::Vector3 vMin, vMax, vFurthest;
+  vMin = mBox.getMinimum();
+  vMax = mBox.getMaximum();
+  vMin.x = std::abs(vMin.x);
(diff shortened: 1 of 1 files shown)


rhaschke:
This PR doesn't make sense to me: it mixes x,y,z components from the minimum and maximum corner of the aabb into the furthest point.
The key property of an axis-aligned bounding box is that it encloses all points. Hence, the minimum or maximum corner are furthest away from the origin of the point cloud.

If you observe issues with culling, please provide an example point cloud (marker) message, e.g. as a yaml or rosbag file, and a rviz config defining the camera pose. Then, I can try to reproduce the issue and look into it.

c-andy-martin:
I will work on that.

However, a simple 2D example shows that when the origin of the cloud is *outside* the bounding box (which is happening in the case I hit this), that the furthest corner is not the min or max of the bounding box:

```
._.
| |
._.


        . <== origin
```

In this case, the furthest point is the top-left point, which is neither the min or max. The old function would calculate the wrong bounding radius, and if you focus such that only that part of the cloud is visible, it disappears. I'll try to make a simple case and send it.

rhaschke:
I was assuming that this is related to the pose of the origin. Thanks for clarifying. If you could provide an example point cloud, that would be great.

c-andy-martin:
After many tried to reproduce this issue, I have discovered that I can not force this bug to manifest. I suspect that Ogre is always using the bounding box, and not this radius, for culling. I have discovered the actual root of my issue, which is a bug in the octomap rviz plugin. I am closing this PR.

rhaschke:
Thanks for investigating.

Issue #1175, opened 2017-12-04
replace RenderPanel::sendMouseMoveEvent by using QWidget::setMouseTracking


simonschmeisser:
https://github.com/ros-visualization/rviz/blob/fa59398651327241e9c6d72019e5fa7731ea0b6e/src/rviz/render_panel.cpp#L91

uses a timer (33ms!) to fake receiving mouse move events when there is no button pressed (for interactive markers for example). Is there a reason why we are not setting QWidget::setMouseTracking
http://doc.qt.io/qt-5/qwidget.html#mouseTracking-prop ?

I'll open a PR later if no one remembers why it is done this way

dhood:
At first I wondered it if was a new addition to qt but it looks like it was available in at the time [of the commit](https://github.com/ros-visualization/rviz/commit/140d43b76d29469e33ac6c4819bb6caf1bc370c5). This approach might be appropriate if we were trying to limit the callback frequency or something, but at 30Hz it doesn't look like the case 😄. So it seems safe to me to swap to using `setMouseTracking(true)` (unless @hersh happens to remember a reason why it wasn't appropriate, but we're going back quite some years now).

Perhaps the reason will reveal itself if/when you give it a shot @simonschmeisser 😉

rhaschke:
@simonschmeisser, could you provide a PR for this please?

PR #1433, opened 2019-09-27
Mouse move

simonschmeisser:
This addresses #1175 

enable mouse tracking for render_panel

when mouseTracking is enabled for a QWidget it will get mouseMoveEvents even though no mouse button is pressed. This behavior was faked in rviz with a 30Hz timer.

This is currently preserving API/ABI but could just as well be non preserving and go to noetic only. I did not observe any positive or negative impact from this change, just cleaner code (eventually).


rhaschke:
Thanks a lot for this cleanup, which also fixes a nasty issue: If accidentally entering the RenderPanel with pressed mouse button, the camera was screwed...
Now, such mouse events are not passed to the RenderPanel anymore (because they are correctly caught by the original QWidget).

PR #1434, opened 2019-10-04
Fix compatibility with Ogre 1.10, 1.11, 1.12

rhaschke:
This is a cleanup of #1242. Most work was already done by @de-vri-es, but some issues were remaining. The main API changes in Ogre are:
* The removal of `SimpleRenderable::setMaterial(const std::string & name)`.
* A _strict_ group handling in resource manager, which avoids searching resources in all available groups. Additionally, methods like `resourceExists(const std::string & name)` and `remove(const std::string & name)` expect the _correct_ group name now (but allow for a default).
To maintain compatibility with Ogre 1.9, which misses those group arguments, I had to push all resources into the default group.

Further, this PR:
* adds a cross-version compatibility function for setting materials on a `SimpleRenderable`
* loads the `freeimage` codec plugin, required to load images (and textures)
* avoids the use of relative paths in shader programs because they break in Ogre 1.11
* prevents removing anonymous scene nodes by name
* Uses OgrePrequisites.h to avoid conflict of class-forward vs. typedef for `Ogre::Vector3`
* recursively adds `ogre_media/fonts` as a resource location

As this is a huge PR with several potential side effects, it is only targeted for Noetic, but I successfully ran `marker_test`, `mesh_marker_test` and tested selection.

For compatibility with Ogre 1.9, we need to rely on now deprecated Ogre API (essentially Ogre's `SharedPtr` was replaced by `std::shared_ptr`), which is why I disabled corresponding warnings.
If a newer Ogre version is (hopefully) released into Ubuntu 20.04, this needs to be cleaned further.

### DIFF
--- a/src/python_bindings/sip/CMakeLists.txt
+++ b/src/python_bindings/sip/CMakeLists.txt
@@ -42,7 +42,12 @@ find_package(python_qt_binding REQUIRED)
 include(${python_qt_binding_EXTRAS_DIR}/sip_helper.cmake)
 
 # maintain context for different named target
-set(rviz_sip_INCLUDE_DIRS ${rviz_INCLUDE_DIRS} "${PROJECT_SOURCE_DIR}/src" ${catkin_INCLUDE_DIRS} $<TARGET_PROPERTY:rviz,EXPORT_HEADER_DIR>)
+set(rviz_sip_INCLUDE_DIRS
+  "${PROJECT_SOURCE_DIR}/src"
+  $<TARGET_PROPERTY:rviz,EXPORT_HEADER_DIR>
+  ${OGRE_OV_INCLUDE_DIRS}
+  ${catkin_INCLUDE_DIRS}
+)
 set(rviz_sip_LIBRARIES ${rviz_LIBRARIES} ${PROJECT_NAME})
 set(rviz_sip_LIBRARY_DIRS ${rviz_LIBRARY_DIRS} ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_LIB_DESTINATION})
 if (MSVC)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,12 @@ if(NOT OGRE_OV_FOUND)
   pkg_check_modules(OGRE_OV REQUIRED OGRE)
 endif(NOT OGRE_OV_FOUND)
 
+# TODO: adapt version for Noetic release
+# Disable deprecation warnings for OGRE >= 1.10
+if(NOT OGRE_OV_OGRE_VERSION VERSION_LESS "1.10.0")
+  add_compile_options(-Wno-deprecated-declarations)
+endif()
+
 ## Find OGRE Plugin path (not necessarily platform-independent, I guess)
 if(NOT DEFINED OGRE_PLUGIN_PATH)
   execute_process(COMMAND
--- a/ogre_media/fonts/liberation_sans.fontdef
+++ b/ogre_media/fonts/liberation_sans.fontdef
@@ -5,3 +5,12 @@ Liberation Sans
   size 18
   resolution 96
 }
+
+# Ogre >= 1.10 has changed the format
+font "Liberation Sans"
+{
+  type truetype
+  source liberation-sans/LiberationSans-Regular.ttf
+  size 18
+  resolution 96
+}
--- a/ogre_media/materials/glsl120/glsl120.program
+++ b/ogre_media/materials/glsl120/glsl120.program
@@ -1,8 +1,8 @@
 
 //includes:
-fragment_program rviz/glsl120/include/circle_impl.frag glsl { source include/circle_impl.frag }
-fragment_program rviz/glsl120/include/pack_depth.frag glsl { source include/pack_depth.frag }
-vertex_program rviz/glsl120/include/pass_depth.vert glsl { source include/pass_depth.vert }
+fragment_program rviz/glsl120/include/circle_impl.frag glsl { source circle_impl.frag }
+fragment_program rviz/glsl120/include/pack_depth.frag glsl { source pack_depth.frag }
+vertex_program rviz/glsl120/include/pass_depth.vert glsl { source pass_depth.vert }
 
 //all shaders, sorted by name
 
--- a/ogre_media/materials/glsl120/nogp/nogp.program
+++ b/ogre_media/materials/glsl120/nogp/nogp.program
@@ -4,7 +4,7 @@
 // and each one is offset according to its texture coords.
 
 //includes:
-vertex_program rviz/glsl120/nogp/pass_depth.vert glsl { source ../include/pass_depth.vert }
+vertex_program rviz/glsl120/nogp/pass_depth.vert glsl { source pass_depth.vert }
 
 vertex_program rviz/glsl120/nogp/billboard_tile.vert glsl
 {
--- a/src/image_view/image_view.cpp
+++ b/src/image_view/image_view.cpp
@@ -33,6 +33,7 @@
 
 #include "rviz/ogre_helpers/qt_ogre_render_window.h"
 #include "rviz/ogre_helpers/initialization.h"
+#include "rviz/ogre_helpers/compatibility.h"
 #include "rviz/image/ros_image_texture.h"
 
 #include "ros/ros.h"
@@ -114,7 +115,7 @@ void ImageView::showEvent( QShowEvent* event )
 
   Ogre::Rectangle2D* rect = new Ogre::Rectangle2D(true);
   rect->setCorners(-1.0f, 1.0f, 1.0f, -1.0f);
-  rect->setMaterial(material->getName());
+  setMaterial(*rect, material);
   rect->setRenderQueueGroup(Ogre::RENDER_QUEUE_OVERLAY - 1);
   Ogre::AxisAlignedBox aabb;
   aabb.setInfinite();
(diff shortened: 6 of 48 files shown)


rhaschke:
When I will get positive feedback from @de-vri-es and @ProfFan, I will fast-forward noetic-devel.

simonschmeisser:
I successfully build ogre 1.12.2 as an ubuntu deb package based on the 1.9.0 debian package and will test this branch with it in the next days and then try to get it into debian testing which would then lead to it being included in ubuntu 20.04 automagically. It's my first attempt at trying to contribute to ubuntu/debian so it will likely take some time. If my initial testing with this branch is successful I'll also create a PPA so you can test the packaging/branch as well.

de-vri-es:
I'm not currently in the best position to test this. @hgaiser: Can you test it, or find someone else at Fizyr to test it?

hgaiser:
I'll see if I can test this tomorrow.

simonschmeisser:
I got the following error when trying to compile with my new ogre-1.12 package
```
In file included from panel_dock_widget.sip:7:0:
/home/simon/proggn/isys/ws/src/rviz/src/rviz/panel_dock_widget.h:34:10: fatal error: rviz/rviz_export.h: No such file or directory
 #include "rviz/rviz_export.h"
          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [siplibrviz_sipcmodule.o] Error 1
```
Did you see that before?

rhaschke:
@simonschmeisser, this issue isn't particularly related to Ogre. Obviously this dynamically created header file, isn't in the include path for the sip python wrapper. I haven't seen this issue before.
Could you please check, whether and where this file is located - in build, devel, or install space?

src/python_bindings/sip/CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ endif()
 if(sip_helper_FOUND)
   list(APPEND rviz_BINDINGS "sip")
   set(rviz_BINDINGS "${rviz_BINDINGS}" PARENT_SCOPE)
+  get_property(rviz_sip_INCLUDE_DIRS DIRECTORY PROPERTY INCLUDE_DIRECTORIES)
```
simonschmeisser:
```suggestion
  get_property(dir_includes DIRECTORY PROPERTY INCLUDE_DIRECTORIES)
  set(rviz_sip_INCLUDE_DIRS ${dir_includes} ${rviz_sip_INCLUDE_DIRS})
```

this combines both the manually crafted list from above as well as the one that is set for the directory (which probably adds the build directory where rviz_export.h is found)

simonschmeisser:
Next error is this:
```
g++ -c -std=c++14 -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -O2 -Wall -W -D_REENTRANT -DNDEBUG -DQT_NO_DEBUG -DQT_CORE_LIB -DQT_GUI_LIB -I. -I/usr/lib/include -I/usr/include/eigen3 -I/usr/include/OGRE -I/usr/include/OGRE/Overlay -I/usr/include -I/usr/include/python2.7 -I/home/simon/proggn/isys/ws/src/rviz/src -I/opt/ros/melodic/include -I/opt/ros/melodic/share/xmlrpcpp/cmake/../../../include/xmlrpcpp -I/home/simon/proggn/isys/ws/build/rviz/src/rviz -I/usr/share/qt5/mkspecs/linux-g++ -I/usr/include/x86_64-linux-gnu/qt5/QtCore -I/usr/include/x86_64-linux-gnu/qt5/QtGui -I/usr/include/x86_64-linux-gnu/qt5/QtWidgets -I/usr/include/x86_64-linux-gnu/qt5/QtPrintSupport -I/usr/include/x86_64-linux-gnu/qt5 -I/usr/X11R6/include -o siplibrviz_siprvizBoolProperty.o siplibrviz_siprvizBoolProperty.cpp
/home/simon/proggn/isys/ws/build/rviz/sip/rviz_sip/siplibrviz_siprvizBoolProperty.cpp:142:10: error: 'bool siprviz_BoolProperty::getReadOnly()' marked 'override', but does not override
     bool getReadOnly() SIP_OVERRIDE;
          ^~~~~~~~~~~
Makefile:18: recipe for target 'siplibrviz_siprvizBoolProperty.o' failed
```
which might be due to getReadOnly being const in the actual code but not in the sip definitions

simonschmeisser:
success!
```
[ INFO] [1570695804.185048816]: rviz version 1.13.5
[ INFO] [1570695804.185090094]: compiled against Qt version 5.13.1
[ INFO] [1570695804.185101978]: compiled against OGRE version 1.12.2 (Rhagorthua)
[ INFO] [1570695804.193143098]: Forcing OpenGl version 0.
[ INFO] [1570695804.332504804]: Stereo is NOT SUPPORTED
[ INFO] [1570695804.332576435]: OpenGl version: 3 (GLSL 1.3).
[ERROR] [1570695804.355941864]: Error: ScriptCompiler - unexpected token in liberation_sans.fontdef(1): 'Liberation'. If this is a legacy script you must prepend the type (e.g. font, overlay).
```

simonschmeisser:
the future looks dark :(
grafik
`
roslaunch fanuc_cr7ia_support test_cr7ia.launch
`
from here: https://github.com/ros-industrial/fanuc 

meshes are stl files

hgaiser:
I followed the same path as @simonschmeisser , but used a different (private) package for the robot model. For me it did visualize the meshes properly.

Other than this I haven't tried much in rviz, but it seems to work from what I can tell, with the exception of the error which is also logged by @simonschmeisser but not mentioned:

```
[ERROR] [1570695804.355941864]: Error: ScriptCompiler - unexpected token in liberation_sans.fontdef(1): 'Liberation'. If this is a legacy script you must prepend the type (e.g. font, overlay).
```

de-vri-es:
Note the font thing was fixed somewhere too. I think the font name simply needed quotes around it in the script. That may have been lost in the PR migration.

See this commit by @ProfFan:
https://github.com/fizyr-forks/rviz/pull/1/commits/48e6dd22e9dccd4ccf89cd7c7021bbbb2f64db73

simonschmeisser:
> I followed the same path as @simonschmeisser , but used a different (private) package for the robot model. For me it did visualize the meshes properly.

I guess your model uses collada/DAE meshes instead of plain STL ones. Their material handling and therefore lighting is handled separately, we fixed that a couple of times recently so I guess @rhaschke knows where to look :)

hgaiser:
> I guess your model uses collada/DAE meshes instead of plain STL ones. Their material handling and therefore lighting is handled separately, we fixed that a couple of times recently so I guess @rhaschke knows where to look :)

Indeed, it uses DAE :)

rhaschke:
> > Error: ScriptCompiler - unexpected token in liberation_sans.fontdef(1): 'Liberation'. If this is a legacy script you must prepend the type (e.g. font, overlay).

> Note the font thing was fixed somewhere too. I think the font name simply needed quotes around it in the script. That may have been lost in the PR migration.

The key issue here is that the syntax of font definition files has changed: Newer versions require the "font" type to be mentioned, while older version cannot handle this.
To have the same rviz code working with old (1.9) and new Ogre versions, I added both, the old and the new syntax to the font definition file, which causes this warning message.

hgaiser:
If it should be considered a warning, can it be changed to be printed as a warning?

rhaschke:
@simonschmeisser, I cannot reproduce your issue https://github.com/ros-visualization/rviz/pull/1434#issuecomment-540460497 with Ogre 1.11.6. The rendering looks fine:
image

rhaschke:
Same result for Ogre 1.12.2. I'm afraid, @simonschmeisser, you need to debug yourself.

rhaschke:
I cleaned up the commit history.

simonschmeisser:
Ah, it might be possible that this the good old locale issue again ... was doing the tests on my private notebook which might not have the fixes installed? Will check at some point during the weekend

rhaschke:
That would be an explanation indeed. Awaiting your feedback.

simonschmeisser:
Yes, sorry for the confusion, `LANG=c` did the trick and it now looks normal. :+1:

rhaschke:
Hi @simonschmeisser,
in https://github.com/ros-visualization/rviz/pull/1434#issuecomment-538784504 you mentioned that you were working on a .deb package to get a newer Ogre version into Ubuntu 20.04 / ROS Noetic. Did you make progress with this?
The 20.04 pre-release I installed at the weekend didn't include a newer Ogre yet.

simonschmeisser:
It's still WIP but it seems we're close I'm told. Basically there are still at least two warnings from the Debian linter that should be gone before uploading it to the Debian incoming queue. You can check out the current state at

https://salsa.debian.org/games-team/ogre/tree/ogre-1.12

simonschmeisser:
@rhaschke I just tried to compile noetic-devel but it failed, your personal branch works, could you check what's happening?

ogre-1.12 has been sitting in the Debian NEW-queue for almost three weeks now, I hope it will be handled soon!

rhaschke:
Looks like, I missed to forward the noetic branch. Sorry for that. I forwarded the branch and verified that it compiles. There are some open issues before a release though. For example, the initial size of the RenderWidget is not correctly recognized anymore.

simonschmeisser:
Thanks, yes I noticed that as well. I had the impression that was on 1.12 only and not on 1.9? Will recheck.

simonschmeisser:
finally ... https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=ogre-1.12

Issue #1380, opened 2019-06-16
Show further properties for selected Markers


B0gdar:
## 1. Brief description
When a Marker is being selected with the selection tools just [shows position and orientation values](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/default_plugin/markers/marker_selection_handler.cpp#L71-L83).

* **Cube**
![imagen](https://user-images.githubusercontent.com/9785667/59563386-a65a9280-9039-11e9-81b0-fd7430f0d492.png)
* **Sphere**
![imagen](https://user-images.githubusercontent.com/9785667/59563592-4c0f0100-903c-11e9-9845-9bfe37919bd2.png)
* **Cylinder**
![imagen](https://user-images.githubusercontent.com/9785667/59563608-782a8200-903c-11e9-9be1-70a98cda999f.png)

## 2. Explanation

When a Marker is being selected with the selection tools just [shows position and orientation values](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/default_plugin/markers/marker_selection_handler.cpp#L71-L83).

In general is enough, but when inspecting volumes or shapes it would be nicer to being able to straightforward inspect the scale as well (p.e. Bounding box fitting applications). Current state just make a menu with entries for getPosition() and getOrientation(). This could be expanded based on marker type and show additional info.

Changes needed to introduce this feature are small, I am not so sure about possible impact with interactive markers if any. I would try to make the changes myself when I find some spot, but everybody is welcome to perform them.

**I would also like to encourage other ROS users that do not know the selection tool to actively make use of it, as it saves time with respect the usage of other ROS packages or external tools.**

 The selection tool is simple but I strongly believe that if a broader audience will make use of the tool, cooler features would be able to be introduced 😄

## 3. Which distros are affected
All distros from indigo, as this tool was introduced long time ago.

## 4. To reproduce the issue

Reproduction is simple, just publish a marker with rostopic (CUBE, type:1; SPHERE, type:2; CYLINDER, type:3):
```bash
rostopic pub /shape visualization_msgs/Marker "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: 'map'
ns: 'MarkerWithShape'
id: 0
type: 1
action: 0
pose:
  position: {x: 0.0, y: 0.0, z: 0.0}
  orientation: {x: 0.0, y: 0.0, z: 0.0, w: 1.0}
[...]

PR #1435, opened 2019-10-08
Add scale and color properties to Marker selection pane

robustify:
This PR addresses #1380 and adds a scale and color property to the marker selection pane to provide information that may be useful for different use cases. The strings for the x, y, and z scale properties reflect the meaning of the scale values depending on the type of marker being displayed.

Here is what the selection pane looks like now after selecting a cube, arrow, and text marker with just the existing position and orientation properties:

before

... and here is what displays after building this PR with the new scale and color properties:

after


rhaschke:
Thanks for this valuable contribution. Just looking at the screenshots, I propose the following changes:
- display the marker type in the value field of the top-level nodes
- use x,y,z for dimension names of cubes
- skip the `Arrow` prefix in the arrow dimension titles (Arrow will be mentioned in the top-level node)
- use `rviz::ColorProperty` for colors

Only later, I will have time to review the PR in more detail.

robustify:
Thanks for the feedback! I think those are good suggestions and I'll update the PR when I get some time soon.

The ABI check is currently failing because of the new methods I added to `MarkerSelectionHandler`. I will also try to re-organize the changes to not add any new methods and not break the ABI.

robustify:
I just realized that is impossible to add new property displays without failing the Data Typ ABI checks. Closing this PR and opening a new one targeting noetic-devel instead.

PR #1436, opened 2019-10-09
Add scale and color properties to Marker selection

robustify:
This PR addresses issue #1380 and adds some extra functionality to the marker selection pane.

This PR started from #1435 and was intended to target `melodic-devel`, but it wasn't possible to make the necessary changes and not break the ABI. Therefore, this PR targets `noetic-devel` instead. Here is an updated screenshot with some changes made since the initial PR was opened:

Before modifications:
before

After modifications:
after


src/rviz/default_plugin/markers/marker_selection_handler.cpp
```diff
@@ -85,53 +85,49 @@ QColor MarkerSelectionHandler::getColor()
                  (int)(marker_->getMessage()->color.a * 255) );
 }
 
-QString MarkerSelectionHandler::getMarkerTypeName()
```
rhaschke: I moved `getMarkerTypeName` to `marker_utils.cpp`.

src/rviz/default_plugin/markers/marker_selection_handler.cpp
```diff
@@ -85,53 +85,49 @@ QColor MarkerSelectionHandler::getColor()
...
   }
-  return marker_type_name;
+  return p;
+}
+
 }
 
 void MarkerSelectionHandler::createProperties( const Picked& obj, Property* parent_property )
 {
-  Property* group = new Property( "Marker " + marker_id_ + " (" + getMarkerTypeName() + ")", QVariant(), "", parent_property );
```
rhaschke: Here, my intention was to use the type name in the value field. Minor change.

src/rviz/default_plugin/markers/marker_selection_handler.cpp
```diff
@@ -140,7 +136,7 @@ void MarkerSelectionHandler::createProperties( const Picked& obj, Property* pare
   orientation_property_ = new QuaternionProperty( "Orientation", getOrientation(), "", group );
   orientation_property_->setReadOnly( true );
 
-  scale_property_ = new MarkerScaleProperty( "Scale", getScale(), marker_->getMessage()->type, "", group );
```
rhaschke: There is no need to duplicate code for a new `MarkerScaleProperty`, which actually behaves like a VectorProperty.
It's possible to rename (and hide) the fields.

PR #1445, opened 2019-11-20
fix yaml-cpp not found on macOS (catalina)

ahoarau:
This fixes the issue : 
```
ld: library not found for -lyaml-cpp
clang: error: linker command failed with exit code 1 (use -v to see invocation)
````
### Description

pkg config **succeeds** in finding yaml but linker fails on macOS (tested on catalina)
Now on newer version of yamlcpp, the cmake config is installed.

### DIFF
--- a/src/rviz/CMakeLists.txt
+++ b/src/rviz/CMakeLists.txt
@@ -144,7 +144,7 @@ target_link_libraries(${PROJECT_NAME}
   ${TinyXML2_LIBRARIES}
   ${X11_X11_LIB}
   ${ASSIMP_LIBRARIES}
-  ${yaml-cpp_LIBRARIES}
+  ${YAML_CPP_LIBRARIES}
 )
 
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/rviz/CMakeLists.txt [hunk 1]
```diff
@@ -144,7 +144,7 @@ target_link_libraries(${PROJECT_NAME}
   ${TinyXML2_LIBRARIES}
   ${X11_X11_LIB}
   ${ASSIMP_LIBRARIES}
-  ${YAMLCPP_LIBRARIES}
+  ${yaml-cpp_LIBRARIES}
```
rhaschke: The variable is named YAML_CPP_LIBRARIES on Ubuntu 18.04. Is it different on Mac?

ahoarau: My mistake. fixed !

rhaschke:
Thanks a lot.

PR #1446, opened 2019-11-20
generate configure headers in devel workspace

v4hn:
Otherwise rviz cannot be built with its dependees together in one devel workspace.


v4hn:
Thanks for the simplifications and merging, CI on the merged commit seems to have passed, but CI for the request failed?

Building your changes with my `catkin build` workspace produces the same error over here:

```
Errors     << rviz:cmake /homeL/pushing/ros/logs/rviz/build.cmake.000.log
CMake Error: Could not open file for write in copy operation /rviz/default_plugin/rviz_default_plugin_export.h.tmp
CMake Error: : System Error: No such file or directory
CMake Error at /usr/share/cmake-3.5/Modules/GenerateExportHeader.cmake:362 (configure_file):
  configure_file Problem configuring file
  /usr/share/cmake-3.5/Modules/GenerateExportHeader.cmake:378 (_do_generate_export_header)
  src/rviz/default_plugin/CMakeLists.txt:76 (generate_export_header)
```

rhaschke:
The variable `EXPORT_HEADER_DIR` was defined too late... I force-pushed a fix.
Travis didn't fail because it runs with root permissions (inside docker) and thus is able to create those files anywhere - also in `/`.

CMakeLists.txt
```diff
@@ -177,7 +177,6 @@ catkin_package(
   CFG_EXTRAS
     rviz-extras.cmake
   INCLUDE_DIRS
-    src
```
v4hn: removing this exported include dir also broke the devel workspace with other dependencies.
RViz does not cleanly split its public headers into `include/` and, e.g., `#include <rviz/frame_manager.h>` will find the system headers and not this one.

rhaschke: We still have https://github.com/ros-visualization/rviz/blob/melodic-devel/CMakeLists.txt#L215,
which should define the correct include folder. Which concrete error message to you see? What's the compiler command line?

rhaschke: I missed to include EXPORT_HEADER_DIR though. WIP: https://github.com/ros-visualization/rviz/pull/1448

v4hn: I'm building moveit `master` on ROS kinetic and see this error because the includes only find the `/opt/ros/kinetic` header:
```
Errors     << moveit_ros_visualization:make /homeL/pushing/ros/logs/moveit_ros_visualization/build.make.001.log
/homeL/pushing/ros/src/upstream/moveit/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp: In member function ‘virtual planning_scene_monitor::PlanningSceneMonitorPtr moveit_rviz_plugin::PlanningSceneDisplay::createPlanningSceneMonitor()’:
/homeL/pushing/ros/src/upstream/moveit/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp:491:77: error: ‘class rviz::FrameManager’ has no member named ‘getTF2BufferPtr’
   std::shared_ptr<tf2_ros::Buffer> tf_buffer = context_->getFrameManager()->getTF2BufferPtr();
                                                                             ^
```

No problem with the exported `src` include in place.

PR #1448, opened 2019-11-21
add missing include folder

rhaschke:


rhaschke:
Fix squashed into offending commit #1446.

PR #1449, opened 2019-11-21
Fixup cmake file issues

rhaschke:
Dear @v4hn,
this is a new attempt for #1446, which I had to revert due to open issues.
For me, creating the export header either in devel or build space, didn't change anything:
The header was not found by downstream packages in the same catkin devel workspace.
To resolve this, I had to add the corresponding include paths to `catkin_package()`, which makes `catkin_lint` complaining. So, I'm not yet satisfied...


v4hn:
> For me, creating the export header either in devel or build space, didn't change anything:
The header was not found by downstream packages in the same catkin devel workspace.

Yes, of course. This mixes up two semi-independent issues.
- devel workspaces should not depend on (generated) files in build. To fix this the headers have to be generated in devel
- the path to the headers, no matter where they are generated, has to be listed in `catkin_package`, as you say.

> To resolve this, I had to add the corresponding include paths to catkin_package(),

That's the second point, yes.

> which makes catkin_lint complaining. So, I'm not yet satisfied...

This looks like a catkin_lint shortcoming to me.
After all, catkin has to have *some* way to generate headers at build time also for the devel workspace (and genmsg does just that all the time).

I just tested your patch with success a moment ago, but I just got a notification that you merged already anyway. ;)

v4hn:
Thanks for your time!

rhaschke:
Yes, I marked the catkin_lint error to be ignored and merged.

PR #1454, opened 2019-11-29
Update bindings imports for Py3-friendliness

mikepurvis:
My rviz installspace looks like this:

```
$ find python3/dist-packages/rviz
python3/dist-packages/rviz
python3/dist-packages/rviz/__init__.py
python3/dist-packages/rviz/librviz_sip.so
python3/dist-packages/rviz/__pycache__
python3/dist-packages/rviz/__pycache__/__init__.cpython-35.pyc
```

To successfully import the bindings lib from `__init__.py`, it needs to be relative. Can confirm that the shiboken library installs to the same place:

https://github.com/ros-visualization/rviz/blob/b9bff9329052764dd7d614688e1310f09da68163/src/python_bindings/shiboken/CMakeLists.txt#L96-L102


src/python_bindings/rviz/__init__.py
```diff
@@ -15,7 +15,7 @@
 
     # Import the shared library generated by the shiboken binding-generator.
     # This depends on rviz/manifest.xml listing ${prefix}/lib in the python path.
```
rhaschke: I guess we can get rid of this comment?

src/python_bindings/rviz/__init__.py
```diff
@@ -24,7 +24,7 @@
 
     # Import the shared library generated by the sip binding-generator.
     # This depends on rviz/manifest.xml listing ${prefix}/lib in the python path.
```
rhaschke: Same here.

mikepurvis:
My workspace is dual python 2/3 and the proposed change works for both.

PR #1457, opened 2019-12-16
Add display: Add button to append topic name after display name

VictorLamoine:
## Add display widget before modifications

Screenshot_20191216_162629

### Description

I have added a button (bottom of the widget) when adding a display type. This button (when checked) appends the name of the topic after the name of the display.

Screenshot_20191216_161708

Screenshot_20191216_161726

This allows to easily get the topic name in the display name:

Screenshot_20191216_162110

The display name will not be updated if the topic changes and the user is warned about that in the tooltip:
Screenshot_20191216_161743

This feature is meant to be an easy way to add the topic name without having to type it; not a way to automatically display the topic name in a display widget. It is very useful to me when adding a lot of image topics; it saves a lot of typing.

### Checklist

### DIFF
--- a/src/rviz/add_display_dialog.cpp
+++ b/src/rviz/add_display_dialog.cpp
@@ -170,6 +170,7 @@ AddDisplayDialog::AddDisplayDialog( DisplayFactory* factory,
                                     const QStringList& disallowed_class_lookup_names,
                                     QString* lookup_name_output,
                                     QString* display_name_output,
+                                    bool* append_topic_name_output,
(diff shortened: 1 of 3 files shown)


rhaschke:
I'm not fully convinced about the implementation: 
- The extra checkbox should only show up in the _"By Topic"_ subwidget, becaue only there it makes sense.
- The resulting display name should be shown directly in the dialog. This will allow to keep API and ABI and even merge in melodic as requested.

Issue #1461, opened 2020-01-09
Smooth rendering of spheres


wxmerkt:
Hi!
I'd like to add support to display spheres with smoothing. I am using the OGRE built-in `PT_SPHERE` `PrefabType` to create a smooth sphere. However, unlike the `rviz_sphere.mesh` (radius 0.5), these are 100 units in radius by default. Using a scaling of `0.01`, I can get the desired behaviour for the marker array plug-in and robot link. Cf. example below:

### Before
![rviz_screenshot_2020_01_09-17_46_02](https://user-images.githubusercontent.com/1664508/72093698-d4246300-330c-11ea-8e5b-ee1abd6f9f4d.png)

### After
![rviz_screenshot_2020_01_09-17_49_14](https://user-images.githubusercontent.com/1664508/72093712-d8e91700-330c-11ea-9e50-a24009543039.png)

### Question
The issue for which I am looking input is how to add this persistently for all plug-ins that may create a `rviz::Shape::Sphere`. If I were to set the scale in `createEntity`, it can easily be overridden by a call to `setScale` from anywhere else. Obviously, I do not want to simply change the convention to avoid breaking plug-ins.

1. Is there a good solution to this I may have overlooked?
2. I believe another solution would be to update the `rviz_sphere.mesh` file. Would this be acceptable?

Many thanks for your input,
Wolfgang

rhaschke:
I would suggest to update rviz_sphere.mesh, but maybe @wjwwood as another suggestion.

wjwwood:
I don't have a better suggestion. How would you change the mesh file?

wxmerkt:
@wjwwood please have a look at #1463 :-)

PR #1462, opened 2020-01-09
Fix for material_name = "" bug

wxmerkt:
Found this while debugging mesh loading - if the material name is empty, a segfault happened. This also defaults to the default material.

### DIFF
--- a/src/rviz/robot/robot_link.cpp
+++ b/src/rviz/robot/robot_link.cpp
@@ -628,7 +628,7 @@ void RobotLink::createEntityForGeometryElement(const urdf::LinkConstSharedPtr& l
       Ogre::SubEntity* sub = entity->getSubEntity(i);
       const std::string& material_name = sub->getMaterialName();
 
-      if (material_name == "BaseWhite" || material_name == "BaseWhiteNoLighting")
+      if (material_name == "BaseWhite" || material_name == "BaseWhiteNoLighting" || material_name.empty())
       {
         sub->setMaterial(default_material_);
       }


rhaschke:
The material shouldn't be empty actually, so I assume there is a deeper bug.
Can you provide a URDF that caused this issue?

wxmerkt:
Hi @rhaschke,
The way I triggered this bug was by switching to a `PreFabType` entity in `createEntity`. While I am now pursuing a new fix to #1461 by replacing the meshes, my first attempt is in this branch: 

https://github.com/ros-visualization/rviz/compare/melodic-devel...wxmerkt:wxm-smooth-spheres

To trigger the bug, I loaded a URDF that had a mesh as the visual and a sphere as the collision element. Without this check for `empty()`, we would trigger a segmentation fault using the above branch with the Fix commit reverted when it tried to `setMaterial` on the submesh.

rhaschke:
If I understand you correctly, the issue is only observed in your branch to solve #1461, but not in the current master. Hence, I think your branch doesn't set the materials correctly.
If you follow another solution now (modifying sphere.mesh), I think this PR becomes obsolete.

wxmerkt:
I am now following another solution, so this won't be triggered. However, with `material = ""`, the current master results in a segfault. Wouldn't be defaulting to the default material instead of a broken pointer resulting in a segmentation fault be a safer way to handle things?

rhaschke:
The segfault is an indicator that something is wrong with the material handling.
Catching this error will just hide the actual problem.

wxmerkt:
Okay, I understand your reasoning. I still believe that if we can trigger this by using OGRE native features (`PrefabType`), this may happen in user plug-ins or with malformed meshes/OGRE meshes. A segmentation fault here is an indicator of a pointer being broken - and the other way to fix it could be to catch that or check it.

rhaschke:
As you said: A segfault indicates a _malformed_ plugin (which doesn't comply to the rules of rviz)
and my reasoning is that this bug should be fixed there then.

wxmerkt:
I understand. I tend to catch/avoid/fix segfaults and present a helpful message/error/warning (which this PR did not do - but may consider a fix along those lines in the future). If a plugin can trigger undefined/uncaught behaviour, that's usually bad.

PR #1467, opened 2020-01-30
Catch exceptions when creating publishers in tools

### LINKED ISSUE #1466: [Bug] Crash upon entering an invalid topic name (ros::InvalidNameException)
When navigating to "2D Nav Goal" > "Topic" in "Tool Properties", RViz crashes immediately with the exception shown below when trying to enter invalid characters. IMHO, Rviz should show an error but it shouldn't crash completely. In my case, I entered `/goal[` (accidentally). Note that this happens in every edit field for "Topics" in the "Tool Properties".

I think the actual content of Rviz (rendered scene) doesn't matter for this.

### Your environment
* OS Version: Ubuntu 18.04 LTS 64 bit
* ROS Distro: Melodic, /rosversion: 1.14.3
* RViz, Qt, OGRE, OpenGl version as printed by rviz: v1.13.7, v5.9.5, v1.9.0 (Ghadamon), v4.6 (GLSL 4.6)

    ```
    terminate called after throwing an instance of 'ros::InvalidNameException'
      what():  Character [(] at element [1] is not valid in Graph Resource Name [/(suppy1/goal].  Valid     characters are a-z, A-Z, 0-9, / and _.
    [rviz-9] process has died [pid 2513, exit code -6, cmd /opt/ros/melodic/lib/rviz/rviz -d /tmp/rviz_config_generator_tmp_saMfAu.rviz __name:=rviz __log:=/.../.ros/log/04fe25dc-429a-11ea-    81b2-e0d55eac4766/rviz-9.log].
    log file: /.../.ros/log/04fe25dc-429a-11ea-81b2-e0d55eac4766/rviz-9*.log
    ```

rhaschke:
Fix #1466: An invalid ros topic name caused rviz to crash due to the uncaught exception.

### DIFF
--- a/src/rviz/default_plugin/tools/goal_tool.cpp
+++ b/src/rviz/default_plugin/tools/goal_tool.cpp
@@ -57,7 +57,11 @@ void GoalTool::onInitialize()
 
 void GoalTool::updateTopic()
 {
-  pub_ = nh_.advertise<geometry_msgs::PoseStamped>( topic_property_->getStdString(), 1 );
+  try {
+    pub_ = nh_.advertise<geometry_msgs::PoseStamped>( topic_property_->getStdString(), 1 );
+  } catch (const ros::Exception& e) {
+    ROS_ERROR_STREAM_NAMED("GoalTool", e.what());
+  }
 }
 
 void GoalTool::onPoseSet(double x, double y, double theta)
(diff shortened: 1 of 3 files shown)


rhaschke:
For now, this needs to be sufficient. I don't have the (time) resources to provide visual feedback now.
I thought about indicating an error via a red background in the input field. But this will involve ABI changes: properties need to have an additional boolean flag to indicate the error condition. Then `getViewData()` can return the corresponding background color. Probably, it would be great to provide the error message as well, e.g. as a tooltip.

PR #1475, opened 2020-03-05
Synchronous updates

simonschmeisser:
The api documentation of `rviz::Display` claims to provide a `ros::NodeHandle` whose associated callback queue is processed synchronously from the gui thread. Following this promise most subclasses of Display ignore thread-safety in their callback-functions. This promise however breaks if an `AsyncSpinner` is used in an application using the `rviz::RenderPanel` (or `ros::spinOnce` gets called in different threads)

This PR therefore adds a separate `CallbackQueue` for synchronous processing of message callbacks such that the promise continues to hold. Since not all Display subclasses are expected to use the designated `NodeHandle`s [1], a second commit adds a periodic `ros::spinOnce` in `rviz::VisualizerApp` that processes the global ros callback queue.

Note that the second commit requires Qt5, I can rewrite it using a member function if desired

1: https://github.com/ros-planning/moveit/blob/e8a4072727abe4c704d009f05baa764bcf106b33/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp#L300


src/rviz/visualization_manager.cpp
```diff
@@ -344,7 +345,7 @@ void VisualizationManager::onUpdate()
     resetTime();
   }
 
-  ros::spinOnce();
+  private_->update_queue_.callAvailable(ros::WallDuration(0.01));
```
rhaschke:
```suggestion
  private_->update_queue_.callAvailable(ros::WallDuration());
  ros::getGlobalCallbackQueue()->callAvailable(ros::WallDuration());
```

src/rviz/visualizer_app.cpp
```diff
@@ -221,6 +221,13 @@ bool VisualizerApp::init( int argc, char** argv )
     ros::NodeHandle private_nh("~");
     reload_shaders_service_ = private_nh.advertiseService("reload_shaders", reloadShaders);
 
+    //process any callbacks that don't use the designated callback queues of VisualizationManager
```
rhaschke: Why do you want to handle the global queue differently?

rhaschke:
@simonschmeisser, what is your feeling about this PR? Did you find an alternative solution?

src/rviz/visualizer_app.cpp
simonschmeisser: `VisualizerApp` is used by the rviz binary but not by library users. Handling the global queue here will make sure that plugins that rely on it won't see any difference. For library users it allows not handling the global queue in the main thread.

simonschmeisser:
I have to admit that I haven't worked on this the last three weeks. The issue is still open and I understand your point about there potentially being too many non-conformant plugins. This approach would allow fixing the set of plugins we use while not changing anything for others.

I will however also check the alternative you mentioned of using the AsyncSpinner for non-gui data handling and the global queue for GUI stuff only.

rhaschke:
@simonschmeisser, can we close this?

PR #1480, opened 2020-03-20
always forward keystrokes in display view to property editor

v4hn:
Current behavior without this patch:
- Open RViz with any display added
- click on the right column in the display view, in the same row as the checkbox to enable/disable the display
=> pressing spacebar can toggle the checkbox
- click on the left column of the same row (the name of the display)
=> pressing spacebar has no effect

Additionally, a visible indicator of the current focus (not the current selection)
highlights the corresponding cell.

This patch modifies selection requests to always focus the right column (if available)
and achieve the first behavior consistently.

This allows to go through a long list of displays by keyboard and toggle them with spacebar.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,6 +206,7 @@ catkin_package(
     roslib
     sensor_msgs
     std_msgs
+    std_srvs
     tf
     urdf
     visualization_msgs
--- a/src/rviz/properties/property_tree_widget.cpp
+++ b/src/rviz/properties/property_tree_widget.cpp
@@ -41,6 +41,21 @@
 namespace rviz
 {
 
+PropertySelectionModel::PropertySelectionModel(QAbstractItemModel* model)
+  : QItemSelectionModel(model)
+{}
+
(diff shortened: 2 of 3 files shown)


v4hn:
Sorry for the hiccup, I did not notice bionic's Qt5 version is that old already.
CI succeeds now.

v4hn:
@rhaschke CI complains that `https://ni.www.techfak.uni-bielefeld.de/abicheck/rviz-melodic.tar.gz` is not accessible (403 Forbidden).

rhaschke:
Hi @v4hn, sorry about the late response. I'm busy with another project these days.
I fixed the permission issue. Travis is checking again now.

v4hn:
> I'm busy with another project these days.

I noticed. :-)
It's sad how much of a difference that makes...

CI succeeds now and I use the patch locally.

rhaschke:
I hope to be more available later this week.
Regarding this PR, I suggest modifying the selection index only if the target is not editable.

v4hn:
> Regarding this PR, I suggest modifying the selection index only if the target is not editable.

I adapted the patch.

PR #1482, opened 2020-03-29
switch libogre-dev to build_depend

simonschmeisser:
this removes the runtime dependency on libogre-dev and therefore simplifies working with multiple ogre versions installed (once they make it into Debian/Ubuntu ...). It also reduces installation size by a few kb

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -20,6 +20,7 @@
 
   <build_depend>cmake_modules</build_depend>
   <build_depend>eigen</build_depend>
+  <build_depend>libogre-dev</build_depend>
   <build_depend>qtbase5-dev</build_depend>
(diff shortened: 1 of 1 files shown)


rhaschke:
I need to revert this PR, because there are several rviz include headers, including OGRE headers by themselves. Hence, OGRE headers are a transitive build dependency.

simonschmeisser:
so does it mean we need both `<build_depend>` and `<build_export_depend>` ?

rhaschke:
Yes.

simonschmeisser:
Should I open a new PR then?

rhaschke:
So far, I just reverted. New release request is out already.

PR #1483, opened 2020-03-30
fix: + include yaml cpp include directory

peter-lehner:
### Description

Added missing cmake command to include yaml-cpp include directory. My source build requires this include to complete correctly with yaml-cpp 0.6.2.

### Checklist

None apply

### DIFF
--- a/src/rviz/CMakeLists.txt
+++ b/src/rviz/CMakeLists.txt
@@ -1,5 +1,7 @@
 find_package(yaml-cpp REQUIRED)
 
+include_directories(${YAML_CPP_INCLUDE_DIR})
+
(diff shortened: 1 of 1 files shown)


src/rviz/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 find_package(yaml-cpp REQUIRED)
 
+include_directories(${YAML_CPP_INCLUDE_DIR})
```
rhaschke: Please mark them as system includes:
```suggestion
include_directories(SYSTEM ${YAML_CPP_INCLUDE_DIR})
```

PR #1488, opened 2020-04-05
Build depends

simonschmeisser:
This is a fixup of #1482 that unfortunately had to be reverted, it now also sets `build_export_depend`

Should ideally also be backported to melodic-devel

side-note: ogre-1.12.4 is now in Debian sid and there is still hope and an open request to get it into Ubuntu 20.04

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -20,6 +20,7 @@
 
   <build_depend>cmake_modules</build_depend>
   <build_depend>eigen</build_depend>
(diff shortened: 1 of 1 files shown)


rhaschke:
@simonschmeisser, unfortunately this PR didn't had the desired effect: the rviz binary still depends on libogre-dev in Noetic.

simonschmeisser:
yes, bloom is unfortunately not able to handle this currently

PR #1494, opened 2020-04-27
clang-tidy fixes

rhaschke:


rhaschke:
Merged despite ABI break (making LinkFactory's destructor virtual), because fixing the potential segfault is more important.

PR #1496, opened 2020-04-27
MeasureTool: hide measure line when deactivating the tool

rhaschke:
Otherwise the line will hang around virtually forever.

### DIFF
--- a/src/rviz/default_plugin/tools/measure_tool.cpp
+++ b/src/rviz/default_plugin/tools/measure_tool.cpp
@@ -74,6 +74,7 @@ void MeasureTool::activate()
 
 void MeasureTool::deactivate()
 {
+  line_->setVisible(false);
 }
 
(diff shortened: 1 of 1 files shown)


guihomework:
just one comment on this. When one measures something, in the current view direction, one cannot necessarily appreciate which surfaces were chosen to anchor the line, and one has to rotate to view it, and for this one has to select the orbit tool again. Deactivating the line forbids that check.

I liked to have this in the past, and would deactivate the line with right click before inishing the measurement "task" by re-selecting the tool. Not practical I admit but possible. So maybe one should have the orbiting possible while measuring ? I know clicking and orbiting is left mouse button, so not sure how to proceed. Maybe with a key-modifier allowing to rotate and pan ?

rhaschke:
Good point. I will revert this change for now.

PR #1497, opened 2020-04-27
Finalize tf2 migration

rhaschke:
This is the only blocker for the upcoming Noetic release (cf. #1498).
I started the migration process as follows:
- [x] Update `InteractiveMarkerDisplay` (8995e37c713d3d7efddaa5b7880c6594c831fe45)
- [x] Replace deprecated `getTFClient()` with `getTF2BufferPtr()`
  - [x] in `TfFrameProperty` (03a732fbf70c0e4c630492bc2f9d3e938f3d47ae)
  - [x] in `TfDisplay`
  - [x] in `InteractiveMarker`
  - [x] in `LaserScanDisplay`
- [x] Use (already migrated) `MessageFilterDisplay` for all remaining displays:
  - [x] `GridCellsDisplay`, which manually reimplements the logic of a `MessageFilterDisplay`
  - [x] `EffortDisplay`, which uses `MessageFilterJointState`, a custom re-imlementation of `tf::MessageFilter`
- [x] Adapt FrameManager and VisualizationManager to use `tf2_ros::Buffer`.
- [x] Remove all tf1-related, already deprecated functions.

Any help is highly appreciated, @jschleicher or @simonschmeisser?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,7 +119,8 @@ find_package(catkin REQUIRED
   sensor_msgs
   std_msgs
   std_srvs
-  tf
+  tf2_ros
+  tf2_geometry_msgs
   urdf
   visualization_msgs
 )
@@ -194,7 +195,7 @@ catkin_package(
     sensor_msgs
     std_msgs
     std_srvs
-    tf
+    tf2_ros
     urdf
     visualization_msgs
 )
--- a/package.xml
+++ b/package.xml
@@ -49,7 +49,8 @@
   <depend>sensor_msgs</depend>
   <depend>std_msgs</depend>
   <depend>std_srvs</depend>
-  <depend>tf</depend>
+  <depend>tf2_ros</depend>
+  <depend>tf2_geometry_msgs</depend>
   <depend>tinyxml2</depend>
   <depend>urdf</depend>
   <depend>visualization_msgs</depend>
--- a/src/python_bindings/sip/visualization_manager.sip
+++ b/src/python_bindings/sip/visualization_manager.sip
@@ -76,9 +76,9 @@ public:
   void setFixedFrame( const QString& frame );
   
   /**
-   * @brief Convenience function: returns getFrameManager()->getTFClient().
+   * @brief Convenience function: returns getFrameManager()->getTF2BufferPtr().
    */
-//  tf::TransformListener* getTFClient() const;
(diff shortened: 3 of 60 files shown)


simonschmeisser:
I tried a bit but found no easy translation for `tf->getLatestCommonTime` this is all deprecated and/or private in tf2

I managed to port LaserScanDisplay easily, so here is a PR https://github.com/rhaschke/rviz/pull/2

rhaschke:
I think I finished the migration. An open question is how to handle `tf_prefix`. I removed it, but probably we should keep it in `RobotModelDisplay` and `EffortDisplay`. Opinions?

simonschmeisser:
Great job Robert! I'll look over it tomorrow.

I know that the XACROs in ROS-I contain a prefix parameter for the purpose of multi-robot systems. But I would expect this to work without an additional prefix in RobotModelDisplay? Maybe @gavanderhoorn knows more?

guihomework:
I hope I got what is meant in the open question. We use this tf_prefix in ROS-I in the RobotModel display.

https://github.com/ubi-agni/agni_robots/blob/kinetic-devel/agni_description/config/tactile_display_both_fixed.rviz#L215

this is in order to have a specific display on a specific prefix that contains TF that are not moving (for fixing the posture of our robot hand for nice display), while still using the same urdf model as for the "moving" robot.

gavanderhoorn:
> I know that the XACROs in ROS-I contain a prefix parameter for the purpose of multi-robot systems. But I would expect this to work without an additional prefix in RobotModelDisplay? Maybe @gavanderhoorn knows more?

The `$prefix` you are referring to is only present at the `xacro` level. It's a standard approach to avoid link/joint/etc clashes when combining multiple xacro macros with others.

`tf_prefix` does make it easier to work with such xacros, but in the end these are not really related.

gavanderhoorn:
As to the removal of `tf_prefix` from all of these displays: I'm not very happy about it, but it is true that `tf_prefix` has been deprecated for quite some time now.

But it's a bit strange to deprecate something and then not provide an upgrade path or alternative. Many of the paragraphs / wiki pages about the deprecation of `tf_prefix` end with sentences such as (this one from [here](http://wiki.ros.org/tf2/Migration#Removal_of_support_for_tf_prefix)):

>  The development time for these tools unfortunately has not been found yet.

rhaschke:
I will keep `tf_prefix`, but not relying on tf1.

PR #1503, opened 2020-05-04
[melodic] Compatibility fix for OGRE 1.10.4 and up.

seanyen:
A recent [lint](https://github.com/ros-visualization/rviz/pull/1502) fix causes a build break when building RViz with OGRE v1.10.4 and up.

This pull request is attempting to make the RViz to be compatible with  OGRE 1.10.4 and up again. And here is some context:

Since OGRE v1.10.4, the signature of [`OgreMovableObject::getName`](https://github.com/OGRECave/ogre/blob/v1.10.4/OgreMain/include/OgreMovableObject.h#L204) is no longer a virtual function. To accommodate that, conditionally declare the matched signature of `getName()` for `rviz::MovableText`.

### DIFF
--- a/src/rviz/ogre_helpers/movable_text.h
+++ b/src/rviz/ogre_helpers/movable_text.h
@@ -203,7 +203,11 @@ protected:
     return mAABB;
   }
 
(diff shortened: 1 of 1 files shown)


seanyen:
@rhaschke This is ready for review and merge. Thanks!

rhaschke:
Thanks for this fix, @seanyen. However, I applied it to the noetic-devel branch only, as only there we have guarantee compatibility with newer versions of Ogre (and we can use the `OGRE_VERSION_CHECK` macro).

PR #1504, opened 2020-05-12
PropertyTree: set custom SelectionModel only with valid model

v4hn:
This seems to be a corner-case one might actually call a bug in qt5.
the `setSelectionModel(new PropertySelectionModel(nullptr))` call
complains that the SelectionModel and the PropertyTreeView use different
models, although both are set to nullptr in user code.

Qt actually sets the model of QAbstractItemView to
the dummy `QAbstractItemModelPrivate::staticEmptyModel`, but the
SelectionModel is actually initialized with `nullptr`.

### DIFF
--- a/src/rviz/properties/property_tree_widget.cpp
+++ b/src/rviz/properties/property_tree_widget.cpp
@@ -103,11 +103,12 @@ void PropertyTreeWidget::setModel(PropertyTreeModel* model)
   }
   model_ = model;
   QTreeView::setModel(model_);
-  QItemSelectionModel* m = selectionModel();
-  setSelectionModel(new PropertySelectionModel(model_));
-  m->deleteLater();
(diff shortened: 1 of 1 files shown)


rhaschke:
@v4hn: What are you doing?

v4hn:
This was `hub` failing to create a correct pull-request. Sorry for the noise.

v4hn:
I just updated the description and target branch.

rhaschke:
How severe is this issue? Is Qt just printing a warning or is it crashing?

v4hn:
> How severe is this issue? Is Qt just printing a warning or is it crashing?

It prints two lines to stderr per call with nullptr input.
For the motion planning display that's 6-8 lines of output.
Behavior is not affected.

PR #1505, opened 2020-05-13
Fix Travis CI for Noetic

rhaschke:
This PR comprises several minor fixes to get Travis CI passing again. Some of them are preliminary  workarounds for open issues:
* We don't yet have an official ROS Noetic docker image
* https://github.com/ros-visualization/python_qt_binding/pull/91: silence make warning
* https://github.com/rhaschke/python_qt_binding/commit/4eedad5e0b777ef0c4ac5a1eced583f8b7a6c937: silence (external source) warnings in sip build

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -23,7 +23,7 @@ AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortBlocksOnASingleLine: true
 AllowShortIfStatementsOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
-AllowShortFunctionsOnASingleLine: Empty
+AllowShortFunctionsOnASingleLine: None
 AllowShortCaseLabelsOnASingleLine: false
 
 AlwaysBreakTemplateDeclarations: true
@@ -59,6 +59,7 @@ BreakBeforeBraces: Custom
 # Control of individual brace wrapping cases
 BraceWrapping:
     AfterClass: true
+    AfterCaseLabel: true
     AfterControlStatement: true
     AfterEnum: true
     AfterFunction: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,19 +12,25 @@ notifications:
 
 env:
   global:
-    - ROS_DISTRO=melodic
-    - ROS_REPO=ros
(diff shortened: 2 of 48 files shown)


rhaschke:
Closing and reopening to retrigger CI.

rhaschke:
@wjwwood, @sloretz: Do you have an idea why the ROS build farm reports a failure here?
The [build job actually succeeds](http://build.ros.org/job/Npr__rviz__ubuntu_focal_amd64/4) - we just have a few cmake and compiler warnings - but all of them are triggered by external sources.

I already filed an issue for the assimp cmake warning: https://github.com/assimp/assimp/issues/3166
All the compiler issues are related to a new [gcc](https://gcc.gnu.org/gcc-9/changes.html)/[clang](https://releases.llvm.org/10.0.0/tools/clang/docs/ReleaseNotes.html#improvements-to-clang-s-diagnostics) warning (`-Wdeprecated-copy`), which is triggered in the sip build, as this activates all warnings (`-Wall`) from Qt's cflags.
To suppress them, we would need a way to pass extra cflags to `python_qt_bindings`.
For the Travis, I use the following quick fix to do so: https://github.com/rhaschke/python_qt_binding/commit/4eedad5e0b777ef0c4ac5a1eced583f8b7a6c937
From my point of view, the correct solution to this issue is to declare all external includes as `-isystem`.

wjwwood:
> @wjwwood, @sloretz: Do you have an idea why the ROS build farm reports a failure here?
The build job actually succeeds - we just have a few cmake and compiler warnings - but all of them are triggered by external sources.

I don't think the Jenkins plugin that reports this stuff to GitHub distinguishes between unstable and failure, but I'm not 100% sure.

> I already filed an issue for the assimp cmake warning: assimp/assimp#3166

FYI https://bugs.launchpad.net/ubuntu/+source/assimp/+bug/1869405 and https://github.com/ros2/rviz/issues/524 and https://github.com/ros2/rviz/pull/534

> From my point of view, the correct solution to this issue is to declare all external includes as -isystem.

I would caution against this, as we've run into subtle issues in the past with include orders being messed up when using `-isystem`. I'd prefer to fix warnings where they are or use the pragma diagnostics to suppress warnings where we include them if possible. Patching `python_qt_bindings` also seems appropriate.

PR #1506, opened 2020-05-17
Add display for TwistStamped

mvieth:
### Description

Display for TwistStamped. The source code from WrenchStamped was taken and adapted (wrench->twist, force->linear, torque->angular).


rhaschke:
Thanks for this contribution. In general, I support having a TwistStamped display as well, but I don't like the code duplication at all.
Could you refactor both displays to use the very same code base, please? The visual class is generic anyway. The display classes just need to distinguish between the message types, and based on this need to provide different (display) names of some properties. This could be easily achieved with a traits template class.

mvieth:
Ok, will do. Not in the next few weeks though, because I want to focus on my master thesis first.

mvieth:
@rhaschke PTAL. There is now the templated class ScrewDisplay, and the three inheriting classes AccelStampedDisplay, TwistStampedDisplay, and WrenchStampedDisplay. The problem was that the class with `Q_OBJECT` cannot be templated. But this way there is minimal code duplication.

Now the ABI is not compatible any more. Should I target another branch? If yes, which one?

rhaschke:
Merged into `noetic-devel` via dd1779e78dbb1468fdb11968632fafd5e7a272f2. Thanks for this contribution!

mvieth:
Nice to hear, thanks!

Issue #1509, opened 2020-05-27
[noetic] Broken 2D Nav Goal and 2D Pose Estimate button (orientation)


doisyg:
Last deb version:
```
[ INFO] [${node}]: rviz version 1.14.0
[ INFO] [${node}]: compiled against Qt version 5.12.8
[ INFO] [${node}]: compiled against OGRE version 1.9.0 (Ghadamon)
[ INFO] [/rviz_1590590368092672263]: Forcing OpenGl version 0.
[ INFO] [/rviz_1590590368092672263]: Stereo is NOT SUPPORTED
[ INFO] [/rviz_1590590368092672263]: OpenGl version: 4.6 (GLSL 4.6).
```
The 2D Nav Goal button seems broken under noetic. No orientation is published on the topic:
```
rostopic echo /move_base_simple/goal
header: 
  seq: 0
  stamp: 
    secs: 1590590378
    nsecs:  90359437
  frame_id: "map"
pose: 
  position: 
    x: 0.4193878173828125
    y: 0.5008087158203125
    z: 0.0
  orientation: 
    x: 0.0
    y: 0.0
    z: 0.0
    w: 0.0
---

```

Though the angle is displayed in the console:
`Setting goal: Frame:map, Position(0.419, 0.501, 0.000), Orientation(0.000, 0.000, 0.000, 0.000) = Angle: -2.977`

doisyg:
Same issue for the 2D Pose Estimate button:
```
rostopic echo /initialpose
header: 
  seq: 0
  stamp: 
    secs: 1590596724
    nsecs: 931619017
  frame_id: "map"
pose: 
  pose: 
    position: 
      x: -0.12504196166992188
      y: 0.07902240753173828
      z: 0.0
    orientation: 
      x: 0.0
      y: 0.0
      z: 0.0
      w: 0.0
  covariance: [0.25, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.25, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.06853892326654787]
---
```
with:
`[ INFO] [/rviz_1590596700755421369]: Setting pose: -0.125 0.079 -2.567 [frame=map]`

[...]

PR #1510, opened 2020-05-27
fix initial pose and goal buttons

doisyg:
Fix https://github.com/ros-visualization/rviz/issues/1509#
The wrong TF2 conversion was used

### DIFF
--- a/src/rviz/default_plugin/tools/goal_tool.cpp
+++ b/src/rviz/default_plugin/tools/goal_tool.cpp
@@ -71,7 +71,7 @@ void GoalTool::onPoseSet(double x, double y, double theta)
   tf2::Quaternion quat;
   quat.setRPY(0.0, 0.0, theta);
   geometry_msgs::PoseStamped goal;
-  goal.pose.orientation=tf2::toMsg(quat);
+  goal.pose.orientation = tf2::toMsg(quat);
   goal.pose.position.x = x;
   goal.pose.position.y = y;
   goal.header.frame_id = fixed_frame;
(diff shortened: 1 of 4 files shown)


rhaschke:
Can you explain a little bit more, what is wrong with `tf2::convert`? Shouldn't that work as well?

doisyg:
http://docs.ros.org/noetic/api/tf2/html/namespacetf2.html#af1255c37833c838543b512febce5e9b1
http://wiki.ros.org/tf2/Tutorials/Quaternions

```
void tf2::convert ( const A &  a,  B & b ) 
a | an object to convert from
b | the object to convert to
```

a and b were inverted, `quat_msg = tf2::toMsg(quat_tf)` does the same and has the advantage of being less ambiguous

doisyg:
But I will happily fix it using convert the other way around if preferred

rhaschke:
I see. Thanks for clarification.
There might be other places in the rviz code base using `tf2::convert` in the wrong order. It would be great, if you could check (and fix) them as well. I prefer the less ambiguous `toMsg()` / `fromMsg()` too. Thanks for your help!

doisyg:
`tf2::convert` is also used in https://github.com/ros-visualization/rviz/blob/noetic-devel/src/test/rviz_logo_marker.cpp and https://github.com/ros-visualization/rviz/blob/noetic-devel/src/test/marker_test.cpp, but properly I believe

rhaschke:
Just to improve readability / reduce ambiguity, I would appreciate it if you could replace those instances as well.

doisyg:
Done

rhaschke:
Could you please fix the clang-format issues indicated by Travis yourself here?

PR #1514, opened 2020-05-30
adding python version of send test pointcloud

flynneva:
### Description

added new test for sending pointcloud2 using python


src/test/send_pc2.py
```diff
@@ -0,0 +1,47 @@
...
+               PointField('y', 4, PointField.FLOAT32, 1),
+               PointField('z', 8, PointField.FLOAT32, 1) ]
+
+    header = Header()
+    header.frame_id = "map"
+    header.stamp = rospy.Time.now()
+
+    count = 0
+    points = []
+    for i in range(width):
```
rhaschke, line 28: What about replacing the (slow) loops with numpy array operations?

flynneva:
the cpp example wasnt compiling on my station for some reason. the cpp file was there, just after compilation I wasnt able to run it bc I dont think it compiled

flynneva:
sure ill switch over to numpy - just threw this together quickly and wasnt thinking about making it fast yet

rhaschke:
Ah. These test examples are only compiled with `catkin_make run_tests`.

flynneva:
i did that. all the other tests were there but the send_point_cloud_2 was not. it was strange. most likely user error (me) so I just made something that worked for me

PR #1528, opened 2020-07-22
fix selection of interactive markers

wolfv:
I think the conversion to float is too lossy and thus selection of interactive marker arrows didn't work for me (I ended up in the size == 3 case).


seanyen:
@wolfv Good catch! The same issue hit on Windows too.

PR #1529, opened 2020-07-22
fix getName override

wolfv:
Without this patch, I get the following compile time error:

```
/home/wolfv/Programs/catkin_ws/src/rviz/src/rviz/ogre_helpers/movable_text.h:206:23: error: ‘const String& rviz::MovableText::getName() const’ marked ‘override’, but does not override
  206 |   const Ogre::String& getName() const override
      |                       ^~~~~~~
```

I am not sure about the implications -- did this change with the OGRE version?
For reference, I am using OGRE 1.10.12 and QT 5.12.5

### DIFF
--- a/src/rviz/ogre_helpers/movable_text.h
+++ b/src/rviz/ogre_helpers/movable_text.h
@@ -203,7 +203,7 @@ protected:
     return mAABB;
   }
 
-  const Ogre::String& getName() const override
+  const Ogre::String& getName() const
   {
     return mName;
   }


rhaschke:
Indeed, the Ogre API has changed since Ogre 1.9, which we use by default. We need to postpone such changes until we migrate to a recent versionn of Ogre.

wolfv:
I don't understand why you close this. We can surely discuss putting this behind compile time #if statements?

rhaschke:
I was closing this because this doesn't apply to the melodic-devel branch. Feel free to reopen (or create a new PR as you did).

PR #1530, opened 2020-07-23
Fix getname override

wolfv:
This fixes building with OGRE 1.10.12

OGRE 1.12.X is still not working as far as I know.


rhaschke:
Hi @wolfv, please have a look into the noetic-devel branch, which already support Ogre 1.12.x.
There is also a macro `OGRE_VERSION_CHECK` to simplify what you are doing here.

rhaschke:
Just checked again: The noetic-devel branch is working fine with Ogre 1.12.x

PR #1535, opened 2020-08-17
[noetic] More portable assimp and OGRE discovery logic

seanyen:
* Since `assimp v5.0.0`, the `assimp` no longer defines `ASSIMP_LIBRARIES` so a backward-compatible logic is added to help locate `assimp`.

* Fallback to find OGRE by Config files if it cannot see it from `pkg-config` files (On some platforms, such as `Windows`, the upstream OGRE doesn't generate `pkg-config` files).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,12 +26,18 @@ find_package(Boost REQUIRED
 find_package(urdfdom_headers REQUIRED)
 
 find_package(ASSIMP REQUIRED)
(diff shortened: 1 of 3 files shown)


seanyen:
This is ready for review.

rhaschke:
Thanks Sean. I reworked your PR slightly (see #1536) to facilitate full transition to cmake-based find_package in the future.
Closing here.

PR #1539, opened 2020-08-24
RobotModelDisplay: Catch exception

wxmerkt:
This gets triggered when an invalid name for the robot_description
parameter is entered, e.g. "robot-description". Instead of dying due to
the exception, catch it and print it to ROS_ERROR_STREAM.

This resolves:
```
terminate called after throwing an instance of 'ros::InvalidNameException'
  what():  Character [-] at element [x] is not valid in Graph Resource Name [something-somethingelse].  Valid characters are a-z, A-Z, 0-9, / and _.
```

Analog to #1467 

As a side note, I used clang-format-3.9 but wasn't sure it's the correct one. Since clang-format handles the same config slightly differently between versions, could we please document in the README which version is meant to be used? (Or I might have missed where it was documented...)

### DIFF
--- a/src/rviz/default_plugin/robot_model_display.cpp
+++ b/src/rviz/default_plugin/robot_model_display.cpp
@@ -140,23 +140,30 @@ void RobotModelDisplay::load()
   context_->queueRender();
 
   std::string content;
-  if (!update_nh_.getParam(robot_description_property_->getStdString(), content))
+  try
   {
-    std::string loc;
-    if (update_nh_.searchParam(robot_description_property_->getStdString(), loc))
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
Thanks for pointing out this issue and providing a solution. I would prefer to report the invalid parameter name in a similar fashion as the other problems, i.e. via the display's `StatusProperty`. Also, we should `clear()` the display in case of this failure. Do you have cycles to apply this change?

wxmerkt:
I agree this would be the most user-friendly option - alas, I don't have any cycles to apply this change :(.

rhaschke:
Handled via a8e34d1a8c993350d63fb483d13ed07bbdc2778f.

Issue #1138, opened 2017-08-10
Change RViz tool display name and add multiple instances of same tool


jay3ss:
I can added multiple instances of the same tool to the tool bar, but I can't change the name of each instance of the tool. Currently this isn't possible because the names are hardcoded and set via the `setName` function. 

It would also be great if multiple instances of the same tool such as `2D Nav Goal` and `2D Pose Estimate` could be added via the `+` button instead of having to add them through a custom configuration file.

This is in response to [this question](https://answers.ros.org/question/268445/change-rviz-tool-display-name/) on ROS Answers.

wjwwood:
That makes sense, but I probably won't have time to implement it right now. If you want to implement it and open a pull request though, I'll try to get it merged in.

jay3ss:
Sure, I'll take a stab at it. I can't find guidelines on contributing, are there any?

wjwwood:
@jay3ss we don't have any written down guidelines, but generally speaking just try to maintain the code style in the file you modify and open a pr against `kinetic-devel`.

jay3ss:
@wjwwood will do, thank you.

rhaschke:
Closing due to inactivity.

PR #1570, opened 2020-12-16
[tool] load parameters before initializing

jcmonteiro:
### Description

With this PR the tool parameters will be available when `onInitialize` is called. One can use this feature to solve #1138 by adding
an `rviz::StringProperty`, e.g. named `Tool Name`, and then use its value to rename the tool in `onInitialize`.

### DIFF
--- a/src/rviz/tool.cpp
+++ b/src/rviz/tool.cpp
@@ -73,6 +73,7 @@ void Tool::setName(const QString& name)
 {
   name_ = name;
   property_container_->setName(name_);
+  Q_EMIT nameChanged(this);
 }
 
 void Tool::setDescription(const QString& description)
--- a/src/rviz/tool.cpp
+++ b/src/rviz/tool.cpp
@@ -71,9 +71,14 @@ void Tool::setCursor(const QCursor& cursor)
 
 void Tool::setName(const QString& name)
 {
+  // Early return if the name did not change
+  if (name == name_)
+    return;
+
+  // Change the name and emit a signal to let slots know
   name_ = name;
   property_container_->setName(name_);
-  Q_EMIT nameChanged(this);
+  Q_EMIT nameChanged(name_);
 }
 
 void Tool::setDescription(const QString& description)
--- a/src/rviz/tool.h
+++ b/src/rviz/tool.h
@@ -192,6 +192,7 @@ public:
 
 Q_SIGNALS:
   void close();
+  void nameChanged(Tool* tool);
 
 protected:
   /** Override onInitialize to do any setup needed after the
--- a/src/rviz/tool.h
+++ b/src/rviz/tool.h
@@ -192,7 +192,7 @@ public:
 
 Q_SIGNALS:
   void close();
-  void nameChanged(Tool* tool);
+  void nameChanged(const QString& name);
 
 protected:
   /** Override onInitialize to do any setup needed after the
--- a/src/rviz/visualization_frame.cpp
+++ b/src/rviz/visualization_frame.cpp
@@ -1185,6 +1185,19 @@ void VisualizationFrame::addTool(Tool* tool)
   tool_to_action_map_[tool] = action;
 
   remove_tool_menu_->addAction(tool->getName());
+
+  QObject::connect(tool, &Tool::nameChanged, this, &VisualizationFrame::onToolNameChanged);
+}
+
+void VisualizationFrame::onToolNameChanged(const QString& name)
+{
+  // Early return if the tool is not present
+  auto it = tool_to_action_map_.find(qobject_cast<Tool*>(sender()));
+  if (it == tool_to_action_map_.end())
+    return;
+
+  // Change the name of the action
+  it->second->setIconText(name);
 }
 
 void VisualizationFrame::onToolbarActionTriggered(QAction* action)
--- a/src/rviz/visualization_frame.h
+++ b/src/rviz/visualization_frame.h
@@ -227,6 +227,9 @@ protected Q_SLOTS:
    * the shortcut key, onToolbarActionTriggered() is called. */
   void addTool(Tool* tool);
 
+  /** @brief React to name changes of a tool, updating the name of the associated QAction */
+  void onToolNameChanged(const QString& name);
+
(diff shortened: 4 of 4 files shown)


rhaschke:
I'm not sure about the purpose of this PR yet. Isn't it common practice to first initialize an object (display, panel, tool, ...) and later load its parameters? I don't like the idea of coupling the loading to the creation of a tool. Regarding #1138: I guess you can always call `setName()` on your tool, can't you?

jcmonteiro:
> Isn't it common practice to first initialize an object (display, panel, tool, ...) and later load its parameters?
Hmm.. I'm not quite sure. After all, the initialization might depend on the parameters. What would be the downside of having the parameters available in `onInitialize`?

> Regarding #1138: I guess you can always call `setName()` on your tool, can't you?
Calling setName has no effect after `onInitialize`. At least it has no effect when called in `activate`. Besides, `activate` takes place only when the tool is clicked, so even if it worked there the name would only change to the correct one once the tool had been clicked.

rhaschke:
> After all, the initialization might depend on the parameters.

It shouldn't. It's always possible to load a different config later. However, the initialization should always work.

> Calling `setName` has no effect after `onInitialize`.

I don't see why this should be the case. Currently, `setName()` is called just shortly before `initialize()`:
https://github.com/ros-visualization/rviz/blob/13d061d973e96c6315c7768a7f316d2cc501299d/src%2Frviz%2Ftool_manager.cpp#L235-L237

jcmonteiro:
> It shouldn't. It's always possible to load a different config later. However, the initialization should always work.

I see... I did some more digging in the code and the issue is the following with `setName` and `setIcon`: they have no effect on the menu entries because they are created only once per tool in
https://github.com/ros-visualization/rviz/blob/2fe6d33f1ca7ceccd916cea94c199ac8a2f25ca5/src/rviz/visualization_frame.cpp#L1177-L1188

From what you explained about the initialization, I believe that the proper solution to #1138 is to change the `setName` and `setIcon` methods so that they force a reload of the action and update the member variables `action_to_tool_map_`, `tool_to_action_map_`, `toolbar_`, and `remove_tool_menu_` from `VisualizationFrame`. After doing so, calling `setName` or `setIcon` from an overloaded `load` method should do the trick. What do you think?

rhaschke:
That sounds reasonable. Would it be sufficient to change the name? I think, having the same icons for the same tool, would facilitate its identification. I suggest introducing a signal `nameChanged()`, such that the VisFrame is informed about name changes.

jcmonteiro:
Completely agree. I'll do it then.

jcmonteiro:
I've force-pushed the changes. Calling `setName` now updates the toolbar and this PR solves #1138.

src/rviz/tool.cpp
```diff
@@ -73,6 +73,7 @@ void Tool::setName(const QString& name)
 {
   name_ = name;
```
rhaschke, line 74: Please add a check to verify that name will actually change. If not, you can return early.

src/rviz/tool.h
```diff
@@ -192,6 +192,7 @@ class RVIZ_EXPORT Tool : public QObject
 
 Q_SIGNALS:
   void close();
+  void nameChanged(Tool* tool);
```
rhaschke, line 195: Please change the argument to `const QString&`, passing the new name. You can always access the sender of a signal by other means, namely [`QObject::sender()`](https://doc.qt.io/archives/qt-4.8/qobject.html#sender).

src/rviz/visualization_frame.cpp
```diff
@@ -1185,6 +1185,17 @@ void VisualizationFrame::addTool(Tool* tool)
...
+  QObject::connect(tool, &Tool::nameChanged, this, &VisualizationFrame::updateTool);
+}
+
+void VisualizationFrame::updateTool(Tool* tool)
+{
+  // Early return if the tool is not present
+  if (tool_to_action_map_.find(tool) == tool_to_action_map_.end())
+    return;
+  QAction* action = tool_to_action_map_[tool];
+  action->setIconText(tool->getName());
```
rhaschke, line 1198: Please avoid searching the map twice:
```suggestion
  auto it = tool_to_action_map_.find(sender());
  if (it == tool_to_action_map_.end())
    return;
  (*it)->setIconText(name);
```

src/rviz/visualization_frame.h
```diff
@@ -227,6 +227,12 @@ protected Q_SLOTS:
    * the shortcut key, onToolbarActionTriggered() is called. */
   void addTool(Tool* tool);
 
+  /** @brief Update the given tool to this frame's toolbar.
+   *
+   * This updates the QAction associated with the passed Tool instance.
+   * @sa addTool(Tool) */
+  void updateTool(Tool* tool);
```
rhaschke, line 234:
```suggestion
  void onToolNameChanged(const QString& name);
```

rhaschke (changes requested):
Thanks for this contribution. In general, this looks good. Some minor remarks though.

jcmonteiro:
Great suggestions! I've force-pushed again to meet them.

src/rviz/visualization_frame.h
```diff
@@ -227,6 +227,12 @@ protected Q_SLOTS:
    * the shortcut key, onToolbarActionTriggered() is called. */
   void addTool(Tool* tool);
 
+  /** @brief Update the given tool to this frame's toolbar.
+   *
+   * Updates the QAction associated with the tool that emitted the signal.
+   * @sa onToolNameChanged() */
```
rhaschke, line 233:
```suggestion
  /** React to name changes of a tool, updating the name of the associated QAction */
```

rhaschke (commented):
Two more nitpicks. Otherwise looks good. Thanks.

jcmonteiro:
> Two more nitpicks. Otherwise looks good. Thanks.

That's completely understandable :+1:. I've pushed the changes.

rhaschke (approved):

rhaschke:
Thanks a lot.

PR #1586, opened 2021-02-09
set radius first for points displayed by PointStamped display

ns130291:
This PR fixes the following problem:
Currently the PointStamped display does not show added points as long as the circular buffer is filled once fully. This problem is most of the times not noticed if the history size is quite short or the topic is updated very fast. But if you have a topic with a slow rate and a long history it can take some time to fill the buffer. 

The solution is to set the radius before calling the `setMessage` function of `PointStampedVisual` as this function uses the radius to set the size of the underlying `Shape`.


rhaschke:
Thanks for this fix.

Issue #379, opened 2012-09-19
Do not reset rviz display/namespace/TF visual data when checkboxes are toggled  (ros-pkg ticket #4840)


hershwg:
With a complex display, one will often want to turn of certain parts of the display (i.e. selected marker displays or namespaces within a marker display, or selected TF frames) to reduce clutter temporarily, but then turn them back on again. The current behaviour seems to be that when these things are turned off (using the checkbox in the Displays property grid), the data is lost; if they are turned on again the associated visuals to not reappear until a new associated message is received on the corresponding topic.

IMO, the checkbox most intuitively means "show/hide", and that if a "clear" behaviour is also desired, that this should be accomplished through other means (perhaps in a context menu for each display).

trac data:
- Owner: **hersh**
- Reporter: **bouffard**
- Reported at: **Wed Mar  2 14:13:12 2011**
- URL: http://code.ros.org/trac/ros-pkg/ticket/4840

goretkin:
I think a related issue is properly saving the which TF frames in the rviz config. At least in groovy-devel, it only saves whether the high-level is checked or not, and doesn't preserve which individual frames are displayed.

PeterMitrano:
I'm considering implementing this, can we agree on a behavior? Would it be acceptable to toggle visiblity instead of deleting the markers?

EricCousineau-TRI:
\cc @hidmic

PeterMitrano:
Discussion happening here: https://discourse.ros.org/t/changing-rviz-behavior-toggling-things-on-and-off/18894/10

We have any idea of what to implement, if anyone can help with implementation that would be appreciated!

PR #1587, opened 2021-02-10
first pass at toggling visibility

PeterMitrano:
### Description
Change the behavior of the Marker & MarkerArray checkboxes to only toggle visibility instead of deleting the markers. I feel like this is the behavior most people expect/prefer (see #379) I can document the change on the wiki if this is merged

### Checklist

- [ ] If you are changing GUI, please include screenshots showing how things looked *before* and *after*.
- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-visualization/rviz/pulls) to support the maintainers of RViz. Refer to the [RViz Wiki](https://github.com/ros-visualization/rviz/wiki/Maintainer-Guide#reviewing-pull-requests) for reviewing guidelines.

### DIFF
--- a/src/rviz/default_plugin/marker_display.cpp
+++ b/src/rviz/default_plugin/marker_display.cpp
@@ -319,11 +319,6 @@ void MarkerDisplay::processAdd(const visualization_msgs::Marker::ConstPtr& messa
     }
   }
 
-  if (!ns_it.value()->isEnabled())
-  {
-    return;
-  }
-
   bool create = true;
   MarkerBasePtr marker;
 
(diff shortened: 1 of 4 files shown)


rhaschke:
Thanks for this contribution. But, to be honest, I'm not sure we should touch this behaviour in Rviz1. Actually, the standard behaviour of most (if not all) display plugins is to reset on disabling/enabling. If we change the behaviour, it should be changed consistently for all default plugins AND we should provide a method to actually reset the display as suggested in the almost 10-years old issue #379. You are welcome though to trigger a new discussion on discourse about this. If there is enough momentum to change the behaviour (and adapt all displays) we should do it.

PeterMitrano:
Thanks for the quick response robert, I will post on the discord to see what others think. I have a few different ideas on how this could be implemented so we might be able to find a middle ground

rhaschke:
Please send me a link to the discourse discussion when initiated. I like the idea of having context menus associated to the top-level display entries offering delete, rename, duplicate, reset, etc.

PeterMitrano:
https://discourse.ros.org/t/changing-rviz-behavior-toggling-things-on-and-off/18894

PR #1596, opened 2021-03-16
RobotLink: check if loading mesh failed before creating entity

simonschmeisser:
fixes #1589 

As discovered by @rhaschke having a robot with an invalid mesh will still create an Ogre entity for it. When a Marker later contains the same invalid mesh it will skip some sanity checks due to the already existing entity and will throw an unhandled exception when actually trying to use the mesh.

PR #1597, opened 2021-03-18
RobotLink: check if loading mesh failed before creating entity  V2

### LINKED ISSUE #1589: Unhandled exception when marker contains mesh from unknown package
When a marker msgs with a mesh from an unknown package is loaded, it throws an Ogre exception which is unfortunately not handled:
```
[ WARN] [1613474597.318764360]: OGRE EXCEPTION(6:FileNotFoundException): Cannot locate resource package://motek_demo_2017/meshes/tilted-base.stl in resource group Autodetect or any other group. in ResourceGroupManager::openResource at /build/o
gre-1.9-B6QkmW/ogre-1.9-1.9.0+dfsg1/OgreMain/src/OgreResourceGroupManager.cpp (line 756)                                                                                                                                                            
terminate called after throwing an instance of 'Ogre::FileNotFoundException'

  what():  OGRE EXCEPTION(6:FileNotFoundException): Cannot locate resource package://motek_demo_2017/meshes/tilted-base.stl in resource group Autodetect or any other group. in ResourceGroupManager::openResource at /build/ogre-1.9-B6QkmW/ogre-1
.9-1.9.0+dfsg1/OgreMain/src/OgreResourceGroupManager.cpp (line 756)


backtrace:


1   __GI_raise                                                                                                                                                                                                                                                                     raise.c 51 0x7f6ae9cadfb7 
2   __GI_abort                                                                                                                                                                                                                                                                     abort.c 79 0x7f6ae9caf921 
3   ??                                                                                                                                                                                                                                                                                        0x7f6aea6a2957 
4   ??                                                                                                                                                                                                                                                                                        0x7f6aea6a8ae6 
5   std::terminate()                                                                                                                                                                                                                                                                          0x7f6aea6a8b21 
6   __cxa_rethrow                                                                                                                                                                                                                                                                             0x7f6aea6a8da9 
7   Ogre::Resource::load(bool)                                                                                                                                                                                                                                                                0x7f6ae717fb08 
8   Ogre::MeshManager::load(std::string const&, std::string const&, Ogre::HardwareBuffer::Usage, Ogre::HardwareBuffer::Usage, bool, bool)                                                                                                                                                     0x7f6ae70ec0a7 
9   Ogre::EntityFactory::createInstanceImpl(std::string const&, std::map<std::string, std::string, std::less<std::string>, Ogre::STLAllocator<std::pair<std::string const, std::string>, Ogre::CategorisedAllocPolicy<(Ogre::MemoryCategory)0>>> const *)                                     0x7f6ae7013ce6 
10  Ogre::MovableObjectFactory::createInstance(std::string const&, Ogre::SceneManager *, std::map<std::string, std::string, std::less<std::string>, Ogre::STLAllocator<std::pair<std::string const, std::string>, Ogre::CategorisedAllocPolicy<(Ogre::MemoryCategory)0>>> const *)            0x7f6ae70fbd44 
11  Ogre::SceneManager::createMovableObject(std::string const&, std::string const&, std::map<std::string, std::string, std::less<std::string>, Ogre::STLAllocator<std::pair<std::string const, std::string>, Ogre::CategorisedAllocPolicy<(Ogre::MemoryCategory)0>>> const *)                 0x7f6ae71c477f 
12  Ogre::SceneManager::createEntity(std::string const&, std::string const&, std::string const&)                                                                                                                                                                                              0x7f6ae71c0f1a 
13  rviz::MeshResourceMarker::onNewMessage(boost::shared_ptr<visualization_msgs::Marker_<std::allocator<void>> const> const&, boost::shared_ptr<visualization_msgs::Marker_<std::allocator<void>> const> const&)                                                                              0x7f6adba6df9a 
14  rviz::MarkerBase::setMessage(boost::shared_ptr<visualization_msgs::Marker_<std::allocator<void>> const> const&)                                                                                                                                                                           0x7f6adba6b571 
15  rviz::MarkerDisplay::processAdd(boost::shared_ptr<visualization_msgs::Marker_<std::allocator<void>> const> const&)                                                                                                                                                                        0x7f6adba559e6 
16  rviz::MarkerDisplay::processMessage(boost::shared_ptr<visualization_msgs::Marker_<std::allocator<void>> const> const&)                                                                                                                                                                    0x7f6adba56103 
17  rviz::MarkerDisplay::update(float, float)                                                                                                                                                                                                                                                 0x7f6adba5651f 
18  rviz::DisplayGroup::update(float, float)                                                                                                                                                                                                                                                  0x7f6aed181b88 
19  rviz::VisualizationManager::onUpdate()                                                                                                                                                                                                                                                    0x7f6aed24bb25 
20  QMetaObject::activate(QObject *, int, int, void * *)                                                                                                                                                                                                                                      0x7f6af0631305 
... <Mehr>                                                                                                                                                                                                                                                                                                   



```

### Your environment
* OS Version: e.g. Ubuntu 18.04
* ROS Distro: [Kinetic | Melodic]
    ```
    [ INFO] [1613475494.475253896]: rviz version 1.13.15
[ INFO] [1613475494.475289937]: compiled against Qt version 5.12.8
[ INFO] [1613475494.475299717]: compiled against OGRE version 1.9.0 (Ghadamon)
[ INFO] [1613475494.483962728]: Forcing OpenGl version 0.
[ INFO] [1613475495.046200133]: Stereo is NOT SUPPORTED
[ INFO] [1613475495.046340502]: OpenGl version: 3,0 (GLSL 1,3).

    ```
* If source build, which git commit? d4603a4f


I'll try to provide a PR soon

simonschmeisser:
alternative to #1596 which also fixes #1589 

I had to split the `RobotLink` constructor in order to be able to connect to the signal before the mesh loading happens

now unfortunately MoveIt's `RobotStateDisplay` does not extend upon `RobotModelDisplay` so that will need to be extended in a follow up

### DIFF
--- a/src/rviz/default_plugin/robot_model_display.cpp
+++ b/src/rviz/default_plugin/robot_model_display.cpp
@@ -99,6 +99,8 @@ void RobotModelDisplay::onInitialize()
 {
   robot_ = new Robot(scene_node_, context_, "Robot: " + getName().toStdString(), this);
 
+  connect(robot_, &Robot::linkMeshLoadingFailed, this, &RobotModelDisplay::onLinkMeshLoadingFailed);
+
   updateVisualVisible();
   updateCollisionVisible();
   updateAlpha();
@@ -114,6 +116,11 @@ void RobotModelDisplay::updateRobotDescription()
 {
   if (isEnabled())
     load();
+}
+
+void RobotModelDisplay::onLinkMeshLoadingFailed(const QString& linkName, const QString& details)
+{
+  setStatus(StatusLevel::Error, linkName, details);
 }
 
 void RobotModelDisplay::updateVisualVisible()
--- a/src/rviz/default_plugin/robot_model_display.h
+++ b/src/rviz/default_plugin/robot_model_display.h
@@ -81,6 +81,7 @@ private Q_SLOTS:
   void updateTfPrefix();
   void updateAlpha();
   void updateRobotDescription();
+  void onLinkMeshLoadingFailed(const QString& linkName, const QString& details);
 
 protected:
   /** @brief Loads a URDF from the ros-param named by our
--- a/src/rviz/robot/robot.cpp
+++ b/src/rviz/robot/robot.cpp
@@ -222,7 +222,10 @@ RobotLink* Robot::LinkFactory::createLink(Robot* robot,
                                           bool visual,
                                           bool collision)
 {
-  return new RobotLink(robot, link, parent_joint_name, visual, collision);
+  RobotLink* l = new RobotLink(robot, link, parent_joint_name);
+  connect(l, &RobotLink::meshLoadingFailed, robot, &Robot::linkMeshLoadingFailed);
+  l->init(link, visual, collision);
+  return l;
 }
 
 RobotJoint* Robot::LinkFactory::createJoint(Robot* robot, const urdf::JointConstSharedPtr& joint)
--- a/src/rviz/robot/robot.h
+++ b/src/rviz/robot/robot.h
@@ -242,6 +242,8 @@ public:
 
   // set joint checkboxes and All Links Enabled checkbox based on current link enables.
   void calculateJointCheckboxes();
+Q_SIGNALS:
+  void linkMeshLoadingFailed(const QString& linkName, const QString& details);
 
 private Q_SLOTS:
   void changedLinkTreeStyle();
--- a/src/rviz/robot/robot_link.cpp
+++ b/src/rviz/robot/robot_link.cpp
@@ -154,9 +154,7 @@ void RobotLinkSelectionHandler::postRenderPass(uint32_t /*pass*/)
 
 RobotLink::RobotLink(Robot* robot,
                      const urdf::LinkConstSharedPtr& link,
-                     const std::string& parent_joint_name,
-                     bool visual,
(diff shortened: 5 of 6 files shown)


rhaschke:
I thought a little more about other alternatives to implement this. 
My first idea was to use exceptions, but they would leave the loop in RobotLink (and Robot) creating all geometries (and links).
Also, the current approach only shows the last error message.
What about storing all failures in a string in RobotLink and provide a method to collect (i.e. concatenate) all these failures with a new function in rviz::Robot? This way, a display could show all of them once after instantiation of the robot.
Sorry for coming up with this idea so late.

simonschmeisser:
This actually shows the last error **per link**, so yes, if your visual and collision meshes are both not found your link will have only one error but if you have multiple links with broken meshes then they will all be displayed as in error. And you can only fix one error at a time, right? (Our second use case besides not crashing is a urdf editor with preview and it would be great to have more feedback there. Currently it just does not display the mesh if you have a typo.)

Your approach with a list of problem strings that might be empty also sounds like a good idea however, I'll think about it

simonschmeisser:
Actually the errors got overridden by successful tf updates so they now have separate sections

grafik

simonschmeisser:
Wouldn't adding such a string break ABI compatibility? What do you think about the current state of this PR?

PR #1609, opened 2021-04-06
ColorEditor: maintain edited text + cursor pos

### LINKED ISSUE #1607: ColorEditor cursor will jump to end when input text
I try to input color value in ColorEditor, after any number input, the cursor will jump to end, see the gif.
cursor_jump

I found that cause by the model_->emitDataChanged(this);
https://github.com/ros-visualization/rviz/blob/ba536b2d5baae4be167f9741a3f5bcb05d129af6/src/rviz/properties/color_property.cpp#L60-L61

after a valid value is input, the dataChanged signal will call a setText() of QLineEdit. That make the cursor jump to end.

So, could the model_->emitDataChanged(this); be delete in ColorProperty? 
Now, I found this issue only in colorEditor.

### Environment
* OS Version: Ubuntu 18.04
* ROS Distro: Melodic
```
[ INFO] [1617528442.570476806]: rviz version 1.13.16
[ INFO] [1617528442.570531904]: compiled against Qt version 5.9.5
[ INFO] [1617528442.570563003]: compiled against OGRE version 1.9.0 (Ghadamon)
[ INFO] [1617528442.574762879]: Forcing OpenGl version 0.
[ INFO] [1617528442.680312957]: Stereo is NOT SUPPORTED
[ INFO] [1617528442.680650647]: OpenGL device: llvmpipe (LLVM 10.0.0, 256 bits)
[ INFO] [1617528442.680780743]: OpenGl version: 3.1 (GLSL 1.4).
```

rhaschke:
Fixes #1607.

### DIFF
--- a/src/rviz/properties/color_editor.cpp
+++ b/src/rviz/properties/color_editor.cpp
@@ -78,13 +78,18 @@ void ColorEditor::resizeEvent(QResizeEvent* event)
 
 void ColorEditor::parseText()
 {
-  QColor new_color = parseColor(text());
+  const QString t = text();
+  QColor new_color = parseColor(t);
   if (new_color.isValid())
   {
     color_ = new_color;
     if (property_)
     {
+      auto pos = cursorPosition();
       property_->setColor(new_color);
+      // setColor() normalizes the text display and thus looses cursor pos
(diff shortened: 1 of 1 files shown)


tsukione (approved):
Ok with this fix

PR #1631, opened 2021-06-07
Fix transform issue in odometry display plugin

UniBwTAS:
### Description

This pull request fixes a transform issue with odometry plugin for rviz. The previous implementation only shows meaningful Arrows/Axes when the "fixed frame" of rviz is equal to the parent frame of the Odometry message. However, when the fixed frame is different, then the pose of the visuals have to be constantly re-computed because they moved w.r.t. the fixed frame. Currently the transform is calculated just once when the message arrives. We re-compute the transform of the scene node (which corresponds to the parent frame of the Odometry message and contains all the Arrows/Axes) whenever the update (=redraw) callback is triggered.

Additionally we introduced a ref_pose_ which is useful for poses with very large positional values. E.g. the Pose in the UTM frame. Without this ref_pose_ this causes glitches because Ogre only works with single point precision (float). So Ogre has problems to calculate the transform fixed_frame (often base_link or frame close to base_link) -> utm -> base_link. By introducing this ref_pose_ Ogre has to handle much smaller positional values and the offset to the ref_pose_ can be calculated in double point precision.

Image before:

Screenshot from 2021-06-07 18-29-19

Image after:

Screenshot from 2021-06-07 18-56-21

Image for better understanding. In the attached rosbag we compare ground truth odometry against dead-reckoning. It contains the transform utm -> odom -> base_link, one odom message for dead reckoning and odom message for ground truth (gps + imu).

Screenshot from 2021-06-07 17-52-23

Files to reproduce (rosbag & rviz config):
https://www.dropbox.com/sh/mo6zo6akakiap6e/AAAShCSywQxV4sWC5JtjF5TEa?dl=0


src/rviz/default_plugin/odometry_display.h
```diff
@@ -94,14 +94,18 @@ private Q_SLOTS:
...
 
   typedef std::deque<rviz::Arrow*> D_Arrow;
   typedef std::deque<rviz::Axes*> D_Axes;
 
   D_Arrow arrows_;
   D_Axes axes_;
 
-  nav_msgs::Odometry::ConstPtr last_used_message_;
+  nav_msgs::Odometry::ConstPtr last_drawn_message_;
+  nav_msgs::Odometry::ConstPtr last_message_;
```
rhaschke, line 106: What's the difference between those two?

src/rviz/default_plugin/odometry_display.cpp
```diff
@@ -372,6 +382,26 @@ void OdometryDisplay::update(float /*wall_dt*/, float /*ros_dt*/)
...
+}
+
+bool OdometryDisplay::applyTransforms()
+{
+  if(!ref_pose_)
+    return false;
+
+  Ogre::Vector3 position;
+  Ogre::Quaternion orientation;
+  if (!context_->getFrameManager()->transform(last_message_->header, *ref_pose_, position, orientation))
```
rhaschke, line 396: Seems like `last_message_` is only used to store its header?
What if the involved frames in the odometry messages change over time?

rhaschke (changes requested):
I understand that you want to see the past path of your robot/car when using a fixed frame that is attached to the robot. 
In principle, I support this PR. However, there are several shortcomings that need to be fixed:
The PR combines several changes, which should be split into individual commits, each providing a concise commit message:
- Smoothly updating the scene node's position + orientation
- Introducing `ref_pose` to be more robust w.r.t. huge positions
- Cosmetic changes, e.g. renaming `length()` :arrow_right: `norm()`

Renaming and introducing new members breaks ABI, which should be strictly avoided. Can you think of alternative implementations? If not, try to restrict incompatible API/ABI changes to a minimum!

- Removing `last_used_message_` and introducing two new last*message members. Is that really needed? It breaks ABI.
  As far as I can see, you need `last_message_` for smoothly updating the scene node's position + orientation.
  Why not moving this update into `processMessage()`? Then you don't need to remember `last_message_`?
- There is a significant drift of the shown trajectory when the fixed transform is set to `odom`. Could be numerical issues, but could be a conceptual problem as well. Using `odom` as fixed frame, I expect to see exactly the old behavior. 
- Instead of using the last message's frame as the reference frame, shouldn't you use the same frame_id as reference, e.g. the one from the first seen message? Why happens if odometry messages are expressed w.r.t. a different parent frame in between?

src/rviz/default_plugin/odometry_display.cpp
```diff
@@ -246,8 +246,7 @@ void OdometryDisplay::updateShapeVisibility()
 
 bool validateFloats(const nav_msgs::Odometry& msg)
 {
-  bool valid = true;
-  valid = valid && rviz::validateFloats(msg.pose.pose);
+  bool valid = rviz::validateFloats(msg.pose.pose);
```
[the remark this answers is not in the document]
UniBwTAS: Ok!

src/rviz/default_plugin/odometry_display.cpp
```diff
@@ -276,47 +275,55 @@ void OdometryDisplay::processMessage(const nav_msgs::Odometry::ConstPtr& message
...
+    Eigen::Vector3d current_position(message->pose.pose.position.x, message->pose.pose.position.y,
+                                     message->pose.pose.position.z);
+    Eigen::Quaternionf last_orientation(last_drawn_message_->pose.pose.orientation.w,
+                                        last_drawn_message_->pose.pose.orientation.x,
+                                        last_drawn_message_->pose.pose.orientation.y,
+                                        last_drawn_message_->pose.pose.orientation.z);
     Eigen::Quaternionf current_orientation(message->pose.pose.orientation.w,
                                            message->pose.pose.orientation.x,
                                            message->pose.pose.orientation.y,
                                            message->pose.pose.orientation.z);
```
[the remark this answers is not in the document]
UniBwTAS: Ok!

src/rviz/default_plugin/odometry_display.h
```diff
@@ -94,14 +94,18 @@ private Q_SLOTS:
...
+  geometry_msgs::Pose::Ptr ref_pose_;
+  Ogre::SceneNode* frame_node_visuals_{};
```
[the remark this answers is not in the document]
UniBwTAS: Yes it is possible. I removed the new one.

src/rviz/default_plugin/odometry_display.h
UniBwTAS: last_drawn_message_ is the message which is actually drawn into the scene and required to calculate the positional and angular difference with the current message in order to decide whether the new message should be drawn into the scene or not.
last_message_ was stored to store the timestamp of the most recent message which is required to get the current transform between the fixed frame and the parent frame of the Odometry message.

However, I was able to undo these changes by simply transforming with the most recent transform available (using ros::Time() in transform(), see new commits). The most recent transform should be at least as new as the most recent message due to MessageFilterDisplay. I think this is ok in most scenarios.

src/rviz/default_plugin/odometry_display.cpp
UniBwTAS: Yes its only used to store its header or mainly its timestamp. See above.

For me it is a very uncommon scenario to publish Odometry messages with different parent/child frame_id's to the same topic. Currently this wouldn't be handled properly.

UniBwTAS: To handle this properly a scene node for each parent frame would be required, which would make everything more complex and definitively break API/ABI.

src/rviz/default_plugin/odometry_display.h
UniBwTAS: I could add a warning when a Odometry message with a different parent frame_id arrives, which says that it should be published on another topic.

src/rviz/default_plugin/odometry_display.cpp
UniBwTAS: I could add a warning when a Odometry message with a different parent frame_id arrives, which says that it should be published on another topic.

UniBwTAS:
I split the original commit into 3 distinct ones.

> Why not moving this update into `processMessage()`? Then you don't need to remember `last_message_`?

The transform has to be calculated continuously when the fixed_frame (e.g. base_link) is moving w.r.t. to the parent frame of the Odometry message (e.g. odom). When the Odometry message comes in not very frequently then the movement of the Trajectory does not look very smooth. So I decided to put it into the `update()` callback. Nevertheless I was able to get rid of `last_message_` by simply using the most recent transform available. I think its ok in most scenarios.

> There is a significant drift of the shown trajectory when the fixed transform is set to odom. Could be numerical issues, but could be a conceptual problem as well. Using odom as fixed frame, I expect to see exactly the old behavior.

I think the drift you mean only comes from the inaccuracies of 2D (no pitch or roll) dead reckoning (red trajectory) compared to that of inertial navigation system (blue trajectory). For me everything looks fine w.r.t. to visualization.

UniBwTAS:
By the way, using the attached rosbag one can see the importance of `ref_pose_` (introduced in the second commit). The blue trajectory looks very ugly due to large positional values in the UTM frame which causes numerical issues:

Screenshot from 2021-06-22 11-28-54

rhaschke (changes requested):
Thanks for cleaning up this PR. 

> Without my commits it is meaningless to compare red and blue trajectory since the latter has utm as its parent frame and hence for it the fixed frame should be set to utm. (However, UTM origin is quite far away from trajecory)

I don't get this statement. Why shouldn't it be possible use odom as the fixed frame for the blue trajectory? Without your changes (left), the blue trajectory is perfectly displayed on top of the red one, implying that the frame conversion works perfectly fine.
With your changes (right), the blue trajectory drifts over time (i.e. history changes):

https://user-images.githubusercontent.com/5376030/122936479-7a352300-d371-11eb-8de4-fcca385ea99d.mp4

It looks like the order of orientation and translation is somehow screwed.

src/rviz/default_plugin/odometry_display.cpp
```diff
@@ -372,6 +368,23 @@ void OdometryDisplay::update(float /*wall_dt*/, float /*ros_dt*/)
...
+  if(!last_used_message_)
+    return;
+
+  Ogre::Vector3 position;
+  Ogre::Quaternion orientation;
+  if (!context_->getFrameManager()->getTransform(last_used_message_->header.frame_id, ros::Time(),
+                                                 position, orientation))
+  {
+    ROS_ERROR("Error transforming odometry '%s' from frame '%s' to frame '%s'", qPrintable(getName()),
+              last_used_message_->header.frame_id.c_str(), qPrintable(fixed_frame_));
```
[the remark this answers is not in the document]
UniBwTAS: Even though the left/old one looks better, the right one is correct. Our dead reckoning (red) is really that bad ;) It is only based on wheel turn rates and a yaw gyroscope from the ESP sensor. Therefore it only works in the 2D plane as it cannot detect roll and pitch of the vehicle. The inertial navigation system, on the other hand, is able to measure roll and pitch. Thats why the blue trajectory pitches and rolls w.r.t. to the fixed frame (odom, which is based on our dead reckoning).

That the left one looks meaningful is a coincidence because the trajectories are "similar". New Odometry messages, or more precisely the arrows, are inserted at the right position of the scene but old ones are not updated when the fixed frame moves w.r.t. these arrows.

rhaschke:
> That the left one looks meaningful is a coincidence because the trajectories are "similar". New Odometry messages, or more precisely the arrows, are inserted at the right position of the scene but old ones are not updated when the fixed frame moves w.r.t. these arrows.

My main concern is that the history of the blue trajectory changes, which shouldn't be the case in my opinion.

rhaschke:
tf2_eigen package is missing (see [buildfarm job](https://build.ros.org/job/Npr_db__rviz__debian_buster_amd64/74/))

UniBwTAS:
> My main concern is that the history of the blue trajectory changes, which shouldn't be the case in my opinion.

No the history of blue arrows is not changing. Just its parent frame `utm` moved with respect to fixed_frame `odom`.

src/rviz/default_plugin/odometry_display.cpp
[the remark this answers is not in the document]
UniBwTAS: > If you are at it: Could you replace this console error output with a display status error (`Display::setStatus`)?
> Remember to reset the status to OK if there is no error.

I added this in the latest commit. I borrowed the code from `grid_display.cpp`.

UniBwTAS:
> tf2_eigen package is missing (see [buildfarm job](https://build.ros.org/job/Npr_db__rviz__debian_buster_amd64/74/))

I added the `tf2_eigen` package into the CMakeList.txt

CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ find_package(catkin REQUIRED
   std_srvs
   tf2_ros
   tf2_geometry_msgs
+  tf2_eigen
```
rhaschke, line 141: This also needs to be added to package.xml as a `build_depend`.

src/rviz/default_plugin/odometry_display.cpp
```diff
@@ -383,18 +383,30 @@ void OdometryDisplay::update(float /*wall_dt*/, float /*ros_dt*/)
   if (!last_used_message_ || !ref_pose_)
     return;
 
+  std::string frame = last_used_message_->header.frame_id;
```
rhaschke, line 386: Please use `const std::string&`. No need to copy the string.

rhaschke (changes requested):
> > My main concern is that the history of the blue trajectory changes, which shouldn't be the case in my opinion.
> 
> No the history of blue arrows is not changing. Just its parent frame `utm` moved with respect to fixed_frame `odom`.

Sure, but in effect, the past trajectory moves in space, which is wrong in my opinion.

src/rviz/default_plugin/odometry_display.cpp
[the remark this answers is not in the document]
UniBwTAS: I finally removed the setStatus commit again, as this part is already handled by the MessageFilterDisplay Super-Class. Otherwise there would be dup
Screenshot from 2021-06-22 20-56-12
licate Transform items:

src/rviz/default_plugin/odometry_display.cpp
UniBwTAS: I dropped this commit anyway as this part is already handled by MessageFilterDisplay

CMakeLists.txt
UniBwTAS: Done

UniBwTAS:
> > > My main concern is that the history of the blue trajectory changes, which shouldn't be the case in my opinion.
> > 
> > 
> > No the history of blue arrows is not changing. Just its parent frame `utm` moved with respect to fixed_frame `odom`.
> 
> Sure, but in effect, the past trajectory moves in space, which is wrong in my opinion.

In my opinion this is absolutely mandatory since otherwise the visualization makes no sense any more as soon you switch to another fixed_frame than the parent frame of the Odometry message.

src/rviz/default_plugin/odometry_display.cpp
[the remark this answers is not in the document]
rhaschke: Sure. Sorry for the noise.

UniBwTAS:
I added a video. Maybe you will better see why the history of arrows is important to us and why it should move w.r.t. to the fixed_frame. Here the fixed_frame is set to `utm`. Now the aerial image and the blue trajectory are fixed. The red trajectory pitches and rolls together with the car (as it is just 2D). Now we can compare how much the origin of the red trajectory drifts over time compared to ground truth blue trajectory:

https://www.dropbox.com/s/p13n2i6f46zsqup/odometry_display_with_utm_as_fixed_frame.mp4?dl=0

UniBwTAS:
The arrows are still fixed as before when fixed_frame is parent frame of odometry messages. In this case the plugin behaves the same as before. However, when the fixed_frame is different from parent frame it now shows meaningful trajectories whereas in previous implementation the visual output was unusable.

UniBwTAS:
We created a synthetic example which shows the problem with the current implementation:
Current Implementation:
https://user-images.githubusercontent.com/11663218/123261362-f9079880-d4f6-11eb-99cf-b7a26722980d.mp4
New Implementation:
https://user-images.githubusercontent.com/11663218/123261423-0c1a6880-d4f7-11eb-82dc-544d70ff268b.mp4

The code generating the synthetic example can be found [here](https://www.dropbox.com/s/p13n2i6f46zsqup/odometry_display_with_utm_as_fixed_frame.mp4?dl=0).

rhaschke:
Thanks for providing this example. If I now understand correctly, the observed drift comes from slightly different odometries published in the two topics of the rosbag file. Depending on which frame is used as the fixed frame, one or the other curve stays fixed, right?

AndreasR30:
> Thanks for providing this example. If I now understand correctly, the observed drift comes from slightly different odometries published in the two topics of the rosbag file. Depending on which frame is used as the fixed frame, one or the other curve stays fixed, right?

Yes

rhaschke (approved):

rhaschke:
I think this is fine to merge as soon as https://github.com/UniBwTAS/rviz/pull/1 is merged into this PR.

UniBwTAS:
> I think this is fine to merge as soon as [UniBwTAS#1](https://github.com/UniBwTAS/rviz/pull/1) is merged into this PR.

Done

UniBwTAS:
@rhaschke Is there still a reason not to merge this pull request?

AndreasR30:
@rhaschke ping

rhaschke:
Sorry for the delay merging this.

rhaschke:
I reverted this PR for the same reason as #1655, namely updating the node's transform doesn't consider the message's timestamp anymore.

PR #1632, opened 2021-06-08
Fix resizeEvent for OGRE==1.10

Tobias-Fischer:
Several users in the RoboStack project reported that rviz scaling didn't work for high-dpi screens: https://github.com/RoboStack/ros-noetic/issues/102

We use Ogre 1.10 in RoboStack. After applying https://github.com/RoboStack/ros-noetic/commit/057e616d4d03c4dc23567dd1b82d2add4b7de2d0 it is now working fine.

Therefore, I think the selector for the Ogre version was wrong and this PR fixes it.


rhaschke (approved):

rhaschke:
Thanks for this fix!

PR #1634, opened 2021-06-11
Add configurable features for camera view panel

tongtybj:
### Description

Add option to modify far clip and vertical FOV of camera view, which are useful when viewing in FPS mode (e.g. frame aligned mode).

- camera view panel before changing:
Screenshot from 2021-06-11 11-29-08

- after change:

Screenshot from 2021-06-11 11-18-54

### DIFF
--- a/src/rviz/view_controller.cpp
+++ b/src/rviz/view_controller.cpp
@@ -62,6 +62,20 @@ ViewController::ViewController()
   near_clip_property_->setMin(0.001);
   near_clip_property_->setMax(10000);
 
+  far_clip_property_ =
+      new FloatProperty("Far Clip Distance", 100.0f,
+                        "Anything farer to the camera than this threshold will not get rendered.", this,
+                        SLOT(updateFarClipDistance()));
+  far_clip_property_->setMin(0.001);
+  far_clip_property_->setMax(100000);
+
+  fov_property_ =
+      new FloatProperty("Vertical FOV", 45.0f, "Vertical Filed Of View (FOV) of camera (degrees).", this,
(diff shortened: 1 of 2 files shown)


rhaschke (commented):
This looks good to me in principle. But it breaks ABI. But we cannot postpone it to the next ROS1 release either, because there will be none. What about targeting this for ROS2 only?
Question: Did you validate that the default values you introduced are the same as used by the camera originally? Could you provide a link, please?

tongtybj:
- Fixed formatting issues
- We found the default values in https://github.com/OGRECave/ogre/blob/master/OgreMain/src/OgreCamera.cpp#L60-L63: 
  ```
  mFOVy = Radian(Math::PI/4.0f);
  mNearDist = 100.0f;
  mFarDist = 100000.0f;
  ```
  However, we are not sure whether the upper limit for mFOVy (90 degree) is proper.
- Question: could you please explain what does "it breaks ABI." mean in a little more detail?

rhaschke:
> We found the default values in https://github.com/OGRECave/ogre/blob/master/OgreMain/src/OgreCamera.cpp#L60-L63

These are different values than you used: far clipping distance 100 vs. 100000

What about reading the initial value from the created OGRE camera? Then you are sure to be always in sync with OGRE defaults.

I'm still hesitating to merge such a ABI change to a base class.

rhaschke:
To avoid ABI breakage in the base class: What about adding those additional properties only in the `FPSViewController`?

rhaschke:
Closing due to inactivity.

Issue #629, opened 2013-04-16
RViz crash when saving files


fersarr:
Hi, I have been using rqt-rviz to see the data from my kinect (via openni) but everytime I try to save a file, any kind (image, configuration file,etc.) the "save as..." window opens but it doesnt load nor say anything, just gray. And my ubuntu screen gets darker (like when programs crash). When I type CTRL-C in the terminal to kill rqt-rviz I get this error:

[ INFO] [1365896813.550288513]: OpenGl version: 3.0 (GLSL 1.3).
python: Fatal IO error 11 (Resource temporarily unavailable) on X server :0.
python: /usr/include/boost/thread/pthread/recursive_mutex.hpp:92: boost::recursive_mutex::~recursive_mutex(): Assertion `!pthread_mutex_destroy(&m)' failed.
Aborted (core dumped)

But that message does not appear if I was using rviz alone (no rqt). But it does crash in both ways. Heres a screenshot

http://s3.postimg.org/mu8o1k2oi/Screenshot3.jpg

There is no problem iwhen I try to save if I I had not launched openni first

PR #1635, opened 2021-06-17
Pause asynchronous ROS updates with synchronous ones

### LINKED ISSUE #868: Memory leak when displaying DepthCloud and showing the 'Add' menu
We experienced a severe memory leak in rviz today.
Here is the problem, we are displaying a `DepthCloud` using a depth and color image, from a rosbag at normal speed. As soon as I press the 'Add' button to add another visualisation the GUI stops updating and the memory usage increase rapidly, i.e. ~500MB per second. This is the reason for our laptop with very little memory, to freeze immediately. If the `DepthCloud` is turned off, no such behaviour can be observed.
This only happens with the `DepthCloud` as far as we can say. Not sure about point clouds.

We are running 14.04 64bit and ROS Indigo. rviz: version 1.11.7 (indigo) compiled against OGRE version 1.8.1 (Byatis).

### LINKED ISSUE #1606: Whenever open a dialog, the data stop update in 3D area and panels
Maybe this is not an issue, just a qustion.
I found every time a open a Dialog, such as NewDisplay, NewPanels, the pointcloud and image stop updating.

I check the code, found every time before open a dialog, will call 
https://github.com/ros-visualization/rviz/blob/7383000e722d581dc8657f2bf6016fca88c4aee0/src/rviz/visualization_frame.cpp#L642
When the dialog is closed, a startUpdate() wil called.

So, I just wondering why the stopUpdate() should be called. I comment the  manager_->stopUpdate(); line, It still working perfectly, and the pointcloud and image is continuous changing, I'm happy with that :)

So, could you help me to get out from this question? thanks.

### Environment
* OS Version: Ubuntu 18.04
* ROS Distro: Melodic
```
[ INFO] [1617528442.570476806]: rviz version 1.13.16
[ INFO] [1617528442.570531904]: compiled against Qt version 5.9.5
[ INFO] [1617528442.570563003]: compiled against OGRE version 1.9.0 (Ghadamon)
[ INFO] [1617528442.574762879]: Forcing OpenGl version 0.
[ INFO] [1617528442.680312957]: Stereo is NOT SUPPORTED
[ INFO] [1617528442.680650647]: OpenGL device: llvmpipe (LLVM 10.0.0, 256 bits)
[ INFO] [1617528442.680780743]: OpenGl version: 3.1 (GLSL 1.4).
 ```

### LINKED ISSUE #1621: DepthCloud memory leak
I am using the `DepthCloud` display type to back-project colour and depth to a 3D point cloud in RViz. When I click the `Add` button I can add a new display type. While this new pop-up window is shown, the point cloud in the 3D view is not updated and the memory consumption by RViz rises until its process gets killed by `oom-kill`.

Edit:
I just realised that this issue has already been reported as https://github.com/ros-visualization/rviz/issues/868.

Reproduce:
1. play a bag file with depth and colour images
2. open RViz
3. add and configure the `DepthCloud` display type
4. try to add a new display type by clicking on the `Add` button

My configuration:
rviz_screenshot_2021_04_15-15_38_36

Memeory consumption of the process:
Screenshot from 2021-04-15 15-27-46

### Your environment
* OS Version: Ubuntu 20.04
* ROS Distro: Noetic
```
rviz version 1.14.5
compiled against Qt version 5.12.8
Forcing OpenGl version 0.
Stereo is NOT SUPPORTED
OpenGL device: Mesa Intel(R) UHD Graphics 630 (CFL GT2)
OpenGl version: 4.6 (GLSL 4.6) limited to GLSL 1.4 on Mesa system.
```

rhaschke:
VisualizationManager uses `[start|stop]Update()` to pause ROS updates during dialog box interaction (#631).
However, asynchronous updates via `threaded_queue_` were still performed, leading to a memory leak (#1621),
because messages were partially processed by the thread (i.e. kept in memory), but not finally processed
(and cleaned up) by the synchronous `Display::update()` call.
Pausing threaded processing too resolves that issue.

Fixes #1621. Fixes #868.
@simonschmeisser: Do you see any issue pausing asynchronous message processing too while displaying dialogs?

### DIFF
--- a/src/rviz/default_plugin/depth_cloud_display.cpp
+++ b/src/rviz/default_plugin/depth_cloud_display.cpp
@@ -159,10 +159,6 @@ void DepthCloudDisplay::onInitialize()
   updateUseAutoSize();
   updateUseOcclusionCompensation();
 
-  // PointCloudCommon sets up a callback queue with a thread for each
-  // instance.  Use that for processing incoming messages.
-  threaded_nh_.setCallbackQueue(pointcloud_common_->getCallbackQueue());
-
   // Scan for available transport plugins
   scanForTransportSubscriberPlugins();
 
--- a/src/rviz/default_plugin/fluid_pressure_display.cpp
+++ b/src/rviz/default_plugin/fluid_pressure_display.cpp
@@ -52,10 +52,6 @@ FluidPressureDisplay::FluidPressureDisplay() : point_cloud_common_(new PointClou
       " Increasing this is useful if your incoming TF data is delayed significantly "
       "from your FluidPressure data, but it can greatly increase memory usage if the messages are big.",
       this, SLOT(updateQueueSize()));
-
-  // PointCloudCommon sets up a callback queue with a thread for each
-  // instance.  Use that for processing incoming messages.
-  update_nh_.setCallbackQueue(point_cloud_common_->getCallbackQueue());
 }
 
 FluidPressureDisplay::~FluidPressureDisplay()
@@ -65,6 +61,9 @@ FluidPressureDisplay::~FluidPressureDisplay()
 
 void FluidPressureDisplay::onInitialize()
 {
+  // Use the threaded queue for processing of incoming messages
+  update_nh_.setCallbackQueue(context_->getThreadedQueue());
+
   MFDClass::onInitialize();
   point_cloud_common_->initialize(context_, scene_node_);
 
--- a/src/rviz/default_plugin/illuminance_display.cpp
+++ b/src/rviz/default_plugin/illuminance_display.cpp
@@ -52,10 +52,6 @@ IlluminanceDisplay::IlluminanceDisplay() : point_cloud_common_(new PointCloudCom
       " Increasing this is useful if your incoming TF data is delayed significantly "
       "from your Illuminance data, but it can greatly increase memory usage if the messages are big.",
       this, SLOT(updateQueueSize()));
-
-  // PointCloudCommon sets up a callback queue with a thread for each
-  // instance.  Use that for processing incoming messages.
-  update_nh_.setCallbackQueue(point_cloud_common_->getCallbackQueue());
 }
 
 IlluminanceDisplay::~IlluminanceDisplay()
@@ -65,6 +61,9 @@ IlluminanceDisplay::~IlluminanceDisplay()
 
 void IlluminanceDisplay::onInitialize()
 {
+  // Use the threaded queue for processing of incoming messages
+  update_nh_.setCallbackQueue(context_->getThreadedQueue());
+
   MFDClass::onInitialize();
   point_cloud_common_->initialize(context_, scene_node_);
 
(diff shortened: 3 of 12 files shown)


simonschmeisser:
Disclaimer first: I'm writing this from my mobile at Lake Garda ... Will be back in one week for more detailed review

You could use Qt's object tree for cleanup:
timer_ = new QTimer(this);

Will be automatically deleted in the destructor of this (calling delete manually as well is fine as it can deregister itself at the parent)

Use 'new' stile connect (Qt5 or newer):
connect(timer_, &QTimer::timeout, this, &VisManager::onUpdate);

Now to the actual issue: does receiving a PointCloud actually trigger an Ogre rendering? I don't fully understand why it was disabled in 2013 and if we still need this at all?

If we do still need this, your PR looks fine and I cannot think of obvious issues. Haven't tested anything though ...

#868 does not sound like a memory leak but just an unbounded queue. Doesn't the PointCloudDisplay already have a queue size that could be used to discard old messages like suggested in the comment?

Am Donnerstag, 17. Juni 2021 schrieb Robert Haschke:
> VisualizationManager uses `[start|stop]Update()` to pause ROS updates during dialog box interaction (#631).
> However, asynchronous updates via `threaded_queue_` were still performed, leading to a memory leak (#1621),
> because messages were partially processed by the thread (i.e. kept in memory), but not finally processed
> (and cleaned up) by the synchronous `Display::update()` call.
> Pausing threaded processing too resolves that issue.
> 
> Fixes #1606. Fixes #868.
> @simonschmeisser: Do you see any issue pausing asynchronous message processing too while displaying dialogs?
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/ros-visualization/rviz/pull/1635
> 
> -- Commit Summary --
> 
>   * Stop asynchronous ROS updates with synchronous ones
> 
> -- File Changes --
> 
>     M src/rviz/visualization_manager.cpp (17)
> 
> -- Patch Links --
> 
> https://github.com/ros-visualization/rviz/pull/1635.patch
> https://github.com/ros-visualization/rviz/pull/1635.diff
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/ros-visualization/rviz/pull/1635
>

-- 
Gesendet von meinem Sailfish Gerät

rhaschke:
First, I didn't modernize any code (e.g. switching to Qt's object tree instead of manual new/delete or using Qt5-style signal+slots) and I don't think it is worth the effort.

> I don't fully understand why it was disabled in 2013 and if we still need this at all?

Me neither. There was an issue back than: #629, which was solved by disabling message processing while showing dialogues.
I thought it would be safer to stick with the current behaviour. But, we could try to revert that.

> #868 does not sound like a memory leak but just an unbounded queue. Doesn't the PointCloudDisplay already have a queue size that could be used to discard old messages like suggested in the comment?

You are right. It's an unbounded queue "only". However, the queue grows rather fast and thus "feels" like a mem leak at first sight. There are several queues involved: First, there is the subscriber message queue, which size is controlled by the display. Second, there is an internal queue to pass partially processed clouds from the processing thread to the main rendering thread. Its size isn't controlled (and grew without bounds).
My solution - in contrast to limiting the queue (https://github.com/ros-visualization/rviz/issues/868#issuecomment-89044085) - fixes the issue even deeper: The messages shouldn't be processed at all - not only in `PointCloudCommon`, but in general by all displays employing `threaded_nh_`.

Issue #1186, opened 2018-01-23
Z-Buffer issues when rendering PointCloud2


HWiese1980:
Hi everyone,

I've been experiencing some weird z-buffer errors wrt. PointCloud2 rendering recently that I don't know what causes them and how to solve the problem. You can see the effect [here](https://gfycat.com/BronzeVengefulAsianconstablebutterfly). It doesn't look like the point cloud itself is wrong. The points are where they should be. But they somehow from some certain perspectives appear to be rendered in the wrong z order. 

From the About box: RViz version is 1.12.15 (Kinetic), OGRE is 1.9.0 (Ghadamon), QT is 5.5.1. OS is Ubuntu 16.04 LTS.

wjwwood:
I've seen something like that before where it's actually caused by clipping of the viewport box when you have a strange reference frame.

I don't have time to look into in detail right now, but if you could provide a small bag file and an rviz config that reproduces the problem, at least someone else could reproduce the issue locally.

HWiese1980:
Okay, I'll see what I can do regarding the bag file.

HWiese1980:
Alright, so here are the corresponding rviz and bag files: 

https://www.dropbox.com/s/lt5z0sxxs3mwtnd/rviz_bag.tar.gz

You can (if at all) best see the effect in Orbit view starting with the following view data:

Distance: 8.3
Yaw: 3.95
Pitch: 0.35
Focal Point: -0.19 2.41 -0.37
Near Clip: 0.01
Focal Shape Size 0.05
Focal Shape Fixed Size: yes
Target Frame: <fixed frame>
Invert Z-Axis: no

and using the mouse moving to Pitch: 0,57

The effect is especially visible at the end of the bag file as "head of the depicted person (me) disappearing behind wall"

wjwwood:
Thanks. Again, I don't have time to look into it right now, but having this to help reproduce increases the chance someone else will have a look and will help me when I eventually have time to look into it.

HWiese1980:
Alright, hope to get some help soon... it's quite disruptive...

HWiese1980:
I figured that it only happens when *Alpha* of the rendered point cloud is set to something less than 1.

rhaschke:
I can reproduce this issue. #1644 / #1656 are _not_ fixing this issue - so it's not related to bounding box / bounding sphere calculation. Due to the fact that `Alpha` needs to be less than 1, it looks like a basic rendering issue. 
[...]

PR #1644, opened 2021-07-01
Fix misc point cloud ogre helper bounds issues

c-andy-martin:
There were a few circumstances where the point cloud ogre helper calculated incorrect bounds. These circumstances were mostly noticed when using the octomap_rviz_plugins to visualize octomaps where the maps would disappear when viewed from certain angles or zoom levels, or when using a non-unity alpha the different prune depths of the tree would render out-of-depth-order due to the bug in getSquaredViewDepth. Sorry I don't have any visuals right now to post with the PR as these commits were made on our branch of rviz a couple years ago and I'm just now getting a bit of time to send the fixes upstream.

### DIFF
--- a/src/rviz/ogre_helpers/point_cloud.cpp
+++ b/src/rviz/ogre_helpers/point_cloud.cpp
@@ -540,7 +540,6 @@ void PointCloud::addPoints(Point* points, uint32_t num_points)
     }
 
     aabb.merge(p.position);
-    bounding_radius_ = std::max(bounding_radius_, p.position.squaredLength());
 
     float x = p.position.x;
     float y = p.position.y;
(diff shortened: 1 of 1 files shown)


c-andy-martin:
One of these commits (the getSquaredViewDepth) may fix #1186

rhaschke (commented):
Thanks a lot for this contribution. While I agree that the original implementation is erroneous, I don't yet agree with your solution approach, which seems to overestimate the bounding box radius (see below). I started an alternative PR: #1656.
Also, I wasn't able to reproduce the actual issue. However, given a few issue reports, I agree that it exists. Could you, please, provide a simple example (ideally a cloud with a few points only that are generated programmatically) illustrating the issue?

PR #1654, opened 2021-08-18
Add skeleton import functionality to mesh_loader

kar-re:
### Description

This adds the functionality of importing an animated mesh to RViz using the same resource_retriever as in mesh_loader. 

Due to the way Ogre3D loads skeleton meshes a developer can automatically import an animated mesh using the same loadMeshFromResource method already present, but with the added functionality of assigning animation states.
The method is mostly the same as loadMeshFromResource but with changes to load the .skeleton file as well into the default resource group. 

Happy to accept any constructive critiques or suggestions on how to make it better!

### Checklist

- [ ] If you are changing GUI, please include screenshots showing how things looked *before* and *after*.
- [Will do if accepted! ] Did you change how RViz works? Added new functionality? Do not forget to update the tutorials and/or documentation on the [ROS wiki](http://wiki.ros.org/rviz)
- [x] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-visualization/rviz/pulls) to support the maintainers of RViz. Refer to the [RViz Wiki](https://github.com/ros-visualization/rviz/wiki/Maintainer-Guide#reviewing-pull-requests) for reviewing guidelines.

### DIFF
--- a/src/rviz/mesh_loader.cpp
+++ b/src/rviz/mesh_loader.cpp
@@ -704,7 +704,7 @@ Ogre::MeshPtr loadMeshFromResource(const std::string& resource_path)
       {
         return Ogre::MeshPtr();
       }
-      loadSkeletonFromResource(resource_path);  // load skeleton to the resource manager
+      loadSkeletonFromResource(resource_path); // load skeleton to the resource manager
 
       Ogre::MeshSerializer ser;
       Ogre::DataStreamPtr stream(new Ogre::MemoryDataStream(res.data.get(), res.size));
--- a/src/rviz/mesh_loader.cpp
+++ b/src/rviz/mesh_loader.cpp
@@ -28,11 +28,15 @@
  */
 
 #include "mesh_loader.h"
+#include <OGRE/OgrePrerequisites.h>
 #include <resource_retriever/retriever.h>
 
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
 
+#include <OGRE/OgreSkeleton.h>
+#include <OGRE/OgreSkeletonManager.h>
+#include <OGRE/OgreSkeletonSerializer.h>
 #include <OGRE/OgreMeshManager.h>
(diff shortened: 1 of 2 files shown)


src/rviz/mesh_loader.cpp
```diff
@@ -699,6 +704,7 @@ Ogre::MeshPtr loadMeshFromResource(const std::string& resource_path)
       {
         return Ogre::MeshPtr();
       }
+      Ogre::SkeletonPtr skeleton = loadSkeletonFromResource(resource_path);
```
rhaschke, line 707: Where / how is this `skeleton` used?

rhaschke (commented):
I don't yet see the use of this PR. Could you explain more, please?
According to the [OGRE doc](https://www.ogre3d.org/docs/api/1.9/class_ogre_1_1_skeleton.html#details), skeleton definitions are loaded on demand, especially when referenced by a Mesh. So, do we really need to explicitly load them?

How do you intend animating a mesh based on the skeleton? The `SkeletonPtr` isn't used yet.

src/rviz/mesh_loader.cpp
rhaschke:
```suggestion
      loadSkeletonFromResource(resource_path);  // load skeleton to the resource manager
```

src/rviz/mesh_loader.cpp
```diff
@@ -699,6 +704,7 @@ Ogre::MeshPtr loadMeshFromResource(const std::string& resource_path)
       {
         return Ogre::MeshPtr();
       }
+      loadSkeletonFromResource(resource_path);  // load skeleton to the resource manager
```
rhaschke, line 707:
```suggestion
      loadSkeletonFromResource(resource_path); // load skeleton to the resource manager
```

kar-re:
Hi @rhaschke, just wanted to say thanks for implementing the code review change, and sorry for the late response. Please let me know if there's anything else I can do! :)

PR #1655, opened 2021-08-20
Smoothly update pose of point cloud on update()

rhaschke:
This is the migration of #1652 to the `melodic` branch. The PR addresses the problem that a point cloud isn't smoothly moving if its `frame_id` is moving w.r.t. rviz' fixed frame.
The solution was to update the pose of the corresponding scene node in each `update()` cycle, not only when the message was received.


rhaschke:
I had to revert this PR due to issues in normal uses cases: #1666.

Issue #1627, opened 2021-05-11
feature: show frame trails


christian-rauch:
This is a feature request to add an option to show trails for TF frames and frame axes.

This already exists for the links in the [`RobotModel`](http://wiki.ros.org/rviz/DisplayTypes/RobotModel) (property `Show Trail`). It would be useful if the same functionality could be provided for arbitrary frames in the TF tree or the Axes widget.

christian-rauch:
Sorry, I didn't mean to put pressure on you to implement such a feature. I also don't plan to contribute any more to the ROS1 ecosystem. I will keep this feature on my TODO list for when I eventually switch to ROS2.

PR #1677, opened 2021-11-02
frame axes trail

christian-rauch:
### Description

Add the option to add a trail (`Ogre::RibbonTrail`) to the frame axes display. This replicates the "Show Trail" features from the `RobotModelDisplay`. This implements and fixes https://github.com/ros-visualization/rviz/issues/1627.

The ribbon trail will show the path a frame took. However, when the frame axes and ribbon trail is initialised, it will show a trail from the origin to where the frame is spawned. I am not sure if this intended behaviour that is also present in the `RobotModelDisplay` view.

Edi: I think the problem simply stems from the visualisation of the axes scene node which is rendered at the origin before the first frame pose has been received.

### Checklist

rviz_screenshot_2021_11_02-11_03_42

rviz_screenshot_2021_11_02-11_04_27

### DIFF
--- a/src/rviz/default_plugin/axes_display.cpp
+++ b/src/rviz/default_plugin/axes_display.cpp
@@ -31,6 +31,7 @@
 
 #include <OGRE/OgreSceneNode.h>
 #include <OGRE/OgreSceneManager.h>
+#include <OGRE/OgreRibbonTrail.h>
 
 #include <rviz/display_context.h>
 #include <rviz/frame_manager.h>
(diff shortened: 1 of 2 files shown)


rhaschke (approved):
Thanks for this contribution!

christian-rauch:
Feel free to squash your fixes into my commit.

@rhaschke Do you have an idea how this "jump" from the origin to the first frame pose can be eliminated?

rhaschke:
> Do you have an idea how this "jump" from the origin to the first frame pose can be eliminated?

I think I resolved that issue already in 371d3bd. In which use case do you still observe it?

christian-rauch:
> I think I resolved that issue already in [371d3bd](https://github.com/ros-visualization/rviz/commit/371d3bd170e46f365d1398170cf232588213df52). In which use case do you still observe it?

Yes. The frame itself is not shown before a pose is set. But once a pose is set, I still see a trail from the origin to that pose.

rhaschke:
Not for me (origin is of course at the base of the robot):
image

Issue #1665, opened 2021-09-15
Marker load mesh resource success, but not show in window?


thinkershare:
Environment:
* OS: Ubuntu 20.04
* ROS version: ROS noetic
* RVIZ version: 1.14.9(QT version 5.12.8, OGRE version 1.9.0(Ghadamon))
* Tip: Windows 11(wsl2 with wsl CUDA driver / 2060 super )

```python
#!/usr/bin/env python3
import rospy;
from visualization_msgs.msg import Marker;

if  __name__ == "__main__":
    frame = 0
    rospy.init_node('kitti_node',anonymous=True)
    model_pub = rospy.Publisher('car',Marker, queue_size=10)

    while not rospy.is_shutdown():
        marker = Marker()
        marker.header.frame_id = "map"
        marker.lifetime = rospy.Duration();

        marker.type = marker.MESH_RESOURCE
        marker.mesh_resource = "package://kitti/Car-Model/frame.dae"
        marker.action = marker.ADD
        marker.scale.x = 80
        marker.scale.y = 80
        marker.scale.z = 80

        marker.color.r = 0.4
        marker.color.g = 0.2
        marker.color.b = 0.8
        marker.color.a = 1

        marker.mesh_use_embedded_materials = False
        marker.pose.position.x, marker.pose.position.y, marker.pose.position.z = 0, 0, 0
        marker.pose.orientation.x, marker.pose.orientation.y, marker.pose.orientation.z, marker.pose.orientation.w = 0, 0, 0, 1
        model_pub.publish(marker)
        rospy.loginfo("kitti published")
```
![屏幕截图 2021-09-15 100654](https://user-images.githubusercontent.com/19685432/133358738-7e23d37c-5949-4c38-b830-35683bca2bda.jpg)

rhaschke:
Does it work in a non-WSL environment? If so, I suspect the WSL environment.
If not, check that the message is actually received by rviz. Are other shapes, e.g. boxes, spheres displayed?

thinkershare:
I don’t have a physical machine, I test shape, boxes, e.g. All this can be displayed normally.

rhaschke:
Did you try another mesh as well?

peci1:
[...]

PR #1682, opened 2021-11-22
Updated issue template to guide users to try running RViz with C locale

peci1:
Updated issue template to guide users to try running RViz with C locale. This could prevent (or pinpoint faster) issues like #1665.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -21,3 +21,4 @@ Describe your issue here and explain how to reproduce it.
     paste your console output here
     ```
 * If source build, which git commit?
+* System locale: e.g. `en_US.UTF-8` (get it by running `echo "$LANG $LC_NUMERIC"` on Linux; before reporting a rendering issue, try running RViz with `LANG=C rviz` and report whether the error is reproducible even in this case)
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -21,4 +21,5 @@ Describe your issue here and explain how to reproduce it.
     paste your console output here
     ```
 * If source build, which git commit?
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -21,3 +21,4 @@ Describe your issue here and explain how to reproduce it.
     paste your console output here
     ```
 * If source build, which git commit?
+* System locale: e.g. `en_US.UTF-8` (get it by running `echo "$LANG $LC_NUMERIC"` on Linux; before reporting a rendering issue, try running RViz with `LANG=C rviz` and report whether the error is reproducible even in this case)
```
rhaschke, line 24:
```suggestion
* System locale, i.e. the output of `echo "$LANG $LC_NUMERIC"`:
  Before reporting a rendering issue, try running RViz with `LANG=C rviz`!
```

rhaschke (commented):
Thanks for commenting on #1665. I'm surprised that this is still an issue as @simonschmeisser only yesterday confirmed that this error should be fixed (https://github.com/ros-planning/moveit/issues/2966#issuecomment-974889032). Can you provide more details, please?

peci1:
I also don't know which part brought this in this time. I remember not having problems with this in a few years. But yesterday I was again staring at an empty rviz for a while until I tried the `LANG=C` workaround and it helped. Up-to-date Ubuntu 18.04/Melodic on x86/64 CPU, AMD Renoir GPU and with cs_CZ locale.

simonschmeisser:
If your workspace is all from code a grep for `std::stof` or `std::stod` might tell

peci1:
Let's move the discussion back to #1665 .

rhaschke:
Merged into melodic-devel: b2386556b2f10de493f53ccb6db82e14c9707220

PR #1686, opened 2021-12-15
Add option to continuously transform visuals in odometry plugin

AndreasR30:
Hello,

I made a new pull request, which is essentially the same as #1631, which was merged but then reverted as it broke the visualization of other users. But now the changes are less invasive and do not affect old visualizations as long the option is unchecked (the default). The new option triggers the same behavior as the activated `frame_locked` flag in the [Marker Message](http://docs.ros.org/en/diamondback/api/visualization_msgs/html/msg/Marker.html).


Martin-Oehler:
This PR makes sense when you want to view the history of measurements as a time-independent path instead of time-specific measurements. In this case, the default behavior is not very useful, because it does not allow you to compare trajectories published in different frames: 

without_patch

With this PR, the history is re-transformed with the latest transform, allowing the comparison of trajectories.

with_patch

In principal, I support merging this PR, but like mentioned in the similar #1687, I would suggest changing the naming of the option to something like "Ignore timestamp" or "Use most recent timestamp" to be consistent with the Map Display ("Use timestamp", does the opposite).

Also, in my opinion, the changes regarding ref poses for high values should be split into a separate PR.

Best regards,
Martin

AndreasR30:
> This PR makes sense when you want to view the history of measurements as a time-independent path instead of time-specific measurements. In this case, the default behavior is not very useful, because it does not allow you to compare trajectories published in different frames:

I am happy to read that you see the advantages of this option for odometry display.

> In principal, I support merging this PR, but like mentioned in the similar #1687, I would suggest changing the naming of the option to something like "Ignore timestamp" or "Use most recent timestamp" to be consistent with the Map Display ("Use timestamp", does the opposite).

I can change the naming in order to be consistent. @rhaschke, what do you think?

> Also, in my opinion, the changes regarding ref poses for high values should be split into a separate PR.

I already split it into a separate commit. Why would you split it into a separate PR? Are you not fully convinced about this feature, yet? @rhaschke, do you also think it is useful to split it into a separate PR?

Issue #1688, opened 2021-12-15
Meshes rendering in the wrong orientation


civerachb-cpr:
It appears that there's a bug in Rviz 1.13.22 where links with STL or DAE visuals are being rendered incorrectly  The underlying links are correctly located, but the visual and collision geometry is a mess.

For example, here is the [Clearpath Warthog](https://github.com/warthog-cpr/warthog) model viewed using the command `roslaunch warthog_viz view_robot.launch`:

Rviz 1.13.18 | Rviz 1.13.22 | Gazebo 9.19.0 |
|-----------------|------------------|---------------------|
| ![warthog-correct](https://user-images.githubusercontent.com/59611394/146228563-67ae0f6c-d983-4130-aa89-b2b4c3fec31f.png) | ![warthog-chris](https://user-images.githubusercontent.com/59611394/146228648-dbb93702-958c-4ce2-812d-8b3f72da9fee.png) | ![warthog-gazebo](https://user-images.githubusercontent.com/59611394/146229283-5bd2cc78-3496-46fb-9ae3-1f6b8f6de1d7.png) |

Adding the RobotModel vizualization to Rviz results in the same mal-formed visual as above; the meshes are all oriented incorrectly in Rviz, even though Gazebo is publishing the TF and joint states correctly.

This also affects simpler models, like the [VLP-16 from velodyne_description](http://wiki.ros.org/velodyne_description)
![dingo_velodyne_tf](https://user-images.githubusercontent.com/59611394/146228815-1692c2d8-fd9c-400d-82fd-9571f12eb9ba.png)

You can see the Velodyne's base link and sensor origin frames are oriented vertically relative to each other, but the mesh itself is rotated off to the side.  The above was created with this included in the URDF for the robot:

```
  <xacro:include filename="$(find velodyne_description)/urdf/VLP-16.urdf.xacro" />
  <xacro:VLP-16 parent="front_mount" topic="velodyne_points">
    <origin xyz="0 0 0" rpy="0 0 0" />
  </xacro:VLP-16>
```

### environment
* OS Version: Ubuntu 18.04
* ROS Distro: Melodic
* RViz, Qt, OGRE, OpenGl version as printed by rviz:
    ```
    [ INFO] [1639587205.192206801]: rviz version 1.13.22
    [ INFO] [1639587205.192246967]: compiled against Qt version 5.9.5
    [ INFO] [1639587205.192257016]: compiled against OGRE version 1.9.0 (Ghadamon)
    ```
[...]

PR #1685, opened 2021-11-26
Drop proprietary COLLADA scaling

rhaschke:
As pointed out by @simonschmeisser in https://github.com/ros-visualization/rviz/pull/1684#issuecomment-977827800, `assimp` should actually provide the correct scaling factor and it actually does: The scaling is embedded in the root node's transform, which was deliberately ignored.
Fixes #1665. @simonschmeisser and @peci1, could you please review and test?

PR #1689, opened 2021-12-16
Fixup assimp mesh loading

rhaschke:
This is a fixup of #1685, fixing a regression revealed in #1688.

Looks like assimp's root node is only _sometimes_, but not always applying a rotation to y-up orientation.
This fixup, drops the rotation part of the root transform, but keeps the scaling (and position offset).

### DIFF
--- a/src/rviz/mesh_loader.cpp
+++ b/src/rviz/mesh_loader.cpp
@@ -222,16 +222,28 @@ void ResourceIOSystem::Close(Assimp::IOStream* stream)
  * already by loadMaterials(). */
 void buildMesh(const aiScene* scene,
                const aiNode* node,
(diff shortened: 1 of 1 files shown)


knorth55:
I checked this PR, and this PR fixes #1688 

Before: rviz `1.13.22`
Screenshot from 2021-12-16 19-06-06

After:  this PR
Screenshot from 2021-12-16 19-12-45

simonschmeisser (approved):
Sorry for doing a source-only review last time. Can confirm this was broken before and works now. Wasn't that hard to run `git rebase --onto origin/noetic-devel origin/melodic-devel` to be able to test it on noetic after all ...

Issue #1694, opened 2022-01-11
Feature request: robot_description topic


Timple:
To ease the transition to ROS2 we would like to use a `robot_description` topic instead of a parameter. 

Would a PR get accepted which expands the model view with a topic option?

Implementation would be an enum box which chooses between Parameter (default) or Topic.

130s:
Is setting robot description ROS Parameter on ROS2 side not an option? Just wondering, but in one of a ros1-2 bridge projects I think I saw that (robot running on ros1, then there's a bridge to receive some in ros2, but in ros2 redundantly set robot description). It's not nice but worked. Just fyi.

Timple:
I've created a plugin which solves this issue: https://github.com/nobleo/rviz_robot_description_topic

PR #1695, opened 2022-01-11
Robot description topic

Timple:
### Description

To ease the transition to ROS2 we would like to use a robot_description topic instead of a parameter. Solves #1694 

### Checklist

- [x] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-visualization/rviz/pulls) to support the maintainers of RViz. Refer to the [RViz Wiki](https://github.com/ros-visualization/rviz/wiki/Maintainer-Guide#reviewing-pull-requests) for reviewing guidelines.

image image

### DIFF
--- a/src/rviz/default_plugin/robot_model_display.cpp
+++ b/src/rviz/default_plugin/robot_model_display.cpp
@@ -37,6 +37,7 @@
 #include <rviz/robot/robot.h>
 #include <rviz/robot/robot_link.h>
 #include <rviz/robot/tf_link_updater.h>
+#include <rviz/properties/enum_property.h>
 #include <rviz/properties/float_property.h>
 #include <rviz/properties/property.h>
 #include <rviz/properties/string_property.h>
@@ -52,6 +53,11 @@ void linkUpdaterStatusFunction(StatusProperty::Level level,
 {
   display->setStatus(level, QString::fromStdString(link_name), QString::fromStdString(text));
 }
+
+enum DescriptionSource
+{
+  PARAMETER, TOPIC
+};
 
 RobotModelDisplay::RobotModelDisplay()
   : Display(), has_new_transforms_(false), time_since_last_transform_(0.0f)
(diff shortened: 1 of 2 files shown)


rhaschke (commented):
I am hesitant to merge this as the robot description is handled as a parameter throughout ROS1.
Changing it to a topic requires lots of changes elsewhere as well.
I don't yet see how the availability of the topic parameter facilitates migration to ROS2. Can't you always run rviz2 to have this functionality there?

Timple:
> how the availability of the topic parameter facilitates migration to ROS2.

Well, having it as a topic allows it to pass the `ros1_bridge`. Parameters don't do that. We plan to port our nodes 1-by-1 to the ROS2 side of the bridge.

> requires lots of changes elsewhere

For us not that much actually. `robot_state_publisher` would need to publish (or read) a topic as well. That's on my todo-list.

All places where we now parse a urdf, we simply make a subscription.

> Can't you always run rviz2

Probably, but the bridge is not in place yet on the real robot (risk). We're in the preparation stage. For large projects this could really take a while I'm affraid.

---

But most of all, this allows us to leverage the force of a topic already in ROS1. Because our robot is becoming more-and-more dynamic, we need restarts on changes in the urdf.

peci1:
There have already been several attempts at dynamically changing the robot model. See https://github.com/ros/robot_state_publisher/pull/31 and https://github.com/peci1/dynamic_robot_state_publisher as a workaround until that PR is merged. These approaches use a `Trigger`-type service to force a reload of the parameter (with dynamic_robot_state_publisher also supporting dynamic_reconfigure, which is basically subscription to a topic). Wouldn't adding a `reload_robot_models` service solve this issue? You could then write a simple node that would listen on the model topic and update the param and call the trigger service of rviz and robot_state_publisher.

Timple:
Thank you for pointing me in this direction and thinking along!

From the activity, it doesn't look like https://github.com/ros/robot_state_publisher/pull/31 is going to be merged any sooner than this.

But my rationale of using a topic is to align already with ROS2. They probably (assumption here) thought about deeply and deemed a topic the right solution.
For me this also makes sense: it can be latched, nodes with a subscriber get the new model immediately. Instead of receiving a trigger that they should reload a parameter.

rhaschke:
Closing now. This feature won't merge into ROS1.

Timple:
For future reference: https://github.com/nobleo/rviz_robot_description_topic

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/announcing-rviz-robot-description-topic/42417/1

PR #1696, opened 2022-01-11
Fix template

Timple:


rhaschke:
Thanks for this hint. I updated the issue template to mention both, Noetic and Melodic: 62ba5cd26344dceeda1b66ada5dd0740f218cbf3

PR #1698, opened 2022-01-19
Add new time sync mode SyncFrame

rhaschke:
... to sync TF lookups within the current rendering frame.
Fixes #1693. @AndreasR30, please review.

 @dgossow, You implemented the _experimental_ `TimePanel` functionality 9 years ago. Unfortunately, there are no PR or code comments for the changeset 8c4cccf54a4581deff9d806d401b18a1a4411854 and I had a hard time guessing the intended functionality.
I understood that `SyncExact` and `SyncApprox` are intended to sync against a time signal provided by a specific source `Display`. For example, all displays deriving from `PointCloudCommon` call `emitTimeSignal()` and thus allow this functionality. However, I didn't get the intention of the `SyncApprox` mode. TF lookups perform an approximate lookup anyway, aren't they? So what's the role of `SyncApprox`?
As this mode essentially doesn't work (throwing lookup errors on the console), I'm tempted to remove it completely.

https://user-images.githubusercontent.com/5376030/150152719-9fe87b12-3633-470f-979e-e6c372274971.mp4

### DIFF
--- a/src/rviz/frame_manager.cpp
+++ b/src/rviz/frame_manager.cpp
@@ -59,32 +59,30 @@ FrameManager::~FrameManager()
 
 void FrameManager::update()
 {
-  boost::mutex::scoped_lock lock(cache_mutex_);
-  if (!pause_)
+  if (pause_)
+    return;
+  else
   {
+    boost::mutex::scoped_lock lock(cache_mutex_);
     cache_.clear();
-  }
-
-  if (!pause_)
-  {
     switch (sync_mode_)
     {
-    case SyncOff:
+    case SyncOff: // always use latest time
       sync_time_ = ros::Time::now();
       break;
-    case SyncExact:
+    case SyncExact: // sync to external source
+      // sync_time_ set via syncTime()
       break;
     case SyncApprox:
-      // adjust current time offset to sync source
-      current_delta_ = 0.7 * current_delta_ + 0.3 * sync_delta_;
-      try
-      {
-        sync_time_ = ros::Time::now() - ros::Duration(current_delta_);
-      }
-      catch (...)
-      {
-        sync_time_ = ros::Time::now();
-      }
+      // sync_delta is a sliding average of current_delta_, i.e.
+      // approximating the average delay of incoming sync messages w.r.t. current time
+      sync_delta_ = 0.7 * sync_delta_ + 0.3 * current_delta_;
+      // date back sync_time_ to ensure finding TFs that are as old as now() - sync_delta_
+      sync_time_ = ros::Time::now() - ros::Duration(sync_delta_);
+      break;
+    case SyncFrame: // sync to current time
+      // date back sync_time_ to ensure finding TFs that are as old as now() - sync_delta_
+      sync_time_ = ros::Time::now() - ros::Duration(sync_delta_);
       break;
     }
   }
@@ -98,7 +96,6 @@ void FrameManager::setFixedFrame(const std::string& frame)
     if (fixed_frame_ != frame)
     {
       fixed_frame_ = frame;
-      fixed_frame_id_ = tf_buffer_->_lookupFrameNumber(fixed_frame_);
       cache_.clear();
       should_emit = true;
     }
@@ -119,8 +116,8 @@ void FrameManager::setSyncMode(SyncMode mode)
 {
   sync_mode_ = mode;
   sync_time_ = ros::Time(0);
-  current_delta_ = 0;
   sync_delta_ = 0;
+  current_delta_ = 0;
 }
 
 void FrameManager::syncTime(ros::Time time)
@@ -128,6 +125,7 @@ void FrameManager::syncTime(ros::Time time)
   switch (sync_mode_)
   {
   case SyncOff:
+  case SyncFrame:
     break;
   case SyncExact:
     sync_time_ = time;
@@ -135,13 +133,13 @@ void FrameManager::syncTime(ros::Time time)
   case SyncApprox:
     if (time == ros::Time(0))
     {
-      sync_delta_ = 0;
+      current_delta_ = 0;
       return;
     }
-    // avoid exception due to negative time
-    if (ros::Time::now() >= time)
+    if (ros::Time::now() >= time) // avoid exception due to negative time
     {
-      sync_delta_ = (ros::Time::now() - time).toSec();
+      // estimate delay of sync message w.r.t. current time
+      current_delta_ = (ros::Time::now() - time).toSec();
     }
     else
     {
@@ -151,48 +149,22 @@ void FrameManager::syncTime(ros::Time time)
   }
 }
 
-bool FrameManager::adjustTime(const std::string& frame, ros::Time& time)
+void FrameManager::adjustTime(ros::Time& time)
 {
   // we only need to act if we get a zero timestamp, which means "latest"
   if (time != ros::Time())
-  {
-    return true;
-  }
+    return;
 
   switch (sync_mode_)
   {
   case SyncOff:
     break;
+  case SyncFrame:
   case SyncExact:
+  case SyncApprox:
     time = sync_time_;
     break;
-  case SyncApprox:
-  {
-    // if we don't have tf info for the given timestamp, use the latest available
-    ros::Time latest_time;
-    std::string error_string;
-    int error_code;
-    if (fixed_frame_id_ == 0) // we couldn't resolve the fixed_frame_id yet
-      fixed_frame_id_ = tf_buffer_->_lookupFrameNumber(fixed_frame_);
-
-    error_code = tf_buffer_->_getLatestCommonTime(fixed_frame_id_, tf_buffer_->_lookupFrameNumber(frame),
-                                                  latest_time, &error_string);
-
-    if (error_code != 0)
-    {
-      ROS_ERROR("Error getting latest time from frame '%s' to frame '%s': %s (Error code: %d)",
-                frame.c_str(), fixed_frame_.c_str(), error_string.c_str(), error_code);
-      return false;
-    }
-
-    if (latest_time > sync_time_)
-    {
-      time = sync_time_;
-    }
   }
-  break;
-  }
-  return true;
 }
 
 
@@ -201,10 +173,7 @@ bool FrameManager::getTransform(const std::string& frame,
                                 Ogre::Vector3& position,
                                 Ogre::Quaternion& orientation)
 {
-  if (!adjustTime(frame, time))
-  {
-    return false;
-  }
+  adjustTime(time);
 
   boost::mutex::scoped_lock lock(cache_mutex_);
 
@@ -243,10 +212,7 @@ bool FrameManager::transform(const std::string& frame,
                              Ogre::Vector3& position,
                              Ogre::Quaternion& orientation)
 {
-  if (!adjustTime(frame, time))
-  {
-    return false;
-  }
+  adjustTime(time);
 
   position = Ogre::Vector3::ZERO;
   orientation = Ogre::Quaternion::IDENTITY;
@@ -261,7 +227,20 @@ bool FrameManager::transform(const std::string& frame,
   {
     tf2::doTransform(pose, pose, tf_buffer_->lookupTransform(fixed_frame_, frame, time));
   }
-  catch (std::runtime_error& e)
+  catch (const tf2::ExtrapolationException& e)
+  {
+    if (sync_mode_ == SyncApprox)
+      return false;
+    // We don't have tf info for sync_time_. Reset sync_time_ to latest available time of frame
+    auto tf = tf_buffer_->lookupTransform(frame, frame, ros::Time());
+    if (sync_time_ > tf.header.stamp && tf.header.stamp != ros::Time())
+    {
+      sync_delta_ += (sync_time_ - tf.header.stamp).toSec(); // increase sync delta by observed amount
+      sync_time_ = tf.header.stamp;
+    }
+    return false;
+  }
+  catch (const std::runtime_error& e)
   {
     ROS_DEBUG("Error transforming from frame '%s' to frame '%s': %s", frame.c_str(),
               fixed_frame_.c_str(), e.what());
@@ -292,10 +271,7 @@ bool FrameManager::frameHasProblems(const std::string& frame, ros::Time /*time*/
 
 bool FrameManager::transformHasProblems(const std::string& frame, ros::Time time, std::string& error)
 {
-  if (!adjustTime(frame, time))
-  {
-    return false;
-  }
+  adjustTime(time);
 
   std::string tf_error;
   bool transform_succeeded = tf_buffer_->canTransform(fixed_frame_, frame, time, &tf_error);
--- a/src/rviz/frame_manager.h
+++ b/src/rviz/frame_manager.h
@@ -67,9 +67,10 @@ class FrameManager : public QObject
 public:
   enum SyncMode
   {
-    SyncOff = 0,
-    SyncExact,
-    SyncApprox
+    SyncOff = 0, // use latest TF updates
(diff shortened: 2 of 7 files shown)


src/rviz/frame_manager.cpp
```diff
@@ -77,10 +77,12 @@ void FrameManager::update()
     cache_.clear();
     switch (sync_mode_)
     {
-    case SyncOff:
+    case SyncOff:   // always use latest time
+    case SyncFrame: // sync to current time
```
AndreasR30, line 81: The problem here is when the TF messages have a small latency w.r.t. ros::Time::now(), which is usually the case in real world scenarios, I think. Then it is likely that one gets continuously extrapolation exceptions in lookupTransform. I think, `sync_time_` should be set to the stamp of the latest available TF. Unfortunately, there are many TFs with different latest stamps.
One way to mitigate this problem would be to check whether it is possible to transform with ros::Time::now() (e.g. with canTransform or lookupTransform + try-catch) in `transform()`. If there is a extrapolation exception fall back to ros::Time(). However, this causes overhead because of additional lookups. And it does not handle the case where a TF message arrives, which is still before ros::Time::now(), during rviz's update cycle.

AndreasR30:
Hello @rhaschke,
Thank you for this pull request. Unfortunately I don't think that it works as intended it real world scenarios.

I was thinking about to use the cache_ in getTransform to solve this problem. However, this would not handle the case correctly:

There is a TF tree `odom -> base_link -> chassis`, where `base_link` and `chassis` are fixed. Then the cache_ would make a snapshot of `odom -> base_link`. But when `odom -> chassis` is requested then there is a cache_ miss and a TF with a different stamp could be obtained in `lookupTransform()`.

I can't currently think of a way to solve this here in `TimePanel`/`FrameManager`. What about using #1692 and make it only an option in the `Global Options`?

rhaschke:
> Unfortunately I don't think that it works as intended it real world scenarios

Could you please provide python code to mimic your real-world scenario?

> What about using #1692 and make it only an option in the Global Options?

I would prefer to find a solution that processes TFs in a separate thread.

AndreasR30:
> I would prefer to find a solution that processes TFs in a separate thread.

Yes we can use the multi-threaded TransformListener by default.

> Could you please provide python code to mimic your real-world scenario?

Here you go:

```python
#!/usr/bin/env python

from __future__ import print_function

import rospy
import math

import tf2_ros
from geometry_msgs.msg import TransformStamped

if __name__ == "__main__":
    rospy.init_node("send_tf_v2")

    static_tf_msg = TransformStamped()
    static_tf_msg.header.stamp = rospy.Time.now()
    static_tf_msg.header.frame_id = "base_link"
    static_tf_msg.child_frame_id = "chassis"
    static_tf_msg.transform.rotation.w = 1
    br_static = tf2_ros.StaticTransformBroadcaster()
    br_static.sendTransform(static_tf_msg)

    br = tf2_ros.TransformBroadcaster()
    rate = rospy.Rate(100)
    radius = 2
    t = 0
    while not rospy.is_shutdown():
        print("sending...")
        for m in range(1, 10):
            tf_msg = TransformStamped()
            tf_msg.header.stamp = rospy.Time.now()
            tf_msg.header.frame_id = 'map'
            tf_msg.child_frame_id = 'base_link'
            tf_msg.transform.translation.x = radius * math.cos(t * 10)
            tf_msg.transform.translation.y = radius * math.sin(t * 10)
            tf_msg.transform.translation.z = 0
            tf_msg.transform.rotation.w = 1
            br.sendTransform(tf_msg)
        t += 0.01
        rate.sleep()
```

And Rviz Config:
```
Panels:
  - Class: rviz/Displays
    Help Height: 78
    Name: Displays
    Property Tree Widget:
      Expanded:
        - /Global Options1
        - /AxesChassis1/Status1
      Splitter Ratio: 0.5
    Tree Height: 839
  - Class: rviz/Selection
    Name: Selection
  - Class: rviz/Tool Properties
    Expanded:
      - /2D Pose Estimate1
      - /2D Nav Goal1
      - /Publish Point1
    Name: Tool Properties
    Splitter Ratio: 0.5886790156364441
  - Class: rviz/Views
    Expanded:
      - /Current View1
    Name: Views
    Splitter Ratio: 0.5
  - Class: rviz/Time
    Experimental: true
    Name: Time
    SyncMode: 0
    SyncSource: ""
Preferences:
  PromptSaveOnExit: true
Toolbars:
  toolButtonStyle: 2
Visualization Manager:
  Class: ""
  Displays:
    - Alpha: 1
      Class: rviz/Axes
      Enabled: true
      Length: 1
      Name: AxesBaseLink
      Radius: 0.10000000149011612
      Reference Frame: base_link
      Show Trail: false
      Value: true
    - Class: rviz/TF
      Enabled: true
      Frame Timeout: 15
      Frames:
        All Enabled: true
        base_link:
          Value: true
        chassis:
          Value: true
        map:
          Value: true
      Marker Alpha: 1
      Marker Scale: 1
      Name: TF
      Show Arrows: true
      Show Axes: true
      Show Names: true
      Tree:
        map:
          base_link:
            chassis:
              {}
      Update Interval: 0
      Value: true
    - Alpha: 0.5
      Cell Size: 0.20000000298023224
      Class: rviz/Grid
      Color: 160; 160; 164
      Enabled: true
      Line Style:
        Line Width: 0.029999999329447746
        Value: Lines
      Name: Grid
      Normal Cell Count: 0
      Offset:
        X: 0
        Y: 0
        Z: 0
      Plane: XY
      Plane Cell Count: 1000
      Reference Frame: <Fixed Frame>
      Value: true
    - Alpha: 1
      Class: rviz/Axes
      Enabled: true
      Length: 1
      Name: AxesChassis
      Radius: 0.10000000149011612
      Reference Frame: chassis
      Show Trail: false
      Value: true
  Enabled: true
  Global Options:
    Background Color: 48; 48; 48
    Default Light: true
    Fixed Frame: map
    Frame Rate: 30
  Name: root
  Tools:
    - Class: rviz/Interact
      Hide Inactive Objects: true
    - Class: rviz/MoveCamera
    - Class: rviz/Select
    - Class: rviz/FocusCamera
    - Class: rviz/Measure
    - Class: rviz/SetInitialPose
      Theta std deviation: 0.2617993950843811
      Topic: /initialpose
      X std deviation: 0.5
      Y std deviation: 0.5
    - Class: rviz/SetGoal
      Topic: /move_base_simple/goal
    - Class: rviz/PublishPoint
      Single click: true
      Topic: /clicked_point
  Value: true
  Views:
    Current:
      Class: rviz/Orbit
      Distance: 10.879761695861816
      Enable Stereo Rendering:
        Stereo Eye Separation: 0.05999999865889549
        Stereo Focal Distance: 1
        Swap Stereo Eyes: false
        Value: false
      Field of View: 0.7853981852531433
      Focal Point:
        X: 0
        Y: 0
        Z: 0
      Focal Shape Fixed Size: false
      Focal Shape Size: 0.05000000074505806
      Invert Z Axis: false
      Name: Current View
      Near Clip Distance: 0.009999999776482582
      Pitch: 0.7853981852531433
      Target Frame: base_link
      Yaw: 0.7853981852531433
    Saved: ~
Window Geometry:
  Displays:
    collapsed: false
  Height: 1136
  Hide Left Dock: false
  Hide Right Dock: false
  QMainWindow State: 000000ff00000000fd000000040000000000000156000003d2fc0200000008fb0000001200530065006c0065006300740069006f006e00000001e10000009b0000005c00fffffffb0000001e0054006f006f006c002000500072006f007000650072007400690065007302000001ed000001df00000185000000a3fb000000120056006900650077007300200054006f006f02000001df000002110000018500000122fb000000200054006f006f006c002000500072006f0070006500720074006900650073003203000002880000011d000002210000017afb000000100044006900730070006c006100790073010000003d000003d2000000c900fffffffb0000002000730065006c0065006300740069006f006e00200062007500660066006500720200000138000000aa0000023a00000294fb00000014005700690064006500530074006500720065006f02000000e6000000d2000003ee0000030bfb0000000c004b0069006e0065006300740200000186000001060000030c00000261000000010000010f000003d2fc0200000003fb0000001e0054006f006f006c002000500072006f00700065007200740069006500730100000041000000780000000000000000fb0000000a00560069006500770073010000003d000003d2000000a400fffffffb0000001200530065006c0065006300740069006f006e010000025a000000b200000000000000000000000200000490000000a9fc0100000001fb0000000a00560069006500770073030000004e00000080000002e10000019700000003000007380000003efc0100000002fb0000000800540069006d00650100000000000007380000031b00fffffffb0000000800540069006d00650100000000000004500000000000000000000004c7000003d200000004000000040000000800000008fc0000000100000002000000010000000a0054006f006f006c00730100000000ffffffff0000000000000000
  Selection:
    collapsed: false
  Time:
    collapsed: false
  Tool Properties:
    collapsed: false
  Views:
    collapsed: false
  Width: 1848
  X: 72
  Y: 27
```
Using this synthetic example you see both. The TF race condition problem when synchronization is `Off` and the execptions when `Frame` in time panel.

AndreasR30:
The line `for m in range(1, 10):` is actually not necessary to reproduce the problems. I just accidentally copied it from `send_tf.py`.

src/rviz/frame_manager.cpp
rhaschke: I see your point. I mitigated that problem by substracting a `sync_delta_` from `ros::Time::now()`, which is adapted when a frame lookup fails with an extrapolation exception.

rhaschke:
Thanks for providing the example code. I think I found a good solution now using an adaptable `sync_delta_`. Of course, during the first few iterations, there might be failing frame lookups, but `sync_delta_` should adapt rather quickly.

AndreasR30:
@rhaschke 

I had not time to test your latest commits yet. But I can tell you where I use SyncApprox in order to understand it better.

SyncApprox can be useful to delay the Transforms with ros::Time(). The latency is obtained from the message delay of the selected display. I sometimes use it in the following scenario:

I have a PointCloud2 with 10Hz and a CNN which takes quite some time to process the cloud. In order to visualize them in a synchronized way in Rviz, I have a node which subscribes to both and republishes both (on a different topic) as soon the detection arrived. This works well. However, things using ros::Time() like RobotDisplay or Grid, are not at the right position in Rviz w.r.t. the point cloud due to its large latency:

SyncApprox helps to delay the TFs with stamp ros::Time() similar to the messages of the selected display:


https://user-images.githubusercontent.com/4062443/151217442-be3269f2-e20c-4f5d-b53f-60cb7250bd1f.mp4


SyncExect does not help here, because the TFs are published in a much higher frequency than the message and they are usually not in sync with the selected message.

rhaschke:
I'm not sure, I understood your scenario. Let me try to summarize with my own words:
- There is a `FAST PC2` publisher (10Hz)
- There is a CNN processor, operating on `FAST PC2`, and republishing it as `SLOW PC2` together with TF frames + markers of recognized objects. I guess, this republisher (re)uses the original timestamp of the received input (`FAST  PC2`) message?
However, because processing is slow this timestamp is - say a second - in the "past" w.r.t. to `ros::Time::now()`.
- rviz publishes both `FAST PC2` and `SLOW PC2`? And it _syncs_ against `SLOW PC2`?

rhaschke:
@AndreasR30, did you have time to test this PR? Please also comment on https://github.com/ros-visualization/rviz/pull/1698#issuecomment-1022541426. Thanks.

AndreasR30:
> There is a FAST PC2 publisher (10Hz)

Yes

> There is a CNN processor, operating on FAST PC2...

Yes

> ... and republishing it as SLOW PC2 together with TF frames + markers of recognized objects.

No, the synchronizer is a separate node, which subscribes to original PC2 from the sensor and to Markers of CNN node. No TFs are republished! It republishes both, Markers and PC2 on a different topic when both are available (e.g. prefixed with `/synced/*`). Now the Markers and the PC2 do match each other in the visualization, when I use those `/synced/*` topics in Rviz. With original topics, the Markers were always _behind_ the PC2 because of CNN's latency of ~200ms.

Now the only problem left, is when the `Fixed Frame` is set to `odom`: The PC2 and the Marker from synced topic lag behind the robot because they are world fixed and they are quite old. One can see this, when the RobotModel is no more in the center of the lidar rings. Setting to `ApproxSync` helps here because it uses for the robot model no more the most recent stamp (by means of `ros::Time()`) but a stamp, which lags behind `ros::Time::now()` by  `current_delta_`, which is obtained from the PC2 latency (in our case ~200ms) of the corresponding `PC2Display`.

> I guess, this republisher (re)uses the original timestamp of the received input (FAST PC2) message?
However, because processing is slow this timestamp is - say a second - in the "past" w.r.t. to ros::Time::now().

Yes, but republisher is a separate node, the sync node.

> Is there some documentation on the syncing behavior that I missed?

No I found this feature to be useful by coincidence.

> It would be great, if you could provide a rosbag of your example (as well as the corresponding rviz config).

I will try to make one or a synthetic example in Python. But it could could take some days, because I'm busy.

AndreasR30:
Here is an synthetic example showing when SyncApprox can be useful. It simulates a PC2 with latency of 200ms. Furthermore it simulates the RobotModelDisplay with a Marker. With synchronization off, the robot won't be in the center of the rings. With synchronization approx, the robot will be in the center of the lidar rings.

**Caution: There is a bug! For sync approx the `fixed_frame_id_` is needed, which is only set in `setFixedFrame()` callback. So you have to change Fixed Frame to something else and go back to `map` in order to make SyncApprox work!**

[sync_approx_test.zip](https://github.com/ros-visualization/rviz/files/7967122/sync_approx_test.zip)

AndreasR30:
Your latest commits seem to work good with the Python script from my previous comment. Aproximate Synchronization works well and the bug mentioned before seems to be eliminated.

However, with the Python script from [this comment](https://github.com/ros-visualization/rviz/pull/1698#issuecomment-1025251369) I still see the race conditions (even after few seconds) with Synchronization set to `Frame`. I cherry-picked your commits on the noetic-devel branch as I only have noetic available now. But I don't think that this causes any difference.

rhaschke:
Thanks for providing the script (https://github.com/ros-visualization/rviz/pull/1698#issuecomment-1025251369). I only found time to shortly look into this today. I noticed the following problems:
1. Your script doesn't work with the released version of rviz, which expects perfectly matching timestamps between PC and TF.
  My proposed version relaxes this constraint and makes your script work in `SyncApprox` mode, which is good.
2. Your script doesn't really need `SyncApprox`: Syncing with `Exact` mode perfectly displays everything relative to the (past) timestamp of the PC. 

Thus, I still don't see a need for the `SyncApprox` mode.

rhaschke:
> There is a bug! For sync approx the `fixed_frame_id_` is needed, which is only set in setFixedFrame() callback.
  So you have to change Fixed Frame to something else and go back to map in order to make SyncApprox work!

I cannot confirm this statement. The `fixed_frame_id_` is correctly configured at startup.
However, I noticed that rviz doesn't handle latched marker messages correctly (not displaying any marker). Is that the actual problem you observed?

rhaschke:
> There is a bug! For sync approx the `fixed_frame_id_` is needed, which is only set in setFixedFrame() callback.

You are right: While `fixed_frame_id_` is looked up during initialization, it actually yields an invalid value of zero at that time because the TF buffer didn't process any messages yet. Fixing this issue (or using your workaround) actually resolves the error messages I observed so far in `SyncApprox` mode. My new code doesn't use this variable anymore and thus doesn't exhibit the issue anymore.

AndreasR30:
For me the PR looks good. The earlier synthetic example of this PR works (unfortunately I can't link it). However, it takes up to 10 seconds until it works properly. As long there is a TF with a larger delay incoming compared to `sync_delta_` the glitches are still there and look even worse than with synchronization off. Not sure whats the reason for it. So this mode is not really the same as using just one thread for TF buffer as proposed in #1692. But I think it is ok.

PR #1704, opened 2022-01-30
Maintain original mesh materials for RobotLinks

AndreasR30:
Currently, rviz modifies a custom material setting loaded together with a mesh model, e.g. within [`updateAlpha()`](https://github.com/ros-visualization/rviz/blob/5531206e77d8092dbdb6a50a5b1a3c23fe5e181e/src/rviz/robot/robot_link.cpp#L390-L413).
This destroys, for example, transparency within those original materials.

With this pull request, the _original_ material is stored additionally to the _active_ one, such that the original one can be restored if needed. This allows to correctly render e.g. glass:

|  old                       |  new                     |
|-------------------------|------------------------|
| Screenshot from 2022-01-30 17-26-22 | Screenshot from 2022-01-30 17-15-24 |

### DIFF
--- a/src/rviz/robot/robot_link.cpp
+++ b/src/rviz/robot/robot_link.cpp
@@ -164,6 +164,7 @@ RobotLink::RobotLink(Robot* robot,
   , trail_(nullptr)
   , axes_(nullptr)
   , material_alpha_(1.0)
+  , use_original_material(true)
   , robot_alpha_(1.0)
   , only_render_depth_(false)
   , is_selectable_(true)
@@ -404,11 +405,13 @@ void RobotLink::updateAlpha()
       {
         material->setSceneBlending(Ogre::SBT_TRANSPARENT_ALPHA);
         material->setDepthWriteEnabled(false);
+        use_original_material = false;
       }
       else
       {
         material->setSceneBlending(Ogre::SBT_REPLACE);
         material->setDepthWriteEnabled(true);
+        use_original_material = true;
       }
     }
   }
@@ -668,6 +671,7 @@ void RobotLink::createEntityForGeometryElement(const urdf::LinkConstSharedPtr& l
         sub->setMaterial(mat);
       }
       materials_[sub] = sub->getMaterial();
+      original_materials_[sub] = sub->getMaterial()->clone(sub->getMaterial()->getName() + "_original");
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/rviz/robot/robot_link.cpp
```diff
@@ -668,6 +671,7 @@ void RobotLink::createEntityForGeometryElement(const urdf::LinkConstSharedPtr& l
         sub->setMaterial(mat);
       }
       materials_[sub] = sub->getMaterial();
+      original_materials_[sub] = sub->getMaterial()->clone(sub->getMaterial()->getName() + "_original");
```
rhaschke, line 674: I don't yet get the point of this source line: Isn't that simply cloning the existing material without changes? How can this result in different behavior?

rhaschke (commented):
The resulting screenshots are convincing. But I don't yet understand what's happening.
Also, I noticed that the car is partially translucent, uncovering the underlying grid...

src/rviz/robot/robot_link.cpp
AndreasR30: The materials of the line above (Line 673) are changed afterwards in `updateAlpha()` method. See line [405](https://github.com/ros-visualization/rviz/pull/1704/commits/e4dbcdb2eccd3b1669a710e3cdfe87b588dc95d3#diff-34b5917b7e73ed148ec80e571247fbdeca5d564aa6e83d7a24becfbe6544e8dbL405) to 411. Its not that easy to _undo_ these changes when alpha is set back to 1. because then one as to store what scene blending mode and depth writing mode were before for a specific material.

src/rviz/robot/robot_link.cpp
```diff
@@ -404,11 +405,13 @@ void RobotLink::updateAlpha()
       {
         material->setSceneBlending(Ogre::SBT_TRANSPARENT_ALPHA);
         material->setDepthWriteEnabled(false);
+        use_original_material = false;
       }
       else
       {
         material->setSceneBlending(Ogre::SBT_REPLACE);
```
AndreasR30, line 412: These 2 lines are responsible to make every material opaque when alpha is 1. After the commit they could be actually removed, because this material is not visible in that mode because it is replaced by the original material in Line 930.

AndreasR30:
> Also, I noticed that the car is partially translucent, uncovering the underlying grid...

I do not understand what you mean here. Maybe it is a bit confusing that the wheels of the car are below zero level, while the vertices of the roof have a positive z value. So the grid _cuts_ the car into two halves.

rhaschke:
> So the grid _cuts_ the car into two halves.

Ah. That explains it. Thanks.

rhaschke:
Could you send me a simple urdf example illustrating the problem? Thanks.

AndreasR30:
Here you go. Please don't forget to source package. Otherwise Rviz/Ogre won't find included materials. Execute *.launch file:
`roslaunch robot_model_glass_test robot.launch`

[robot_model_glass_test.zip](https://github.com/ros-visualization/rviz/files/8013697/robot_model_glass_test.zip)

rhaschke (approved):
This looks good in general. Thanks for the contribution.
I have added a few performance improvements, committed as a PR to your fork: https://github.com/UniBwTAS/rviz/pull/2
Please test it on your side and merge the PR here.

Issue #1701, opened 2022-01-27
Disable Mesa hack for 21.x ?


simonschmeisser:
In #1508 you, @rhaschke , stated that in mesa 21.0.3 the bug requiring the old GLSL version has been fixed. Yet here https://github.com/rhaschke/rviz/blob/62ba5cd26344dceeda1b66ada5dd0740f218cbf3/src/rviz/ogre_helpers/render_system.cpp#L193 it still checks for Mesa 2x.y (including Mesa 21.y)

Should that check be limited?

I have read (but not yet confirmed) that Octomap rendering is significantly faster with the newer shaders (but I have seen slowdowns compared to older rviz versions)

rhaschke:
Yes, I think this workaround can be limited to Mesa < 21.0.3. Could you test this and file a PR?

PR #1598, opened 2021-03-19
Enable Mesa workaround also on Mesa 21

peci1:
### Description

This enables the Mesa workaround for Mesa 21.x, too (and all future Mesa 2?.x versions).

My system reports:

```
GL_RENDERER   = AMD RENOIR (DRM 3.40.0, 5.10.24-051024-generic, LLVM 11.0.1)
GL_VERSION    = 4.6 (Compatibility Profile) Mesa 21.0.0 - kisak-mesa PPA
GL_VENDOR     = AMD
```

The fix merged in #1588 is limited to Mesa 20.x only, but the problems it works around are present in 21.x, too.

### Checklist

- [ ] If you are addressing rendering issues, please provide:
  - [ ] Images of both, broken and fixed renderings.
  - [ ] Source code to reproduce the issue, e.g. a `YAML` or `rosbag` file with a `MarkerArray` msg.
- [ ] If you are changing GUI, please include screenshots showing how things looked *before* and *after*.
- [ ] Choose the proper target branch: latest release branch, for non-ABI-breaking changes, *future* release branch otherwise.
      Due to the lack of active maintainers, we cannot provide support for older release branches anymore.
- [ ] Did you change how RViz works? Added new functionality? Do not forget to update the tutorials and/or documentation on the [ROS wiki](http://wiki.ros.org/rviz)
- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-visualization/rviz/pulls) to support the maintainers of RViz. Refer to the [RViz Wiki](https://github.com/ros-visualization/rviz/wiki/Maintainer-Guide#reviewing-pull-requests) for reviewing guidelines.

PR #1714, opened 2022-02-12
Drop Mesa workaround

rhaschke:
Addressing #1701. Needs to be tested on various systems!

### DIFF
--- a/src/rviz/ogre_helpers/render_system.cpp
+++ b/src/rviz/ogre_helpers/render_system.cpp
@@ -171,7 +171,6 @@ void RenderSystem::loadOgrePlugins()
 
 void RenderSystem::detectGlVersion()
 {
-  bool mesa_workaround = false;
   if (force_gl_version_)
   {
     gl_version_ = force_gl_version_;
@@ -184,10 +183,6 @@ void RenderSystem::detectGlVersion()
     int major = caps->getDriverVersion().major;
     int minor = caps->getDriverVersion().minor;
     gl_version_ = major * 100 + minor * 10;
-
-    std::string gl_version_string = (const char*)glGetString(GL_VERSION);
-    // The "Mesa 2" string is intended to match "Mesa 20.", "Mesa 21." and so on
-    mesa_workaround = gl_version_string.find("Mesa 2") != std::string::npos && gl_version_ >= 320;
   }
 
   switch (gl_version_)
@@ -217,15 +212,6 @@ void RenderSystem::detectGlVersion()
       glsl_version_ = 0;
     }
     break;
-  }
-  if (mesa_workaround)
-  { // https://github.com/ros-visualization/rviz/issues/1508
-    ROS_INFO("OpenGl version: %.1f (GLSL %.1f) limited to GLSL 1.4 on Mesa system.",
-             (float)gl_version_ / 100.0, (float)glsl_version_ / 100.0);
-
-    gl_version_ = 310;
-    glsl_version_ = 140;
-    return;
   }
   ROS_INFO("OpenGl version: %.1f (GLSL %.1f).", (float)gl_version_ / 100.0, (float)glsl_version_ / 100.0);
 }


src/rviz/ogre_helpers/render_system.cpp
```diff
@@ -184,10 +183,6 @@ void RenderSystem::detectGlVersion()
     int major = caps->getDriverVersion().major;
     int minor = caps->getDriverVersion().minor;
     gl_version_ = major * 100 + minor * 10;
-
-    std::string gl_version_string = (const char*)glGetString(GL_VERSION);
-    // The "Mesa 2" string is intended to match "Mesa 20.", "Mesa 21." and so on
-    mesa_workaround = gl_version_string.find("Mesa 2") != std::string::npos && gl_version_ >= 320;
```
simonschmeisser, line 190: My suggestion would have been to just change this to "Mesa 20"

simonschmeisser:
I was just testing around with the PC2 display on AMD Renoir  and noticed that most of the more fancy Styles seem to be unsupported with GLSL 1.5. Changing the Size has no effect for anything besides Points and it's always displayed as super tiny dots. I was not able to trigger any unusual behavior with the Points style

We have serious performance regressions with high resolution octomaps when limiting to the older shader versions though, that's why I opened the discussion.


```
[ INFO] [1644874202.850198482]: rviz version 1.14.14
[ INFO] [1644874202.850307278]: compiled against Qt version 5.12.8
[ INFO] [1644874202.850323799]: compiled against OGRE version 1.9.0 (Ghadamon)
[ INFO] [1644874202.865892892]: Forcing OpenGl version 0.
[ INFO] [1644874203.187769729]: Stereo is NOT SUPPORTED
[ INFO] [1644874203.187885107]: OpenGL device: AMD RENOIR (DRM 3.41.0, 5.13.0-28-generic, LLVM 12.0.0)
[ INFO] [1644874203.187928259]: OpenGl version: 4,6 (GLSL 4,6).
```

src/rviz/ogre_helpers/render_system.cpp
simonschmeisser: Was the Mesa 20 part of some HWE or part of the stock 18.04, ie are there likely going to be people with non HWE systems running into this again?

peci1:
Interesting, I also tested that on Renoir and it did not work for me with Mesa 21, that's why I requested the workaround to be extended to all 2x versions.

https://github.com/ros-visualization/rviz/pull/1598

simonschmeisser:
Your test was with Mesa 21.0.0 while the fix appeared in 21.0.3 (I'm not sure it has been pinpointed exactly but maybe I misunderstood something ...)

peci1:
@simonschmeisser Good point, haven't noticed that... Could you put together a minimum working example? It's been some time since I was solving this, so I can't remember what were the steps to receive a failure...

src/rviz/ogre_helpers/render_system.cpp
peci1:
```
$ LANG=C apt-cache policy mesa-va-drivers 
mesa-va-drivers:
  Installed: 21.3.6~kisak1~b
  Candidate: 21.3.6~kisak1~b
  Version table:
 *** 21.3.6~kisak1~b 500
        500 http://ppa.launchpad.net/kisak/kisak-mesa/ubuntu bionic/main amd64 Packages
        100 /var/lib/dpkg/status
     20.0.8-0ubuntu1~18.04.1 500
        500 http://cz.archive.ubuntu.com/ubuntu bionic-updates/universe amd64 Packages
     19.2.8-0ubuntu0~18.04.2 500
        500 http://security.ubuntu.com/ubuntu bionic-security/universe amd64 Packages
     18.0.0~rc5-1ubuntu1 500
        500 http://cz.archive.ubuntu.com/ubuntu bionic/universe amd64 Packages
```

peci1: That's on Bionic HWE with kisak-mesa PPA. So it seems 20.x is available via bionic-updates for everyone.

rhaschke:
I'm not sure I fully understood your comments. Thus, let me try to summarize. @peci1 and @simonschmeisser, please correct!
- With Mesa < 21.0.3 we had a bug with PC2 display showing huge point sizes: 
  #1508
  A corresponding test is available in [`src/test/send_point_cloud_2.py`](https://github.com/ros-visualization/rviz/blob/noetic-devel/src/test/send_point_cloud_2.py)
- Mesa 21.0.3 was fixing this and it is released, both for [Bionic (via bionic-updates) and Focal (natively)](https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=libgl1-mesa-dev&searchon=names)
  Hence, it should be safe to drop the Mesa workaround.
- @simonschmeisser observed some new? issue on AMD Renoir, with PC2 display showing all styles other than points as tiny dots only, i.e. ignoring the size attribute
  Is this reproducible on other graphic cards or is it specific to AMD? Please file a new issue for this.

peci1:
> * Mesa 21.0.3 was fixing this and it is released, both for Bionic (via bionic-updates)

Bionic has only Mesa 20. So this fix hasn't been released there. I (and probably others) are using a customized Mesa PPA to get a newer Mesa on Bionic, as that's the only way to get the new Ryzen GPUs working (together with mainline kernels).

peci1:
I tested on Bionic+Ryzen GPU+Mesa 21 with and without the Mesa workaround:

With workaround: everything works (I haven't tested performance, though).

Without workaround: Points are displayed with correct size, but there is this (new?) issue with everything except points displaying as a single white dot:
rviz_screenshot_2022_03_07-12_35_45

I'll open an issue for the "newly discovered" bug, but I think this PR needs to be hold off until that issue is solved.

rhaschke:
> > * Mesa 21.0.3 was fixing this and it is released, both for Bionic (via bionic-updates)
> 
> Bionic has only Mesa 20. So this fix hasn't been released there.

You are right. I was writing my comment in a rush obviously...

peci1:
> note that you can manually enforce an older OpenGL standard: `rviz --opengl 310`.

Yes, but this combines very badly with already existing launch files starting rviz. Maybe if it were an environment variable, it would be easier to set.

rhaschke:
What about `alias rviz='rviz --opengl 310'` in your `.bashrc`?

peci1:
`<node name="rviz" pkg="rviz" type="rviz" args="-d path/to/config.rviz />` inside an (externally provided) launch file is the hard thing... the alias won't help there

rhaschke:
Closing for now, as our GLSL 1.50 scripts are not Mesa-compatible yet.

PR #1725, opened 2022-03-17
Fix GLSL 1.50 for Mesa

rhaschke:
As pointed out in #1721, Mesa doesn't correctly render PointCloud geoms other than points with GLSL 150, while NVIDIA renders them fine.
The problem is that Mesa uses are more strict GLSL compiler than NVIDIA, thus rejecting the existing scripts. Essentially, `Mesa` doesn't support the `compatibility` profile. Further, it's not allowed to redefine `gl_PerVertex` or use the `gl_` prefix for user variables (https://github.com/ros-visualization/rviz/issues/1721#issuecomment-1062205008).

I was able to fix most syntax issues, such that `glslangValidator` doesn't complain anymore.
However, the `compatibility` issue remains. Using the `compatibility` profile, color values can be passed from CPU to GPU as packed values (`uint`):
https://github.com/ros-visualization/rviz/blob/ce068cbd932d4c3f971df2ed0f9d60b9f5aa5d81/src/rviz/ogre_helpers/point_cloud.cpp#L532
and be read in GLSL via `gl_Color` - as a `vec4`:
https://github.com/ros-visualization/rviz/blob/697ed3ff0496b79b1bb4270aabf457ee2f2ae736/ogre_media/materials/glsl150/pass_pos_color.vert#L13
However, when dropping `compatibility` profile, we need to unpack the color ourselves:
```
in vec3 position;
in uint packed_color;
```
But, I have no idea, how to accomplish that... I'm stuck here.

PR #1726, opened 2022-03-17
GLSL 330 shaders

### LINKED ISSUE #1721: Mesa 21 without Mesa workaround incorrectly renders PointCloud2

### Your environment
* OS Version: Ubuntu 18.04
* ROS Distro: Melodic
    ```
    [ INFO] [1646652899.525315495] [ros.rviz]: rviz version 1.13.24
    [ INFO] [1646652899.525380866] [ros.rviz]: compiled against Qt version 5.9.5
    [ INFO] [1646652899.525397976] [ros.rviz]: compiled against OGRE version 1.9.0 (Ghadamon)
    [ INFO] [1646652899.533460029] [ros.rviz]: Forcing OpenGl version 0.
    [ INFO] [1646652899.757093342] [ros.rviz]: Stereo is NOT SUPPORTED
    [ INFO] [1646652899.757155850] [ros.rviz]: OpenGL device: AMD RENOIR (DRM 3.44.0, 5.16.10-051610-generic, LLVM 13.0.0)
    [ INFO] [1646652899.757174358] [ros.rviz]: OpenGl version: 4.6 (GLSL 4.6).
    ```
* If source build, which git commit? 0462dd27
* System locale, i.e. the output of `echo "$LANG $LC_NUMERIC"`: `C C`

On this system, I'm using a more recent Mesa from kisak-mesa PPA (mesa 21.3.7).

Currently, there is the mesa version workaround that forces using GLSL 1.4 on systems with Mesa >= 20. In order to resolve #1701, I tried disabling the Mesa workaround, since the original issue (large points) should have been fixed since Mesa 21.0.3. The big points issue is fixed, but there is another problem: every style of PointCloud2 display except "Points" is displayed just as a white dot regardless of the settings:

rviz_screenshot_2022_03_07-12_35_45

### How to test

- run `rviz` in one terminal, configure a `PointCloud2` display for topic `points2`, set fixed frame to `map`.
- in a second terminal, run this script: https://github.com/ros-visualization/rviz/blob/noetic-devel/src/test/send_point_cloud_2.py
- play around with the pointcloud display style

simonschmeisser:
This comes with a lot of hubris and a lack of fundamental knowledge so please be patient and don't hold back anything that comes up when you see this lame attempt ...

Also this is a WIP of course, a proper solution would either try to add 330 besides 150 in proper folders or remove 150 altogether.

I'm trying to fix #1721 and just noticed your parallel attempt @rhaschke 

This was verified with GPUpad syntactically but I could not get neither gpupad nor rviz to render anything (or not crash in the case of rviz ...)

@paroj may I ask you for help here or in #1725


simonschmeisser:
I was taking some information from https://ogrecave.github.io/ogre/api/latest/_high-level-_programs.html

ogre_media/materials/glsl150/box.geom
```diff
@@ -19,13 +19,13 @@ uniform mat4 worldviewproj_matrix;
 uniform vec4 size;
 uniform vec4 auto_size;
 
-in gl_PerVertex {
-	vec4 gl_Position;
-	vec4 gl_FrontColor;
-} gl_in[];
+in PerVertex {
+	vec4 frontColor;
+} gs_in[];
```
paroj, line 24: I think these are the only strictly necessary changes in this PR

rhaschke:
@simonschmeisser, GPUpad seems to be an interesting tool to develop the GLSL scripts.
However, I couldn't find an option to save a whole session (like the examples). Did you?

simonschmeisser:
Simply File > Save as   It mostly references external files though. So you first need to save all of the scripts and then the project

simonschmeisser:
image
Using 1.23.1 Appimage

rhaschke:
Hm. I'm able to save a newly created session. But I cannot save (as) an example session?

rhaschke:
Closing in favor of #1725, which seems to work with Mesa (and a hack in OGRE).

PR #1730, opened 2022-03-18
Drop OGRE/ from #include directives

rhaschke:
Implements https://github.com/ros-visualization/rviz/pull/1630#issuecomment-1072626521. Untested.

### DIFF
--- a/src/rviz/ogre_helpers/movable_text.cpp
+++ b/src/rviz/ogre_helpers/movable_text.cpp
@@ -41,16 +41,16 @@
 
 #include "movable_text.h"
 
-#include <OGRE/OgreVector3.h>
-#include <OGRE/OgreQuaternion.h>
-#include <OGRE/OgreRoot.h>
-#include <OGRE/OgreCamera.h>
-#include <OGRE/OgreSceneNode.h>
-#include <OGRE/OgreMaterialManager.h>
-#include <OGRE/OgreHardwareBufferManager.h>
-#include <OGRE/Overlay/OgreFontManager.h>
-#include <OGRE/Overlay/OgreFont.h>
-#include <OGRE/OgreUTFString.h>
+#include <OgreVector3.h>
+#include <OgreQuaternion.h>
+#include <OgreRoot.h>
+#include <OgreCamera.h>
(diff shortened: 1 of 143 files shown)


src/rviz/ogre_helpers/movable_text.cpp
```diff
@@ -41,16 +41,16 @@
...
+#include <OgreVector3.h>
+#include <OgreQuaternion.h>
+#include <OgreRoot.h>
+#include <OgreCamera.h>
+#include <OgreSceneNode.h>
+#include <OgreMaterialManager.h>
+#include <OgreHardwareBufferManager.h>
+#include <Overlay/OgreFontManager.h>
+#include <Overlay/OgreFont.h>
+#include <OgreUTFString.h>
```
simonschmeisser, line 53: this makes https://github.com/ros-visualization/rviz/commit/a4919f52fb37da074fcea1648e257691037a3feb superfluous as proposed originally in #1645  :wink:

v4hn:
https://github.com/OGRECave/ogre/pull/2411

src/rviz/ogre_helpers/movable_text.cpp
```diff
@@ -41,16 +41,16 @@
...
-#include <OGRE/Overlay/OgreFont.h>
-#include <OGRE/OgreUTFString.h>
+#include <OgreVector3.h>
+#include <OgreQuaternion.h>
+#include <OgreRoot.h>
+#include <OgreCamera.h>
+#include <OgreSceneNode.h>
+#include <OgreMaterialManager.h>
+#include <OgreHardwareBufferManager.h>
+#include <Overlay/OgreFontManager.h>
```
paroj, line 51: CMake will add Overlay to the include path, so you should not use the `Overlay/` prefix.

Also, there is native movable text in 1.12.10+:
https://www.ogre3d.org/2020/12/13/ogre-1-12-10-released#3dtext

v4hn (approved):
As stressed by upstream in the linked PR, Ogre headers must never be included with a folder prefix.

So this patch makes sense after also removing all the `Overlay/` references.

PR #1732, opened 2022-03-25
RobotLink: Simplify material mode handling

rhaschke:
rviz::RobotLink knows 3 material modes, showing:
- original material of the mesh
- a user-selected color
- an error color (uniform white)

Switching between these modes requires iterating through all subentities of the link, 
which might be rather time-consuming. Also, this was done in each update() cycle!

This commit simplifies the handling, performing the costly update only if really necessary.


rhaschke:
Closing and reopening to trigger CI.

PR #1733, opened 2022-03-27
Switch to GL3+ RenderSystem

simonschmeisser:
This is WIP and based on #1725 

- [ ] Find out why it does not find `resources.cfg`
- [ ] Find out how to decide which RenderSystem to use even though the GL version detection is part of the RenderSystem
- [ ] PointCloud shaders are plain white currently. Debug with RenderDoc


simonschmeisser:
Sorry for bothering you once more @paroj, I tried enabling GL3+ rendering system but fail at early steps.
```
terminate called after throwing an instance of 'Ogre::InvalidStateException'
  what():  InvalidStateException: RenderSystem does not support FixedFunction, but technique of 'Grid0Material' has no Vertex Shader. Use the RTSS or write custom shaders. in SceneManager::_setPass at /home/sis/experimente/debian/ogre-1.12/OgreMain/src/OgreSceneManager.cpp (line 880)
```
I get this quite often and assume it's because I didn't setup the resource locations properly. I looked at how that is done normally and found this in OgreBites:
https://github.com/OGRECave/ogre/blob/4040a7988e94629bc1de66159b7eb57ddfeeb272/Components/Bites/src/OgreApplicationContextBase.cpp#L424

I think we just want to load the resources.cfg installed by the debian package in /usr/share/OGRE-VERSION/resources.cfg so I can either copy paste that function, get the location from CMake/pkgcfg or possibly create a `OgreApplicationContextBase` instance.

For simplicity I just copied the RTShaderSystem part and set the path manually for my version of ogre. Do I need to set the cache path? (Does not seem to make a difference and stays empty)

Do I need to change the material files?

paroj:
you missed the listener: https://github.com/OGRECave/ogre/blob/4040a7988e94629bc1de66159b7eb57ddfeeb272/Components/Bites/src/OgreApplicationContextBase.cpp#L102

simonschmeisser:
Thanks @paroj, I made some little progress. The coordinate grid can now be displayed but other materials and shaders still crash:
```
terminate called after throwing an instance of 'Ogre::InvalidStateException'
  what():  InvalidStateException: RenderSystem does not support FixedFunction, but technique of 'robot link material' has no Vertex Shader. Use the RTSS or write custom shaders. in SceneManager::_setPass at /home/sis/experimente/debian/ogre-1.12/OgreMain/src/OgreSceneManager.cpp (line 880)
```
for our typical mesh display as well as for the picking shader ( https://github.com/ros-visualization/rviz/blob/noetic-devel/ogre_media/materials/scripts/default_pick_and_depth.material )

rhaschke:
@simonschmeisser, what is the status of this PR. Obviously, your changes are not yet supported in default Ogre 1.9:
```
render_system.cpp:274:33: error: ‘RGN_INTERNAL’ is not a member of ‘Ogre’
    274 |                           Ogre::RGN_INTERNAL);
```

simonschmeisser:
```
terminate called after throwing an instance of 'Ogre::InvalidStateException'
  what():  InvalidStateException: RenderSystem does not support FixedFunction, but technique of 'robot link table_base:' has no Vertex Shader. Use the RTSS or write custom shaders. in SceneManager::_setPass at ./OgreMain/src/OgreSceneManager.cpp (line 844)
```
We have some entities where we create a `Material` from code without any material script. I tried following the advice in https://forums.ogre3d.org/viewtopic.php?t=96852 and added `createShaderBasedTechnique` but that does not seem to be sufficient?

simonschmeisser:
I added a try-catch around `ogre_root_->renderOneFrame();` and can now turn on and off plugins without rviz crashing. Interestingly `MarkerArrayDisplay` is working correctly. It seems to use the ogre default material:
https://github.com/ros-visualization/rviz/blob/68ce34b0c9053b3226ab16feaa09bad62d255347/src/rviz/default_plugin/markers/mesh_resource_marker.cpp#L145

Issue #916, opened 2015-09-10
Clickable image


jolting:
I think it would to be nice to be able to click on an image and do something useful with it similar to the way you can select a point with the point tool.

I was thinking something like this:
jolting/rviz@fc6922cc2316aecc7d04ffc9073bdd287fe9596c

wjwwood:
That's an interesting idea. If you have time I'd be happy to look over a pull request to the effect. It might be something worth exposing to interactive markers as well.

miguelriemoliveira:
Hello everyone,

I think I have a working solution for extending the image display with mouse clicking functionality. Would that be interesting? Should I do a pull request?

rhaschke:
Sure.

miguelriemoliveira:
OK, I will try and get back with the results.

miguelriemoliveira:
Hi @rhaschke, I have just created a [PR](https://github.com/ros-visualization/rviz/pull/1737) with the new functionality. 
Hope it can be integrated into rviz.

PR #1737, opened 2022-04-11
Image display with mouse clicking functionality

miguelriemoliveira:
### Description

The idea is that the user can click an image display in rviz and receive, through a published ros topic, the image pixel coordinates where the click occurred. This could be useful in many situations, for example for manually annotating the images.
Here's the [issue](https://github.com/ros-visualization/rviz/issues/916) that is requesting this functionality.

The MouseClick class handles mouse clicking functionalities integrated into the ImageDisplay, which:
   - uses a qt event filter to capture mouse clicks;
   - handles image resizing;
   - checks if the click was inside or outside the image;
   - scales the pixel coordinates to get them w.r.t. the image (not the window) size;
   - handles disabling and enabling the image display;
   - handles inserting of a non valid topic name (stops publishing, but does no crash);
  
Mouse clicks image pixel coordinates are published as [ros geometry_msgs PointStamped](http://docs.ros.org/en/api/geometry_msgs/html/msg/PoseStamped.html). The z component is ignored. 

The topic name where the mouse clicks are published is created after the subscribed image topic as: <image_topic>/mouse_click.

### Checklist

This is not a rendering issue.

  - [x] Source code to reproduce the issue, e.g. a `YAML` or `rosbag` file with a `MarkerArray` msg.

For testing, I have used this [bagfile](https://drive.google.com/file/d/15BZMaKk6r-A9kmC7ENyVdNUyM6UipqpB/view?usp=sharing) which contains images from two separate cameras, **camera_2** and **camera_3**. 
The topics are:

      /camera_2/rgb/image_raw   35 msgs    : sensor_msgs/Image
      /camera_3/rgb/image_raw   35 msgs    : sensor_msgs/Image 

Use rosbag to play the bag:

     rosbag play two_rgb_cameras.bag -l

and then run rviz and create an image display and configure the topic name (or you can use this [rviz configuration for this bag file](https://drive.google.com/file/d/1Xb-pUeYdIQQTb4fs-GoRqMyIiU-z9-gt/view?usp=sharing) ).  
At this point a rostopic list should show the clicked mouse topic created as:

    <image_topic>/mouse_click

which you can listen using, e.g., for **camera_2**:

    rostopic echo /camera_2/rgb/image_raw/mouse_click

Now if you click somewhere on the image in rviz the pixel coordinates will be shown on the rostopic echo terminal.

This image shows my workspace while testing the functionality. Terminal title bars show the commands that are running:

rviz_mouse_click

This does not change the appearance of the gui, but integrated with a ros node that produces new images based on the image clicks, we can have something like [this](https://github.com/lardemua/atom/issues/378#issuecomment-1086651584), where the clicked points are being used to build a polygon annotated on the image which is being subscribed.

I am not entirely sure but from what I read there are no ABI-breaking changes, so the target is the latest release branch.

I will do it once/if the PR is accepted.


miguelriemoliveira:
Hello again,

I noticed that the **Npr__rviz__ubuntu_focal_amd64 — Build finished.** failed. I looked into the details and I think its nothing about the code I commited, but other warnings of different files in rviz. Should I do something or just wait for your review?

rhaschke:
I didn't have time to look into this yet (and won't within the upcoming week). The failure of the ROS build farm is due to warnings in the python wrapper we cannot silent. No need for action.

miguelriemoliveira:
Dear @rhaschke ,

I imagine you are very busy, but I wanted to ask if you plan to integrate this functionality into RVIZ. From our tests it works fine, and I am sure it will be useful to many.

I am asking because we are about to release a ros calibration framework that requires this functionality to run some image annotations.

Right now we have to ask people to download our rviz fork

https://lardemua.github.io/atom_documentation/getting_started/#clone-rviz-fork

which is far from ideal. It would be much better if the users could just use rviz out of the box.

Could I please ask you to take a look at this? Thank you so much for any time you can spare.

Thank you,
Miguel

src/rviz/default_plugin/image_display.cpp
```diff
@@ -241,6 +251,19 @@ void ImageDisplay::processMessage(const sensor_msgs::Image::ConstPtr& msg)
   texture_.addMessage(msg);
 }
 
+void ImageDisplay::setTopic(const QString& topic, const QString& datatype)
+{
+  ImageDisplayBase::setTopic(topic, datatype);
+  mouse_click->setTopic(topic);
+}
+
```
rhaschke, line 259: There is no need to override `ImageDisplayBase::setTopic()` as this will call `updateTopic()` anyway.

src/rviz/default_plugin/image_display.cpp
```diff
@@ -72,6 +72,8 @@ ImageDisplay::ImageDisplay() : ImageDisplayBase(), texture_()
                       this, SLOT(updateNormalizeOptions()));
 
   got_float_image_ = false;
+
+  mouse_click = new MouseClick();
```
rhaschke, line 76: Nitpick: For deletion, you rely on RenderPanel being mouse_click's parent.
However, this is only set in `onInitialize()`!

src/rviz/default_plugin/image_display.h
```diff
@@ -98,6 +101,8 @@ public Q_SLOTS:
   FloatProperty* max_property_;
   IntProperty* median_buffer_size_property_;
   bool got_float_image_;
+
+  MouseClick* mouse_click;
```
rhaschke, line 105: Please follow the naming convention for class members:
```suggestion
  MouseClick* mouse_click_;
```

src/rviz/image/mouse_click.h
```diff
@@ -0,0 +1,61 @@
...
+  void setTopic(const QString& image_topic);
+  void updateTopic(const QString& image_topic);
+
+private:
+  virtual bool eventFilter(QObject* obj, QEvent* event);
+
+  bool has_dimensions_;
+  int img_width_, img_height_, win_width_, win_height_; // To assess if the clicks are inside the image.
+  boost::shared_ptr<ros::NodeHandle> node_handle_;
+  boost::shared_ptr<ros::Publisher> publisher_;
```
rhaschke, line 55: Why do you need to use pointers here?

src/rviz/image/mouse_click.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+
+void MouseClick::onInitialize(QObject* parent)
+{
+  setParent(parent);
+  parent->installEventFilter(this);
+
+  has_dimensions_ = false;
+  is_topic_name_ok_ = false;
+  node_handle_.reset(new ros::NodeHandle());
```
rhaschke, line 25: This should inherit from Display's `update_nh_`. I suggest to pass the NodeHandle as an extra const-ref argument.

src/rviz/image/mouse_click.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+void MouseClick::publish()
+{
+  if (is_topic_name_ok_)
+  {
+    publisher_.reset(
+        new ros::Publisher(node_handle_->advertise<geometry_msgs::PointStamped>(topic_, 1000)));
```
rhaschke, line 33: Why would you need a queue of 1000 here?

src/rviz/image/mouse_click.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+
+void MouseClick::onInitialize(QObject* parent)
+{
+  setParent(parent);
+  parent->installEventFilter(this);
+
+  has_dimensions_ = false;
+  is_topic_name_ok_ = false;
```
rhaschke, line 24: I would expect these initializations in the constructor already.

src/rviz/image/mouse_click.cpp
```diff
@@ -0,0 +1,127 @@
...
+      int pix_y = -1;
+      if (img_aspect > win_aspect) // Window is taller than the image: black bars over and under image.
+      {
+        pix_x = int(float(windowPos.x()) / float(win_width_) * float(img_width_));
+
+        int resized_img_height = int(float(win_width_) / float(img_width_) * float(img_height_));
+        int bias = int((float(win_height_) - float(resized_img_height)) / 2.0);
+        pix_y = (float(windowPos.y()) - bias) / float(resized_img_height) * float(img_height_);
+      }
+      else // Window wider than the image: black bards on the side.
```
rhaschke, line 72:
```suggestion
      else // Window wider than the image: black bars on the side.
```

src/rviz/image/mouse_click.cpp
```diff
@@ -0,0 +1,127 @@
...
+      {
+        pix_y = int(float(windowPos.y()) / float(win_height_) * float(img_height_));
+
+        int resized_img_width = int(float(win_height_) / float(img_height_) * float(img_width_));
+        int bias = int((float(win_width_) - float(resized_img_width)) / 2.0);
+        pix_x = (float(windowPos.x()) - bias) / float(resized_img_width) * float(img_width_);
+      }
+
+      // Publish if clicked point is inside the image.
+      if (pix_x > 0 && pix_x < img_width_ && pix_y > 0 && pix_y < img_height_)
```
rhaschke, line 82: Shouldn't be one boundary (zero) included here?
```suggestion
      if (pix_x >= 0 && pix_x < img_width_ && pix_y >= 0 && pix_y < img_height_)
```

src/rviz/image/mouse_click.cpp
```diff
@@ -0,0 +1,127 @@
...
+      {
+        geometry_msgs::PointStamped point_msgs;
```
rhaschke, line 84:
```suggestion
        geometry_msgs::PointStamped point_msg;
```

rhaschke (changes requested):
In general, this looks good. I have a few remarks though. Please also consider formatting issues raised by CI.
Why do you constrain this functionality to `ImageDisplay` and don't include `CameraDisplay` as well? Implementing this at the level of ImageDisplayBase should cover both, shouldn't it?

miguelriemoliveira:
Hi @rhaschke ,

thank you for taking time to look into this. I will work on it by the end of this week and get back to you.

Regards,
Miguel

src/rviz/default_plugin/image_display.cpp
miguelriemoliveira: I am sorry but my cpp is a bit rusty. Like every sane person, I've moved to python a few years ago : - )

What do you suggest we do here?

src/rviz/default_plugin/image_display.cpp
miguelriemoliveira: OK, I will remove this override.

src/rviz/image/mouse_click.cpp
miguelriemoliveira: Ok, makes sense.

src/rviz/image/mouse_click.cpp
miguelriemoliveira: That's just a copy paste from other stuff. I will set it to the default.

src/rviz/image/mouse_click.h
miguelriemoliveira: Its been some years since I've used cpp, but at the time I was using boost pointers over regular pointers whenever possible because I thought they were safer ... not sure if that makes sense tough

miguelriemoliveira (commented):
Hi @rhaschke ,

I went through your suggested changes. They all make sense to me. 

I do not have a lot of experience with this pull request / review mechanisms so I am a bit lost:
Should I implement something from my side for you to review, or is accepting the changes you proposed enough?

About the question of why I did not add this mouse click functionality at the level of the ImageDisplayBase it was just because I did not think of that. 

My cpp is rusty, and this rviz code is not easy to grasp.

If you want I can do a new PR with this change (and the others).

miguelriemoliveira:
>  Please also consider formatting issues raised by CI.

You mean these?

image

I go in the links but cannot find any relevant information. How can I check these formatting issues.

Thanks for the help.

src/rviz/default_plugin/image_display.cpp
rhaschke: Pass a parent QObject* to MouseClick's constructor a forward this argument to the base class constructor (see below).

src/rviz/image/mouse_click.cpp
```diff
@@ -0,0 +1,127 @@
+#include "rviz/image/mouse_click.h"
+
+#include <ros/names.h>
+
+
+namespace rviz
+{
+
+MouseClick::MouseClick() : QObject()
```
rhaschke, line 9: Provide a proper parent in the constructor as suggested in https://github.com/ros-visualization/rviz/pull/1737#discussion_r934052113:
```suggestion
MouseClick::MouseClick(QObject* parent) : QObject(parent)
```

src/rviz/default_plugin/image_display.h
rhaschke: This variable rename requires corresponding changes in other places as well!

src/rviz/image/mouse_click.h
rhaschke: My point was to not use pointers at all for these, but regular object instances.

rhaschke (changes requested):
> I do not have a lot of experience with pull request / review mechanisms so I am a bit lost:
> Should I implement something from my side for you to review, or is accepting the changes you proposed enough?

Accepting and applying my suggestions is the first step. This will add more commits to your PR branch. However, some of those suggested changes (e.g. variable renames) require actions in other locations as well. Some review comments require genuine actions of you.
Its your task to work on this and improve your PR branch until it becomes acceptable for merging.

> If you want I can do a new PR with the requested changes.

You don't do a new PR, but you just push additional commits to your existing PR branch.
https://stackoverflow.com/questions/7947322/preferred-github-workflow-for-updating-a-pull-request-after-code-review

> Please also consider formatting issues raised by CI.

I was referring to the CI job called `Format / pre-commit`: [Here is the log](https://github.com/ros-visualization/rviz/runs/7601920836).

miguelriemoliveira:
Hi @rhaschke ,

thanks for the tips, and sorry for the delayed response. You caught me on vacations!

I will try to work on this tonight and get back to you asap.

src/rviz/image/mouse_click.h
miguelriemoliveira: Hi @rhaschke ,

So I just set the **node_handle_** as a regular object instance.

However, for the **publisher_** I think I need it to be a pointer so I can create a new publisher on a different topic in these methods:

https://github.com/miguelriemoliveira/rviz/blob/61974117adaf5f632800ccc906fe84d3204d5a93/src/rviz/image/mouse_click.cpp#L28-L40

If the **publisher_** is a regular object I don't know how to do this ...

src/rviz/default_plugin/image_display.h
miguelriemoliveira: Done. I have corrected and it compiles.

miguelriemoliveira:
> In general, this looks good. I have a few remarks though. Please also consider formatting issues raised by CI.
> Why do you constrain this functionality to `ImageDisplay` and don't include `CameraDisplay` as well? Implementing this at the level of ImageDisplayBase should cover both, shouldn't it?

Hi @rhaschke ,

So I tried to move the mouse_click functionality to the ImageDisplayBase but could not get pass this point:

https://github.com/miguelriemoliveira/rviz/blob/61974117adaf5f632800ccc906fe84d3204d5a93/src/rviz/default_plugin/image_display.cpp#L121-L136

because my mouse_click.onInitialize() method requires a QObject as argument, which I get from the 

`render_panel_ = new RenderPanel();`

But this does not exist in the ImageDisplayBase, only on the ImageDisplay ... 

So I don't know how to move the mouse_click into the ImageDisplayBase ... not sure its possible ... any ideas?

src/rviz/image/mouse_click.cpp
miguelriemoliveira: Done.

src/rviz/image/mouse_click.cpp
miguelriemoliveira: Set queue to 1.

src/rviz/image/mouse_click.cpp
miguelriemoliveira: Got it. I am passing the update_nh_ to the MouseClick constructor.

src/rviz/image/mouse_click.cpp
miguelriemoliveira: I have done this but it had an implication which I am not sure is ok.

Because I only get the parent **render_panel_**  during the ImageDisplay::onInitialize() I have to call the cnstructor of the MouseClick from there, instead of the ImageDisplay::ImageDisplay() ...

miguelriemoliveira:
Hi @rhaschke ,

I was working on this tonight and I think I have a version ready for you to review.

Above I gave point to point comments to the things you suggested. 

I could not move the functionality to the ImageDisplayBase, as I explain above. 

Please let me know how I can further improve the code.

Thanks for your help,
Miguel

miguelriemoliveira:
Hi @rhaschke ,

I just saw some changes on this from your side. Are you coming back to this issue?

rhaschke:
> I just saw some changes on this from your side. Are you coming back to this issue?

Yes, I want to merge this PR now. Sorry for the looong delay. This slipped out of my attention.

miguelriemoliveira:
> > I just saw some changes on this from your side. Are you coming back to this issue?
> 
> Yes, I want to merge this PR now. Sorry for the looong delay. This slipped out of my attention.

Great news. I thought this was lost forever. I was going around telling my friends I was "almost" an RViz contributor :).

Thanks.

peci1:
@miguelriemoliveira I see a problem with this PR: When I click Add->By topic, all image streams now also show a PointStamped display in addition to the Camera/DepthCloud/Image displays. This is almost certainly wrong.

Snímek obrazovky pořízený 2024-05-09 11-41-27


Also, the frame_id of the published points could be copied from the image stream, couldn't it?

Last, the current implementation only allows a single topic for each source image stream. But what if I had e.g. two displays with displaying the same topic but wanted to distinguish which one has been clicked? I understand this is a corner case I don't really face right now, but it might be something to consider.

peci1:
Also, the timestamp of the click is `ros::Time::now()`. In some cases, it would be more intuitive to get the timestamp of the currently displayed image.

(Hopefully) last comment: rqt_image_view [uses geometry_msgs/Point](https://github.com/ros-visualization/rqt_image_view/blob/3803409bd001ca0583f8cc2a80941f53488afc2c/src/rqt_image_view/image_view.cpp#L413) instead of PointStamped. It's a pity these two are not aligned, but I don't know what way to fix this dichotomy would be better. On one hand, this is a new functionality so it should be easier to bend it, on the other hand, having the header might be useful sometimes.

miguelriemoliveira:
Hi @peci1 ,

I filled this PR about an year ago.
At that time it was quite hard to pick up on the cpp code from rviz, as I usually use python.

At the moment I do not have enough time to go back to this. Perhaps in a couple of weeks it will be possible.

Also, this was already merged, right? So should we work on a different PR?

PR #1739, opened 2022-04-21
Don't disable display if associated widget tab changes

### LINKED ISSUE #1738: Associated widget which is docked creating a tabbed panels gets disabled if tab is not visible
Currently if I dock my panel on top of another panel creating tabbed panels and then you switched tab it disables the display. Is there a way to disable this functionality?

rhaschke:
Fixes #1738

If an associated widget is tabbed and deselected, its visibility changes to false, which triggers disabling the display.
However, in this case, the widget is still there and thus the display shouldn't be disabled.

### DIFF
--- a/src/rviz/display.cpp
+++ b/src/rviz/display.cpp
@@ -386,9 +386,6 @@ void Display::setAssociatedWidget(QWidget* widget)
 
 void Display::associatedPanelVisibilityChange(bool visible)
 {
-  if (associated_widget_panel_) // consider the actual visibility of the panel
-    visible = associated_widget_panel_->isVisible();
-
   associated_widget_visible_ = visible;
   // If something external makes the panel visible/invisible, make sure to enable/disable the display
   setEnabled(visible);
--- a/src/rviz/display.cpp
+++ b/src/rviz/display.cpp
@@ -386,6 +386,9 @@ void Display::setAssociatedWidget(QWidget* widget)
 
 void Display::associatedPanelVisibilityChange(bool visible)
 {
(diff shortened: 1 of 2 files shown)


rhaschke:
This PR requires more work:
For a camera display, it makes sense to disable the display if the widget becomes invisible, because its tab is deselected.
Instead, to solve your issue #1738, @Levi-Armstrong, you could simply disconnect the `visibilityChanged` signal for your display as the `closed` signal will still disable the display if the panel is closed:
https://github.com/ros-visualization/rviz/blob/1f622b8c95b8e188841b5505db2f97394d3e9c6c/src/rviz/display.cpp#L370-L372

However, by doing so, the widget doesn't become visible anymore if re-enabled.

Levi-Armstrong:
Since it seems like there are two valid use case; would it make sense to add a second parameter to the method for adding the associated widget to indicate which functionality is desired?

rhaschke:
> Would it make sense to add a second parameter to the method for adding the associated widget to indicate which functionality is desired?

I agree that an extra boolean flag would be most clear to the users.
But I'm hesitating to change public API. Do you see a big issue in disconnecting the mentioned signal, i.e.:
`disconnect(panel, SIGNAL(visibilityChanged(bool)), display, SLOT(associatedPanelVisibilityChange(bool)));`
after calling `setAssociatedWidget()`?

rhaschke:
@Levi-Armstrong, do you think this PR (and https://github.com/ros-visualization/rviz/pull/1739#issuecomment-1107884401) works for you?

Levi-Armstrong:
I think so. I will pull this down and confirm.

PR #1744, opened 2022-05-16
Added regex filter field for TF  display

BlazPotokar:
### Description

When there is a lot of frames on `/tf` it can be hard to properly visualize them in RViz especially if frames overlap. Usually solution to this is to enable and disable desired frames in `Frames` field of the `TF display`. But it quickly become very annoying task if there is a lot of frames (e.g.  100 +) and each one needs to be found in a long list first and then disabled/enabled.

Because of that I suggest adding regex field which filters frames and displays only the ones matching entered expression. For convenience I also suggest adding additional field for blacklist filter.
Benefit of this approach is that user can quickly setup one or more displays, each showing specific group of frames and he can turn each one on/off as needed.
Scene with a lot of frames ...
image
... can be easily filtered so only interesting frames are displayed
image

Example when multiple TF displays are added. Each one has its own settings and frame filter.
image

### DIFF
--- a/src/rviz/default_plugin/tf_display.cpp
+++ b/src/rviz/default_plugin/tf_display.cpp
@@ -28,9 +28,13 @@
  */
 
 #include <boost/bind.hpp>
+#include <regex>
 
 #include <OgreSceneNode.h>
 #include <OgreSceneManager.h>
+#include <QValidator>
+#include <QLineEdit>
+#include <QToolTip>
 
 #include <tf/transform_listener.h>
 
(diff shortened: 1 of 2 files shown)


src/rviz/default_plugin/tf_display.cpp
```diff
@@ -40,6 +41,7 @@
 #include "rviz/ogre_helpers/axes.h"
 #include "rviz/ogre_helpers/movable_text.h"
 #include "rviz/properties/bool_property.h"
+#include "rviz/properties/color_property.h"
```
rhaschke, line 44: Why do you need to add ColorProperty?

src/rviz/default_plugin/tf_display.cpp
```diff
@@ -178,9 +179,13 @@ TFDisplay::TFDisplay() : Display(), update_timer_(0.0f), changing_single_frame_e
                                            "Whether or not arrows from child to parent should be shown.",
                                            this, SLOT(updateShowArrows()));
 
-  scale_property_ =
+  scale_property_ = 
       new FloatProperty("Marker Scale", 1, "Scaling factor for all names, axes and arrows.", this);
+  
+  filter_property_ = new StringProperty("Filter", ".*", "Regex filter", this);
 
+  filter_blacklist_property_ = new StringProperty("Filter (blacklist)", "", "Regex filter", this);
```
rhaschke, line 187: I suggest moving these properties as children of `frames_category_`. This makes it clear, where the filters are applied to. However, it makes the property harder to discover. What do you think?
Please rename "Filter" -> "Filter (whitelist)".

src/rviz/default_plugin/tf_display.cpp
```diff
@@ -372,22 +377,34 @@ void TFDisplay::updateFrames()
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #endif
 
-  context_->getTFClient()->getFrameStrings(frames);
+context_->getTFClient()->getFrameStrings(frames);
 
 #ifndef _WIN32
 #pragma GCC diagnostic pop
 #endif
   std::sort(frames.begin(), frames.end());
```
rhaschke, line 385: I suggest filtering frame names before sorting them (for efficiency).

src/rviz/default_plugin/tf_display.cpp
```diff
@@ -372,22 +377,34 @@ void TFDisplay::updateFrames()
...
       if (frame.empty())
-      {
         continue;
+
+      try
+      {
+        if (!std::regex_match(frame, whitelist_regex))
+          continue;
+        if (!blacklist_regex_string.empty())
+          if (std::regex_match(frame, blacklist_regex))
```
rhaschke, line 404: Prefer logical AND over nested IF:
```suggestion
        if (!blacklist_regex_string.empty()) && std::regex_match(frame, blacklist_regex))
```

src/rviz/default_plugin/tf_display.cpp
```diff
@@ -724,7 +741,8 @@ TFDisplay::M_FrameInfo::iterator TFDisplay::deleteFrame(M_FrameInfo::iterator it
   if (delete_properties)
   {
     delete frame->enabled_property_;
-    delete frame->tree_property_;
+    // delete frame->tree_property_;  // Here happens segmentation fault. With skipping deletinion, memory leak
+                                      //  appears but is not significant and only happens when user changes fields.
```
rhaschke, line 745: This needs to be fixed!

src/rviz/default_plugin/tf_display.h
```diff
@@ -119,7 +119,8 @@ private Q_SLOTS:
   BoolProperty* all_enabled_property_;
 
   FloatProperty* scale_property_;
-
+  StringProperty* filter_property_;
```
rhaschke, line 122: Rename:
```suggestion
  StringProperty* filter_whitelist_property_;
```

rhaschke (changes requested):
Thanks for this contribution. Please consider my comments.

rhaschke:
@BlazPotokar, do you plan to fix the remaining issues and finish this PR?

BlazPotokar:
Thank you so much for resolving the remaining issues!

PR #1743, opened 2022-05-14
Add better support for Qt5 style slots

kunaltyagi:
Closes #1742

Please check the proposed changes. I'll modify other properties once this is considered as an appropriate direction

API: adds ctor in properties
ABI: no change, except as indicated by API

### Description

Properties now have a ctor which allows providing normal C++ function pointers and lambda variables instead of Q_SLOTS

### Checklist

- [x] Choose the proper target branch: latest release branch, for non-ABI-breaking changes, *future* release branch otherwise.
      Due to the lack of active maintainers, we cannot provide support for older release branches anymore.
- [ ] Did you change how RViz works? Added new functionality? Do not forget to update the tutorials and/or documentation on the [ROS wiki](http://wiki.ros.org/rviz)
- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-visualization/rviz/pulls) to support the maintainers of RViz. Refer to the [RViz Wiki](https://github.com/ros-visualization/rviz/wiki/Maintainer-Guide#reviewing-pull-requests) for reviewing guidelines.

PR #1790, opened 2023-03-30
Modernize to use Qt5 slots

### LINKED ISSUE #1742: [feature request] Better Qt5 support for Slots
Currently, all properties have a ctor which takes in slots as `char*`. However, since Qt5, there's an option to take slot as a member function or even a functor.

Links
* [Qt4 details](https://woboq.com/blog/how-qt-signals-slots-work.html)
* [Qt5 details](https://woboq.com/blog/how-qt-signals-slots-work-part2-qt5.html)

Would a PR touching multiple property and derived class ctor be entertained?

rhaschke:
Closes #1743. Closes #1742.
This is  work-in-progress to modernize rviz' usage of signals and slots and switch to Qt5-style compile-time signal-slot definitions.

### DIFF
--- a/src/rviz/properties/property.h
+++ b/src/rviz/properties/property.h
@@ -132,9 +132,31 @@ public:
   Property(const QString& name = QString(),
            const QVariant& default_value = QVariant(),
            const QString& description = QString(),
-           Property* parent = nullptr,
-           const char* changed_slot = nullptr,
-           QObject* receiver = nullptr);
+           Property* parent = nullptr);
+
+  template <typename Func, typename R>
+  Property(const QString& name,
+           const QVariant& default_value,
+           const QString& description,
+           Property* parent,
+           Func&& changed_slot,
+           const R* receiver)
+    : Property(name, default_value, description, parent)
+  {
+    connect(receiver, std::forward<Func>(changed_slot));
+  }
+
+  // this variant is required to allow omitting the receiver argument
+  template <typename Func, typename P>
+  Property(const QString& name,
+           const QVariant& default_value,
+           const QString& description,
+           P* parent,
+           Func&& changed_slot)
+    : Property(name, default_value, description, parent)
+  {
+    connect(parent, std::forward<Func>(changed_slot));
+  }
 
   using QObject::connect; // inherit QObject's connect functions
 
--- a/src/rviz/properties/property.h
+++ b/src/rviz/properties/property.h
@@ -136,6 +136,40 @@ public:
            const char* changed_slot = nullptr,
            QObject* receiver = nullptr);
 
+  using QObject::connect; // inherit QObject's connect functions
+
+  /// Connect changed() signal to given slot of receiver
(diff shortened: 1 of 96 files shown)


rhaschke:
Providing new `connect()` methods allows simple and modular definition of slots for the `changed()` signal (16c9482f38f54d97328d0391ceb2e60e3e07052a).
Requiring that the slot can be passed to the ctor as well, requires two template variants for _all_ properties as illustrated in 0539624a76303411a26f4925112c6410703d7082: one with the optional receiver argument and one without.
It's not possible to combine those two into a single function with a default argument (`receiver=nullptr`) as before, we need to know the exact class type of either `receiver` or `parent` to pass pointer-to-member slots.

src/rviz/properties/property.h
```diff
@@ -132,9 +132,69 @@ class RVIZ_EXPORT Property : public QObject
...
+    static_assert(std::is_convertible<R*, typename QtPrivate::FunctionPointer<Func>::Object*>::value,
+                  "receiver type doesn't match that expected by member function");
+    return QObject::connect(this, &Property::changed, receiver, std::forward<Func>(slot), type);
+  }
+
+  /// Connect changed() signal to given slot functor, considering context
+  template <typename Func>
+  inline typename std::enable_if<!QtPrivate::FunctionPointer<Func>::IsPointerToMemberFunction,
+                                 QMetaObject::Connection>::type
+  connect(const QObject* context, Func&& slot, Qt::ConnectionType type = Qt::AutoConnection)
```
simonschmeisser, line 185: I find this slightly confusing and would suggest either a different name for the two parameter "variant" or dropping it altogether and just using the four parameter variant in the constructor directly.

src/rviz/properties/property.h
```diff
@@ -132,9 +132,69 @@ class RVIZ_EXPORT Property : public QObject
...
+  // this variant is required to allow omitting the receiver argument
+  template <typename Func, typename P>
+  Property(const QString& name,
+           const QVariant& default_value,
+           const QString& description,
+           P* parent,
+           Func&& changed_slot)
+    : Property(name, default_value, description, parent)
+  {
+    connect(parent, std::forward<Func>(changed_slot));
```
simonschmeisser, line 158:
```suggestion
    connect(this, &Property::changed,  parent, std::forward<Func>(changed_slot));
```
That's much clearer for me as it avoids another layer of indirection

rhaschke: This extra function call is required to dispatch to the old-style `char* slot` variant of connect implemented [here](https://github.com/ros-visualization/rviz/pull/1790/files#diff-ea41e84c9ac9bb8b449bea8284147f9f6e63cd62acd37874e1865567cef1c59eR76-R84) to maintain old behavior.
Surely, we could think of another, more specific method name. However, I used `connect`, because it's actually simply an overload of QObject's connect implementing a connection from the `changed` signal.

PR #1795, opened 2023-05-15
call propertyHiddenChanged synchronously

### LINKED ISSUE #1657: Crash when quickly toggling robot_description_property in RobotStateDisplay
I have a `RobotStateDisplay` that I use as a preview for an URDF text editor. When clicking somewhat quickly on the reload button it crashes. The reload button basically changes the robot_description on the param server and then calls
```
    robot_description_property_->setValue(QString());
    robot_description_property_->setValue(DefaultParamName);
```
to trigger a reload. I saw crashes a while ago when a Qt Quick Controls bug would emit the signal of a button twice and thus called update very rapidly. But now a collegue showed me that simply semi-rapidly clicking also leads to a crash after a while.

The backtrace is unfortunately very non-descript:
```
1  tf2_ros::TransformListener::subscription_callback_impl(ros::MessageEvent<tf2_msgs::TFMessage_<std::allocator<void>> const> const&, bool)                              0x7fe0435e2b8b 
2  rviz::PropertyTreeWidget::propertyHiddenChanged                                                                                          property_tree_widget.cpp 126 0x7fe047c2d80d 
3  QObject::event(QEvent *)                                                                                                                                              0x7fe045be8fb2 
4  QWidget::event(QEvent *)                                                                                                                                              0x7fe0480ad273 
5  QFrame::event(QEvent *)                                                                                                                                               0x7fe04815012e 
6  QAbstractScrollArea::event(QEvent *)                                                                                                                                  0x7fe048152e63 
7  QAbstractItemView::event(QEvent *)                                                                                                                                    0x7fe0482c7d9b 
8  QApplicationPrivate::notify_helper(QObject *, QEvent *)                                                                                                               0x7fe04806ce0c 
9  QApplication::notify(QObject *, QEvent *)                                                                                                                             0x7fe048074360 
10 QCoreApplication::notifyInternal2(QObject *, QEvent *)                                                                                                                0x7fe045bb88d8 
11 QCoreApplicationPrivate::sendPostedEvents(QObject *, int, QThreadData *)                                                                                              0x7fe045bbb497 
12 ??                                                                                                                                                                    0x7fe045c16373 
13 g_main_context_dispatch                                                                                                                                               0x7fe033b76537 
14 ??                                                                                                                                                                    0x7fe033b76770 
15 g_main_context_iteration                                                                                                                                              0x7fe033b767fc 
16 QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>)                                                                                            0x7fe045c1598f 
17 QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>)                                                                                                               0x7fe045bb6b2a 
18 QCoreApplication::exec()                                                                                                                                              0x7fe045bbfd30 
19 main                                                                                                                                     main.cpp                 296 0x557cf43ce080 

```

basically it seems that hiding invalidates some parts of the property tree that are later accessed in the slot

### Your environment
* OS Version: Ubuntu 18.04
* ROS Distro:  Melodic
    ```
[ INFO] [1629822612.423814117]: rviz version 1.14.8
[ INFO] [1629822612.423847305]: compiled against Qt version 5.12.10
[ INFO] [1629822612.423854628]: compiled against OGRE version 1.9.0 (Ghadamon)

    ```
* If source build, which git commit? 679c2853

simonschmeisser:
Property might no longer exist when event is triggered

fixes #1657

### DIFF
--- a/src/rviz/properties/property_tree_widget.cpp
+++ b/src/rviz/properties/property_tree_widget.cpp
@@ -109,7 +109,7 @@ void PropertyTreeWidget::setModel(PropertyTreeModel* model)
     m->deleteLater();
 
     connect(model_, SIGNAL(propertyHiddenChanged(const Property*)), this,
-            SLOT(propertyHiddenChanged(const Property*)), Qt::QueuedConnection);
+            SLOT(propertyHiddenChanged(const Property*)));
     connect(model_, SIGNAL(expand(const QModelIndex&)), this, SLOT(expand(const QModelIndex&)));
     connect(model_, SIGNAL(collapse(const QModelIndex&)), this, SLOT(collapse(const QModelIndex&)));
 


rhaschke:
I'm not sure why a queued connection was used here. But instead of changing the connection type (which might have other side effects), what about changing the argument from `Property*` to `QPointer<Property>`?
This way, we would notice any invalidation of the held Property and you could easily check this in `PropertyTreeWidget::propertyHiddenChanged`.

simonschmeisser:
I'll test `QPointer` as well, didn't have it in my mind.

simonschmeisser:
I didn't observe any difference between the QueuedConnection variant with QPointer and the synchronous version with the raw pointer.

rhaschke:
@simonschmeisser, could you please revert the 2nd commit? Thanks.

Issue #1796, opened 2023-05-21
What does "unknown reason for failure" mean when rviz is trying to display Lidar (/scan) data?


pitosalas:
ROS1, Ubuntu 22.04, Noetic.

When displaying LaserScan data in Rviz, I get this error:

<img width="618" alt="Screenshot 2023-05-21 at 12 47 38 PM" src="https://github.com/ros-visualization/rviz/assets/44681/6b2e363f-2fa4-4aad-8d46-ac41114b1c3a">

The tf and tf tree all look reasonable to me. I am not sure what the problem is with the tf that is causing the problem. I suspect it's the transform from /odom to /base_footprint but that looks ok:

<img width="645" alt="Screenshot 2023-05-21 at 12 54 28 PM" src="https://github.com/ros-visualization/rviz/assets/44681/bec1edfa-b4d0-4f95-88f5-5ea011aad949">

Can someone help me figure out what is causing it? I am happy to look at rviz source code if I knew where to look. Or any hint at all? Thanks!

rhaschke:
Please provide the full error message. Particularly, I am missing the two involved TF frames.

rhaschke:
The error originates from here:
https://github.com/ros-visualization/rviz/blob/f35dee1b8ebd0dccd225f2b48832dc0036dc6bfc/src/rviz/frame_manager.cpp#L331

This means that this call doesn't report an error _anymore_, which is strange:
https://github.com/ros-visualization/rviz/blob/f35dee1b8ebd0dccd225f2b48832dc0036dc6bfc/src/rviz/frame_manager.cpp#L325

pitosalas:
Here is some more info: The full message (which is not helpful):

<img width="847" alt="Screenshot 2023-05-21 at 3 06 03 PM" src="https://github.com/ros-visualization/rviz/assets/44681/bd071966-7912-40dc-91b4-59f5c4d2d3e3">

The tf tree in rviz:

<img width="469" alt="Screenshot 2023-05-21 at 3 11 00 PM" src="https://github.com/ros-visualization/rviz/assets/44681/6489f574-7367-4a6f-aafc-e41981f330c9">

Here is the tf tree in rat:

<img width="1482" alt="Screenshot 2023-05-21 at 3 20 02 PM" src="https://github.com/ros-visualization/rviz/assets/44681/11918bad-4a3d-4083-8d4c-a142deb8aaf5">

Here is output from tf_echo and tf_monitor:

```
[...]

PR #1797, opened 2023-05-25
Improve error handling in LaserScanDisplay

rhaschke:
... to avoid issues like #1796 in future.


rhaschke:
Need to target Melodic...

PR #1807, opened 2023-10-29
Fix Qt 5.15 API deprecation warnings

rhaschke:


rhaschke:
**Open issue:** When adding a `RobotModel` display, the `Global Options` property gets lost (only on Jammy).

PR #1813, opened 2024-01-15
Qt6 compat

simonschmeisser:
This fixes some errors (and warnings) when compiling with Qt6. Most was deprecated with 5.0 already.

The last commit is meant purely informational and should not be merged.
Unfortunately there is no PyQt6 available on Ubuntu 22.04 so I disabled the python bindings. Also choosing Qt6 over Qt5 does not seem to work somehow.


src/rviz/viewport_mouse_event.h
```diff
@@ -54,8 +54,13 @@ class ViewportMouseEvent
     : panel(p)
     , viewport(vp)
     , type(e->type())
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
```
rhaschke, line 57: I guess the new API is only available since Qt 6?
```suggestion
#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
```

src/CMakeLists.txt
```diff
@@ -3,4 +3,6 @@ add_subdirectory(image_view)
 if (CATKIN_ENABLE_TESTING)
   add_subdirectory(test)
 endif()
-add_subdirectory(python_bindings)
+if (QT_VERSION_MAJOR EQUAL 5)
+  add_subdirectory(python_bindings)
+endif()
```
rhaschke, line 8: What are the major issues with providing python support?

CMakeLists.txt
```diff
@@ -105,9 +105,12 @@ find_package(OpenGL REQUIRED)
 
 set(CMAKE_AUTOMOC ON)
 
-find_package(Qt5 REQUIRED COMPONENTS Core Widgets OpenGL)
-set(QT_LIBRARIES Qt5::Widgets)
-set(QTVERSION ${Qt5Widgets_VERSION})
+find_package(QT NAMES Qt6 REQUIRED COMPONENTS Core)
+# in theory: find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Core)
```
rhaschke, line 109: I would expect that _all_ components are available from the same Qt version:
```suggestion
find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Core Widgets OpenGL)
```

rhaschke (commented):
Thanks for these improvements towards Qt6. It would be great to add an appropriate CI config as well. I have only a few remarks:

src/rviz/viewport_mouse_event.h
simonschmeisser: > In addition to the existing imported targets, Qt 5.15 introduced versionless targets. That is, to link against [Qt Core](https://doc.qt.io/qt-6/qtcore-index.html) one can both reference Qt6::Core, or Qt::Core

https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html

src/CMakeLists.txt
simonschmeisser: Ubuntu 22.04 does not include a Qt6 version of pyqt/pyside so I was not able to compile `python-qt-binding`

CMakeLists.txt
simonschmeisser: In theory my version would fail if there is Qt6 Core installed but not OpenGL, yours would fallback to Qt5 if that is more complete.

In practice somehow it always picks Qt5 even if Qt6 is available

src/rviz/viewport_mouse_event.h
rhaschke: Ok. But this is unrelated to the API change required here, i.e. `QMouseEvent::[x|y]()` vs. `QMouseEvent::position().[x|y]()`.

CMakeLists.txt
rhaschke: https://github.com/isys-vision/rviz/pull/2

src/rviz/viewport_mouse_event.h
simonschmeisser: Sorry, I wasn't reading the context very carefully and have a cold ...

> This function was introduced in Qt 5.14.

https://doc.qt.io/qt-5/qwheelevent.html#position

I assume the previous way was deprecated in 5.15 but haven't looked it up. Actually I just copy pasted from line 76/81 below

rhaschke: In contrast to those lines we are dealing with a `QMouseEvent` here, not a `QWheelEvent`. Without the proposed change, I had a compiler failure with Qt 5.15.

CMakeLists.txt
simonschmeisser: Your version successfully picked Qt6 here when available, thanks! Maybe the `NAMES` needs a certain CMake version to work? I took it from the Qt docs ...

simonschmeisser:
Thanks for your patience Robert and sorry for not being fully concentrated!

Issue #1826, opened 2024-05-13
Map Display jumps due to Qt::QueuedConnection


MichaelGrupp:
After the recent upgrade from 1.14.20 to 1.14.23, we observed glitches when rendering dynamic maps such as costmaps.

I tracked the problem down to this commit: https://github.com/ros-visualization/rviz/commit/93ac44b26babb298046694efe78f8f986f7219e9 (PR https://github.com/ros-visualization/rviz/pull/1793)

This "hack" seems to cause a delay in map rendering that leads to a broken map display.

### With QueuedConnection

Using an unmodified 1.14.23 from the ROS noetic package distro. The costmap is jumping while moving. Note that the costmap frame_id is in a fixed frame.

https://github.com/ros-visualization/rviz/assets/8985495/6e592999-bd06-4403-afaa-9803da4773c7

### Without QueuedConnection

I cloned RViz from source at 1.14.23, removed that `Qt::QueuedConnection` flag, and built it. Using this patched RViz, the costmap is rendered at the correct spot again without jumping:

https://github.com/ros-visualization/rviz/assets/8985495/b0494338-cfb6-4262-9a3f-1a53f9a75c76

### Your environment
* OS Version: e.g. Ubuntu 20.04
* ROS Distro: Noetic
* RViz, Qt, OGRE, OpenGl version as printed by rviz:
```
[${node} INFO 1715596133.955965167]: rviz version 1.14.23
[${node} INFO 1715596133.956069614]: compiled against Qt version 5.12.8
[${node} INFO 1715596133.956126962]: compiled against OGRE version 1.9.0 (Ghadamon)
[/rviz_1715596402254328779 INFO 1715596402.270469085]: Forcing OpenGl version 0.
[/rviz_1715596402254328779 INFO 1715596402.742861634, 1715596391.783960104]: Stereo is NOT SUPPORTED
[/rviz_1715596402254328779 INFO 1715596402.743000877, 1715596391.783960104]: OpenGL device: AMD RENOIR (DRM 3.44.0, 5.16.15-051615-generic, LLVM 12.0.0)
[/rviz_1715596402254328779 INFO 1715596402.743048266, 1715596391.783960104]: OpenGl version: 4.6 (GLSL 4.6) limited to GLSL 1.4 on Mesa system.
```
* If source build, which git commit? -
* System locale, i.e. the output of `echo "$LANG $LC_NUMERIC"`: `en_US.utf8` (`LANG=C rviz` doesn't change anything here)
[...]

PR #1827, opened 2024-05-13
WIP: workaround for map display glitch (issue #1826)

MichaelGrupp:
Illustrates the workaround described in #1826 

PR is marked as work in progress because this probably shouldn't be merged if it breaks NVIDIA users, as described in the comment. However, I don't have an idea for a better fix at the moment.


rhaschke (commented):
Dropping the queued connection, brings back the segfault on NVIDIA hardware which is even worse. So, that's not a viable solution.

PR #1793, opened 2023-05-04
Fixed crash when raycasting while map is updated.

StefanFabian:
When using the scene manager raycast in one of our tool rviz crashed when the tool touched the map area while a new map was received.
Apparently, the connection is not queued (anymore?) and therefore the update does not happen on the main thread.
This small change should fix this and is fully ABI compatible.

Fixes #1792

PR #1828, opened 2024-05-15
Fix #1826: jitter in Map display updates

### LINKED ISSUE #1826: Map Display jumps due to Qt::QueuedConnection

rhaschke:
The queued connection update introduced in #1793 caused the map display to first update the pose of the map and then the map itself (in the next update cycle). The resulting jittering can be perfectly seen when throttling the rviz frame rate.

Updates to the visualization should be handled in update() only. Thus, now, if a map update is received, a flag is set to perform the costly map update. If that flag is not set, only the transform is updated.

Fixes #1826

### DIFF
--- a/src/rviz/default_plugin/map_display.cpp
+++ b/src/rviz/default_plugin/map_display.cpp
@@ -219,12 +219,9 @@ void Swatch::updateData()
 }
 
 
-MapDisplay::MapDisplay() : Display(), loaded_(false), resolution_(0.0f), width_(0), height_(0)
+MapDisplay::MapDisplay()
+  : Display(), loaded_(false), map_updated_(false), resolution_(0.0f), width_(0), height_(0)
 {
-  // HACK: Using a direct connection triggers a segfault on NVIDIA hardware (#1793) when rendering
-  //       *and* having performed a depth rendering before (e.g. due to raycasting for selections)
-  // A queued connection delays the update of renderables after the current VisualizationManager::onUpdate() call
-  connect(this, &MapDisplay::mapUpdated, this, &MapDisplay::showMap, Qt::QueuedConnection);
   topic_property_ = new RosTopicProperty(
       "Topic", "", QString::fromStdString(ros::message_traits::datatype<nav_msgs::OccupancyGrid>()),
       "nav_msgs::OccupancyGrid topic to subscribe to.", this, &MapDisplay::updateTopic);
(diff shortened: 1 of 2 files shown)


MichaelGrupp (approved):
I tested again all the cases that were shown in the ticket with this patch and can confirm that this fixes the issue. Also code-wise: LGTM

Thanks a lot for the fast fix!!

PR #1843, opened 2024-12-04
Warning Dialog for ROS 1 EOL.

kscottz:
image

I've been discussing the ROS 1 EOL with @sloretz. 20% of the ROS community has not upgraded to ROS 2, and some of them seem to be unaware of the impending ROS 1 EOL. We would like to add a series of warnings / notifications to ROS 1 and RViz to encourage users to upgrade to ROS 2. We have made [similar changes to Gazebo Classic](https://github.com/gazebosim/gazebo-classic/pull/3405). This is the first change in a series of changes we would like to add to Noetic in the next few weeks. 

### Description

This pull request adds a critical dialog on startup to notify users of the ROS 1 EOL date and where they can find migration information. Users can disable the warning by setting the `DISABLE_ROS1_EOL_WARNINGS` environment variable. 

We are still deciding on a final URL location. I will update the pull request when I have that information.


wjwwood:
👍 from me generally.

I'm not sure how much work it would be, but a "notification toast" which doesn't take focus from the rest of the window might be better. On the other hand it's easy to disable the warnings, so maybe it's fine as proposed.

Could use something like https://github.com/niklashenning/qt-toast or implement something like it manually.

src/rviz/noetic_eol_dialog.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <rviz/noetic_eol_dialog.h>
+
+namespace rviz
+{
+NoeticEOLDialog::NoeticEOLDialog(QWidget* parent) : QMessageBox(parent)
+{
+  setIcon(QMessageBox::Critical);
+  std::stringstream ss;
+  ss << "ROS 1 goes end-of-life 2025-05-31\n\n";
+  ss << "Users are encouraged to migrate to ROS 2 as soon as possible!\n\n";
```
rhaschke, line 49:
```suggestion
  ss << "Users are encouraged to migrate to ROS 2 as soon as possible!\n";
  ss << "Alternatively, switch to the ROS One distribution\n\n";
```

rhaschke (commented):
I think the dialog should (also) mention the [ROS-O initiative](https://github.com/ros-o/ros-o) providing ROS1 support for newer Ubuntu releases.

src/rviz/noetic_eol_dialog.cpp
kscottz: This second new line is intentional.

kscottz:
> I think the dialog should (also) mention the [ROS-O initiative](https://github.com/ros-o/ros-o) providing ROS1 support for newer Ubuntu releases.

@rhaschke The plan is to point to documentation laying out available migration options for current Noetic users. Our plan is to include ROS-O on that list. I want to keep the notification itself short and sweet as we simply can't provide enough context in warning message.

kscottz:
@wjwwood A toast would be nice but I am leary about pulling in another dependency. Let me see if I can adjust the warning to be non-blocking.

src/rviz/noetic_eol_dialog.cpp
rhaschke: Yes, I just shifted the extra newline to the next comment line.

rhaschke:
> The plan is to point to documentation laying out available migration options for current Noetic users. Our plan is to include ROS-O on that list.

I think it is important to mention both ROS2 and ROS-O in the warning. Otherwise, if only ROS2 is mentioned, users wanting to stick to ROS1 might not even click the link.

sloretz:
> I think it is important to mention both ROS2 and ROS-O in the warning. Otherwise, if only ROS2 is mentioned, users wanting to stick to ROS1 might not even click the link.

Mentioning ROS-O with ROS 2 might give the impression that ROS-O is officially supported by the OSRA, or is continuing development of ROS 1, but [ROS-O's goal is only to build on newer distros](https://github.com/ros-o/ros-o/blob/6d1edf7d51f74eac07ed0d5a9e6b714a90fba5a2/README.md?plain=1#L5-L7). IIUC someone interested in ROS-O might also be interested in [Canonical's ROS ESM](https://ubuntu.com/robotics/ros-esm).

How about we limit the dialog to just a link to a page describing the different options (ROS 2, ROS-O, ROS ESM, forking ROS 1, etc)?

```
ROS Noetic goes end-of-life 2025-05-31

Learn about your options here <some URL TBD>

To disable this dialog, set the DISABLE_ROS1_EOL_WARNINGS environment variable.
```

kscottz:
@rhaschke [I have a draft](https://github.com/ros-infrastructure/www.ros.org/pull/54) of the post we would like to use in this warning. Any feedback you have here would be welcome.

If that looks good I can address the remaining comments and set the appropriate URL.

PR #1844, opened 2024-12-11
Patch 1

akazxd:
`boost::filesystem::extension` was deprecated in Boost 1.77 and removed in 1.85
https://www.boost.org/doc/libs/1_87_0_beta1/libs/filesystem/doc/release_history.html

The alternative, calling the member function `extension()` on `path` works on Noetic too.


rhaschke:
Which version of boost are you using? Which Ubuntu release provides that version as a default?

akazxd:
using boost 1.86, on Arch linux. I was trying to build from AUR.

rhaschke (approved):

rhaschke:
Thanks!

PR #1850, opened 2025-07-08
wayland fallback on X11-Ogre

v4hn:
When running on wayland with Ogre without wayland support, explicitly force Qt to use the XCB protocol running through Xwayland.

Backported with adaptations from https://github.com/ros2/rviz/issues/847 .

@rhaschke as discussed in person.
Tested on Debian trixie to run under X11 as well as wayland.

### DIFF
--- a/src/rviz/main.cpp
+++ b/src/rviz/main.cpp
@@ -36,22 +36,25 @@ int main(int argc, char** argv)
 {
 #if !OGRE_USE_WAYLAND
   // If Ogre does not support wayland, but we are running on wayland,
-  // explicitly force the xcb platform in Qt unless the user set it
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
Thanks for this PR. I need to look into CI. Looks like OSRF broke Noetic by removing the urdf package key :frowning_face:

simonschmeisser:
What version of ogre are you using? From Debian package? Maybe wayland support can be enabled?

simonschmeisser:
Wayland support was actually only added super recently in 14.something in Ogre so this workaround will likely be needed forever

rhaschke (approved):

PR #1851, opened 2025-07-08
Do not warn about noetic EOF in upstream post-noetic-release

v4hn:
This made sense for the last release in Noetic, but there is nothing wrong in ROS-O releases and defining the environment variable everywhere is a cumbersome workaround.

Maybe it makes sense to push this in a new branch if we want to keep noetic-devel targeted only at ROS noetic?

Revert "Warning Dialog for ROS 1 EOL (#1843)"

This reverts commit 8fe12bbb8e62ebebf16e85517dabd371443de6dc.


rhaschke (approved):
Fair enough. Maybe we should branch off to `ros-o` to clarify that development is continuing for ROS-O? I think I never released a new rviz version with this warning...

v4hn:
a new branch here would make sense to me. How about `main`?

PR #1852, opened 2025-07-11
Enable to reset image display when new image has not been received for a while

pazeshun:
With this PR, you can switch on `Reset When Timed Out` function, which makes image display show "No Image" again when new image has not been received for `Timeout` seconds.
This PR makes it easier to notice that the image publisher is in trouble.

# Before this PR

https://github.com/user-attachments/assets/55ee0e1f-c0b3-45c2-9175-3067269e16ce

# After this PR

https://github.com/user-attachments/assets/366eac6f-f3e0-41c6-bc36-a80ac79daa44

Note that the default state of `Reset When Timed Out` is OFF, which leads to the same behavior before this PR.

### DIFF
--- a/src/rviz/image/image_display_base.cpp
+++ b/src/rviz/image/image_display_base.cpp
@@ -41,7 +41,7 @@
 
 namespace rviz
 {
-ImageDisplayBase::ImageDisplayBase() : Display(), sub_(), tf_filter_(), messages_received_(0), timeout_tm_(0), is_img_up_(false)
+ImageDisplayBase::ImageDisplayBase() : Display(), sub_(), tf_filter_(), messages_received_(0)
 {
   topic_property_ = new RosTopicProperty(
       "Image Topic", "", QString::fromStdString(ros::message_traits::datatype<sensor_msgs::Image>()),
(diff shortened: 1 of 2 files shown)


src/rviz/image/image_display_base.cpp
```diff
@@ -261,6 +278,18 @@ void ImageDisplayBase::updateTopic()
   context_->queueRender();
 }
 
+void ImageDisplayBase::updateResetTO()
+{
+  if (reset_to_property_->getBool())
+  {
+    reset_to_timer_->start(33);
```
rhaschke, line 285: Why this fixed timeout value? How is it motivated?

rhaschke (commented):
Interesting feature, but the implementation seems to be overly complicated.
You are running a periodic QTimer and update the `timeout_time_` in its associated slot. Instead, you could simply update a `last_received_` timestamp within `incomingMessage()` and directly call reset() in the slot function.
Are you fearing data races?

src/rviz/image/image_display_base.cpp
```diff
@@ -65,6 +65,18 @@ ImageDisplayBase::ImageDisplayBase() : Display(), sub_(), tf_filter_(), messages
 
   unreliable_property_ = new BoolProperty("Unreliable", false, "Prefer UDP topic transport", this,
                                           &ImageDisplayBase::updateTopic);
+
+  reset_to_property_ = new BoolProperty("Reset When Timed Out", false,
+                                        "Reset when new image has not been received for Timeout [sec].",
+                                        this, &ImageDisplayBase::updateResetTO);
```
rhaschke, line 71: Please drop this property and encode no timeout via `timeout_property_ = 0`.
This is to avoid cluttering the property interface and is frequently used in similar situations.

src/rviz/image/image_display_base.h
```diff
@@ -77,9 +79,15 @@ protected Q_SLOTS:
   /** @brief Update queue size of tf filter  */
   virtual void updateQueueSize();
 
+  /** @brief Prepare for or stop resetting when timed out */
+  virtual void updateResetTO();
```
rhaschke, line 83: Why is this virtual? Do you see any need to overwrite that method in sub classes?

src/rviz/image/image_display_base.h
```diff
@@ -131,6 +139,15 @@ protected Q_SLOTS:
...
   BoolProperty* unreliable_property_;
+
+  BoolProperty* reset_to_property_;
+
+  FloatProperty* timeout_property_;
+
+  ros::Time timeout_tm_;
+  bool is_img_up_;
+
+  QTimer* reset_to_timer_;
```
rhaschke, line 150: Use more verbose variable names!

pazeshun:
@rhaschke Thank you very much for your support, and I am deeply sorry for my super late response.
I confirmed your changes perfectly work in our usage.
One thing I should mention is that the current default behavior (`Timeout` is `1.0`) is different from the default behavior before this PR.
If you consider this is problem, please merge https://github.com/ros-visualization/rviz/pull/1856.
Anyway, thank you very much again.

rhaschke:
I changed the default behavior deliberatively to report issues after 1s.

pazeshun:
I see. Thank you very much!

PR #1853, opened 2025-07-11
Backport #1852 to melodic

pazeshun:
Melodic version of #1852.
This PR is just for someone who is using melodic for some reasons and interested in #1852.


rhaschke:
Melodic is EOL since 2023. We are not going to merge anything to melodic (we cannot even release Noetic binaries anymore). If you are interested in using ROS1 on modern Ubuntu release, consider [ROS-O binaries](https://ros.packages.techfak.net).

